PR #5, opened 2015-10-29
Fancy log index

mvollmer:


mvollmer:
I have tested this with https://github.com/mvollmer/cockpit/tree/fancy-log-index

sink/sink
```diff
@@ -210,12 +211,21 @@ class Badger(object):
...
                 f.write('<html><head><meta http-equiv="refresh" content="0; {url}"></head></html>'.format(url=data['link']))
 
+class Extras(object):
+    def __init__(self, config):
+        pass
+
+    def push(self, data):
+        if 'extras' in data:
+            for url in data['extras']:
+                urllib.urlretrieve(url, os.path.basename(url))
```
stefwalter: If this fails, the logging shouldn't fail right? Should we catch exceptions here?

sink/README.md
```diff
@@ -71,3 +73,16 @@ The badge itself is a small image that includes `description` and
...
 `status` is used instead.
+
+### Extras format
+
+The `"extra"` part of a status object should be an array of strings.
+Each string is a URL and is downloaded into the log directory.  The
+last component of the URL will be the name of the file in the log
+directory.
+
+### Index format
```
stefwalter: Can we call this "log" instead of "index"? It makes it easier to understand.

sink/sink
```diff
@@ -210,12 +211,21 @@ class Badger(object):
...
+            del data['extras']
+
 class Status(object):
     def __init__(self, config, identifier):
-        self.reporters = [ GitHub(config), Irc(config), Badger(config) ]
-        self.link = config.get("Sink", "Url").replace('@@', identifier)
+        self.reporters = [ GitHub(config), Irc(config), Badger(config), Extras(config) ]
+        self.dirlink = config.get("Sink", "Url").replace('@@', identifier)
         self.data = None
-        sys.stderr.write(self.link + "\n")
```
stefwalter: The log link still needs to be written for logs that have no status object.

sink/sink
mvollmer: All `push` calls run inside `try` already.  Is that enough?

sink/README.md
mvollmer: Yes, done. I also added more words.

sink/sink
mvollmer: Fixed.

sink/sink
stefwalter: Ah yes, it does. That's good enough.

stefwalter:
I've posted a cleaner mechanism for the log/index stuff here: https://github.com/cockpit-project/cockpituous/pull/6

What do you think? If you like #6, could you rebase ontop of that?

The 'extra' stuff still makes sense of course.

mvollmer:
> What do you think? If you like #6, could you rebase ontop of that?

Yes, very nice.  I have rebased this, updated https://github.com/cockpit-project/cockpit/pull/3082, and test it all.  Works.

mvollmer:
Test run: https://mvo.fedorapeople.org/logs/test-6bf1f7cb-fedora-23-x86_64/log.html

PR #8, opened 2015-11-02
verify: Mention .ssh/config

mvollmer:
So that the container will log into cockpit@fedorapeople.org and not
user@fedorapeople.org.


mvollmer:
I guess this is not the last word... What about letting the container do `ssh logsink` instead of `ssh fedorapeople.org` and tell people to put the right things into .ssh/config for `logsink` so that the credentials they have in .ssh work?

stefwalter:
Yeah, we also need to set up known_hosts.

PR #11, opened 2015-11-04
verify: Set TMPDIR to /build/tmp inside the container

### LINKED ISSUE #10: /tmp in the infra-verify container runs out of space
It was full of rpmbuild.XXXX things.

mvollmer:
So that tools/make-srpm has enough space.

Fixes #10


stefwalter:
This needs to go into the Dockerfile instead.

stefwalter:
Fixed in Dockerfile instead.

PR #13, opened 2015-11-13
verify: Fetch the current origin/master before starting

mvollmer:
Normally, check-verify updates our local repo, but if it doesn't
because it is buggy, we would get stuck with that buggy master.


stefwalter:
Looks good to me.

PR #14, opened 2015-11-19
sink: Create log directories in a collision free way

stefwalter:
This means if two sinks start logging to a certain directory,
the earlier one will end up as a backup.XXXX directory, and won't
continue to log to the main directory or overwrite files there

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -303,6 +303,34 @@ class Buffer(object):
         result += os.read(self.fd, 1024)
         return result
 
+# Create the directory in a race free way
+def mkdir_and_chdir(base, identifier):
+    directory = os.path.abspath(os.path.join(base, identifier))
+
+    # 1. Create a temporary non-empty directory and get its handle
+    tempdir = tempfile.mkdtemp(prefix=identifier, dir=base)
+    os.close(os.open(os.path.join(tempdir, ".sink"), os.O_WRONLY | os.O_CREAT))
+    dirfd = os.open(tempdir, os.O_RDONLY)
+
+    # 2. Rename directory into place
+    while True:
+        try:
+            os.rename(tempdir, directory)
+            break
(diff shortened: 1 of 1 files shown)


stefwalter:
@mvollmer Could you take a look at this?

mvollmer:
I closed this too early, sorry.

sink/sink
```diff
@@ -303,6 +303,31 @@ class Buffer(object):
...
+
+    # 2. Rename directory into place
+    while True:
+        try:
+            os.rename(tempdir, directory)
+            break
+        except OSError as ex:
+            # 3. If that raced, then move the target into our directory
+            if ex.errno == errno.EEXIST or ex.errno == errno.ENOTEMPTY:
+                os.rename(directory, tempfile.mkdtemp(prefix="backup.", dir="."))
```
mvollmer: This places the backup directory into ".", which is normally the home directory of the user that runs the sink, not the newly created directory.

sink/sink
```diff
@@ -303,6 +303,31 @@ class Buffer(object):
         result += os.read(self.fd, 1024)
         return result
 
+# Create the directory in a race free way
+def mkdir_and_chdir(base, identifier):
+    directory = os.path.abspath(os.path.join(base, identifier))
+
+    # 1. Create a temporary non-empty directory and get its handle
+    tempdir = tempfile.mkdtemp(prefix=identifier, dir=base)
+    os.close(os.open(os.path.join(tempdir, ".sink"), os.O_WRONLY | os.O_CREAT))
```
mvollmer: What is the point of creating this file?

stefwalter: Renaming over an empty directory does not result in EEXIST. Therefore in order to claim this directory (and not have someone blow it away) it needs an dir entry in it.

PR #15, opened 2015-12-15
Use github-task as the entry point

stefwalter:


stefwalter:
This depends on https://github.com/cockpit-project/cockpit/pull/3332

PR #17, opened 2016-01-05
Release debian source

mvollmer:


mvollmer:
This depends on https://github.com/cockpit-project/cockpit/pull/3423

mvollmer:
I have tested release-debian-source, but not the whole end-to-end process.  Is there a better way to test this than just running it for real?

release/release-debian-source
```diff
@@ -0,0 +1,175 @@
...
+      # - don't run autogen.sh
+      # - don't install or package test assets
+      sed -i -e 's/--with autoreconf//' -e 's/install-test-assets//' debian/rules
+      sed -i -e '/Package: cockpit-test-assets/,/^$/d' debian/control
+    )
+
+    changelog_lines >$WORKDIR/debian/changelog
+    ( cd $REPODIR
+      dpkg-source -b $absworkdir
+      dpkg-scansources . >Sources
```
stefwalter: I need to add these to the Dockerfile that creates the release container. Where can I find them?

release/release-debian-source
```diff
@@ -0,0 +1,175 @@
...
+      gzip <Sources >Sources.gz
+    )
+}
+
+commit()
+(
+    trace "Publishing Debian source repository"
+
+    rsync --delete -a -e ssh $REPODIR/ fedorapeople.org:/project/cockpit/debian-unstable
```
stefwalter: So will the repo only contain that one release? Is it okay that this will blow away everything else?

stefwalter:
Is this script idempotent? If we run it repeatedly will it converge to a single consistent result? Many of the other scripts have checks to see if something has already been released and prevent another release. Does this script need something like that?

martinpitt:
Looks like this is obsolete -- we already [do publish a Debian source archive](http://repo-cockpitproject.rhcloud.com/debian/dists/jessie/main/source/Sources.gz). @mvollmer, can you please confirm that this can be closed?

mvollmer:
I have no memory of this... :-D  So let's close this.

PR #18, opened 2016-01-27
WIP - sink: Support for creating issues and pull requests

mvollmer:

PR #19, opened 2016-01-28
sink: Add generic GitHub REST engine

mvollmer:


mvollmer:
This replaces #18

mvollmer:
This works OK with https://github.com/cockpit-project/cockpit/pull/3608 for making issues and pull requests for weekly images.  I haven't yet tried to use it for verify statuses.

mvollmer:
> This works OK with cockpit-project/cockpit#3608 for making issues and pull requests for weekly images. I haven't yet tried to use it for verify statuses.

https://github.com/cockpit-project/cockpit/pull/3608 now also creates statuses, so this works as well.

mvollmer:
Also tested via https://github.com/cockpit-project/cockpit/issues/3648

sink/README.md
```diff
@@ -40,17 +40,87 @@ a tarball to follow. This tarball will be extracted into the log directory.
...
+This will create a new issue in the "cockpit" repo of the
+authenticated user and add a comment to it.
+
+If a string is of the form "{path}" (i.e., it starts with "{" and ends
+with "}"), it is replaced completely by the value indicated by the
+path.  This might be string or some other JSON value such as a number.
+[ I don't know whether we need this, GitHub seems to be happy with
+strings instead of numbers, for example. ]
+
+For compatibility, the `"github"` object can also have the following
```
stefwalter: I'd like to suggest we remove documentation for this, and remove support in the next month or so.

sink/README.md
```diff
@@ -40,17 +40,87 @@ a tarball to follow. This tarball will be extracted into the log directory.
...
+
+  { "github":
+    { "token": "......",
+      "requests": [
+        { "method": "GET",
+          "resource": "/user",
+          "result": "user"
+        },
+        { "method": "POST",
+          "resource": "repos/${user/login}/cockpit/issues",
```
stefwalter: Have you thought about using the github syntax for this instead? I notice variables are represented like ```/orgs/:org/issues``` with ```org``` being the variable interpolated.

sink/README.md
mvollmer: Okay, makes sense.

sink/README.md
mvollmer: Hmm, how would we write `${user/login}`? `:user:login`? Or `:{user/login}`?

stefwalter: Yes, it may be that we cannot use this scheme. 

I wonder why this particular case isn't hard coded by the caller though? Wouldn't they want the issue to be created at a specific resource?

mvollmer:
Changed the interpolation syntax to something closer to Ruby and JSONPath: `:user.login`.

mvollmer:
Changed `:sink/link` to just `:link`.

stefwalter:
Great. Merged.

PR #22, opened 2016-02-18
Dockerhub Enhancements

petervo:
Adds support for pushing to a fork and multiple docker files.

### DIFF
--- a/release/release-dockerhub
+++ b/release/release-dockerhub
@@ -36,7 +36,7 @@ CLEANUP=""
 
 usage()
 {
-    echo "usage: release-dockerhub [-qvxz] [-p SRPM] [-s SOURCE_REPO ] REPO" >&2
+    echo "usage: release-dockerhub [-qvxz] [-p SRPM] REPO [SOURCE_REPO]" >&2
     exit ${1-2}
 }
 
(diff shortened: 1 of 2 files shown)


release/release-dockerhub
```diff
@@ -29,13 +29,14 @@ SRPM=${RELEASE_SRPM:-$PWD/srpm}
...
 REPO=""
+SOURCE_REPO=""
 WORKDIR=""
 PUSH=""
 CLEANUP=""
 
 usage()
 {
-    echo "usage: release-dockerhub [-qvxz] [-p SRPM] REPO" >&2
+    echo "usage: release-dockerhub [-qvxz] [-p SRPM] [-s SOURCE_REPO ] REPO" >&2
```
stefwalter: Can we make this just be an optional two arguments instead of one?

stefwalter: That would make it much more clear in the script file what is happening.

petervo: We always need a target repo. Having two ways to pass it in seems more confusing to me.

Or did you mean add a -r REPO required argument and remove the $1 REPO argument?

stefwalter: I can make the changes ... one string argument would indicate $SOURCE_REPO == $REPO ... two would mean they are different.

petervo: I pushed a fixup commit without amending in case it conflicts with what you are doing.

PR #25, opened 2016-03-22
release-debian: new release script

larsu:
Builds and publishes a debian repository from the source repository
generated by release-debian-source.

Uses pbuilder, which requires root access for creating a chroot
environment. This, run the container in privileged mode and allow the
user inside the container to execute pbuilder as root.


release/release-debian
```diff
@@ -0,0 +1,145 @@
...
+{
+    echo "release-debian: $@" >&2
+}
+
+check()
+{
+    true
+}
+
+prepare()
```
stefwalter: Is this function idempotent? If it's called again, on exactly the same input, will it skip it's task?

larsu: Reprepro will not add packages with the same name and version again, but the pbuilder is created and packages are built every time.

Do you think we should have a check at the beginning of the function? I wasn't aware that this is called often.

stefwalter: Well so many of these release things fail, that it'll be scheduled to run again. The entire script and all jobs run again, and those that have already run typically just produce output and skip most of their tasks.

release/release-debian
```diff
@@ -0,0 +1,145 @@
...
+{
+    trace "Creating Debian repository"
+
+    rm -rf debian
+    cp -r tools/debian .
+
+    printf "USENETWORK=yes\n" > pbuilderrc
+
+    sudo pbuilder clean
+    sudo pbuilder create --distribution unstable
```
stefwalter: Why does this download and install packages every single time its run. Is that normal? It takes a long time to make this happen.

larsu: Because `docker build` doesn't allow privileged operations. If it did, we could move the "create" step into the Dockerfile. We'd still need to call `pbuilder update` here to make sure we're testing on latest unstable. I'm not sure if that would gain anything (depends on how much churn there is in the base packages).

Pbuilder always downloads and installs build dependencies when building a package.

We could mount /var/cache/pbuilder/aptcache from somewhere, so that we don't need to download the packages again after exiting the container.

stefwalter: Okay, lets outgrow the Dockerfile for this. I'll create a commit for that for you to play with.

release/release-debian
larsu: Makes sense. I've added a commit that does this.

larsu:
Maybe we should also pull build dependencies into the docker image? Then we'll have them in the cache as well.

stefwalter:
> Maybe we should also pull build dependencies into the docker image? Then we'll have them in the cache as well.

Yes that would be good. I noticed that it's running through a long install process during release-debian even with 'pbuilder create' in the 'make release-container' stage.

larsu:
Ok, done. There isn't a good way to extract the dependencies from the control file from that script, so I've copied them in.

PR #27, opened 2016-03-24
release: Escape asterisks in release-koji regular expressions

stefwalter:
This was causing %changelog in rpms to have both stars and dashes


andreasn:
Here is an example of when that happened
https://bodhi.fedoraproject.org/updates/FEDORA-2016-305d744aa3

PR #28, opened 2016-04-05
verify: Run 'npm install' whenever package.json differs from master

stefwalter:
This helps us get all the right dependencies in place

### DIFF
--- a/verify/cockpit-verify
+++ b/verify/cockpit-verify
@@ -14,10 +14,7 @@ if [ ! -d cockpit ]; then
     git clone https://github.com/cockpit-project/cockpit
 fi
 
-if [ ! -d cockpit/node_modules ]; then
-    ( cd cockpit && npm install )
-fi
-
+(cd cockpit && npm install)
 mkdir -p mock
 
 echo "Starting testing"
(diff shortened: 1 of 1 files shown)


verify/cockpit-verify
```diff
@@ -28,7 +26,17 @@ while true; do
     git -C cockpit fetch origin master
     git -C cockpit reset --hard FETCH_HEAD
     cockpit/test/vm-download --prune
+
+    if ! git diff -C cockpit --exit-code origin/master package.json; then
+        install=1
+    fi
+
+    if [ "$install" = "1" ]; then
+        (cd cockpit && npm install)
```
dperpeet: During the first run, this will do npm install from the first branch it's on, but in subsequent runs it compares the package.json to master. Essentially something different could be compared than is executed.
Maybe store the revision to compare against? Or a checksum?

stefwalter: Lets just run it explicitly and and copy/paste this line.

verify/cockpit-verify
```diff
@@ -28,7 +25,12 @@ while true; do
     git -C cockpit fetch origin master
     git -C cockpit reset --hard FETCH_HEAD
     cockpit/test/vm-download --prune
+
+    if ! git diff -C cockpit --exit-code origin/master package.json; then
```
dperpeet: This will leave an updated install for subsequent runs with branches that don't deviate from master's package.json, but it's probably ok for now.

stefwalter:
This is broken. It doesn't use the right node_modules.

PR #30, opened 2016-04-13
release-debian: also publish to ubuntu ppa

larsu:
Publish debian source package to a [ppa on launchpad][1].

I've added the public ssh (for uploading packages via sftp) and pgp (for verifying the signed packages) keys to the launchpad user *cockpit-project*.

@stefwalter, please validate the pgp key by decrypting the email launchpad sent to cockpituous@gmail.com and following the link inside.

Launchpad will send an email every time new packages are uploaded to signal whether the upload was accepted. It will also notify about build failures.

[1]: https://launchpad.net/~cockpit-project/+archive/ubuntu/cockpit

### DIFF
--- a/release/cockpit-script
+++ b/release/cockpit-script
@@ -29,5 +29,6 @@ job release-bodhi F24
 # Upload documentation
 job release-guide doc/guide/html cockpit-project/cockpit-project.github.io
 
-# Create and publish a Debian repository
+# Create and publish a Debian repository and Ubuntu PPA
 job release-debian
+job release-ubuntu-ppa
--- a/release/release-debian
+++ b/release/release-debian
@@ -85,8 +85,7 @@ Description: Apt repository for Cockpit
 SignWith: default
 EOF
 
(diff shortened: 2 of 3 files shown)


stefwalter:
Can we make this a separate script? Why does it need to be in the release-debian script?

larsu:
> Can we make this a separate script? Why does it need to be in the release-debian script?

We could, but do we ever want to publish the debian repository but not the ppa? They depend on each other anyway.

stefwalter:
> We could, but do we ever want to publish the debian repository but not the ppa? They depend on each other anyway.

 It should be easy to remove one, repeat one, test one without redoing the other.

Yup, there's quite a few release-xxx scripts like that here. Given the $TAG an release-ppa script could do the right thing, when run after release-debian, no?

larsu:
> Yup, there's quite a few release-xxx scripts like that here. Given the $TAG an release-ppa script could do the right thing, when run after release-debian, no?

It would need the location of the built source packages as well, but I guess we can just assume that to be in the base directory?

I'll make it a separate script.

stefwalter:
> It would need the location of the built source packages as well, but I guess we can just assume that to be in the base directory?

That works.

PR #31, opened 2016-06-29
Verify update

dperpeet:


verify/README
```diff
@@ -31,7 +39,17 @@ update github status:
     /home/cockpit/.ssh/id_rsa
     /home/cockpit/.ssh/id_rsa.pub
     /home/cockpit/.config/github-token
-    /home/cockpit/.config/github-whitelist
```
mvollmer: Why is this removed?

dperpeet: Right, I thought we didn't use it anymore. I'll change the docs.

PR #32, opened 2016-08-09
sink: Add option to prune logs

dperpeet:
Any logs older than 30 days that aren't release logs will be
deleted. Their log summary (log) and view file (log.html) will
be archived and moved to an "archive" subdirectory.

Can be run with `./sink/sink -p test`

This is meant to be run as a timed job and will quit after pruning.

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -2,10 +2,12 @@
 
 # Rough configuration file follows, place it in ~/.config/sink
 
+# PruneInterval [days] defaults to 0 (disabled)
 DEFAULTS = """
 [Sink]
 Url: http://%(user)s.fedorapeople.org/logs/%(identifier)s/log
 Logs: ~/public_html/logs
+PruneInterval: 0
 
 [Irc]
 Server: chat.freenode.net:6667
@@ -42,6 +44,7 @@ BADGE_TEMPLATE = """\
 
 import argparse
 import errno
+import fcntl
 import httplib
 import json
 import os
@@ -52,6 +55,7 @@ import string
 import subprocess
 import sys
 import tempfile
+import time
 import traceback
 import urlparse
 import urllib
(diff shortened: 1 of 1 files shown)


stefwalter:
Do you think the sink just be configured to do this automatically on a given system whenever run? Perhaps as a second process (see the IRC code)? Out of interest for more "just works"? What do you think?

dperpeet:
> Do you think the sink just be configured to do this automatically on a given system whenever run? Perhaps as a second process (see the IRC code)? Out of interest for more "just works"? What do you think?

I thought about that, and it might be worth a follow-up.

A second process would be crucial, in my estimation. Also, I would say it had to be opt-in (disabled by default) and throttled via some mechanism, e.g. a file in the log dir whose last-modified is used to track the last run to make sure the script runs ~daily.

Or do you think it's worth removing the command line flags and only implementing the automatic approach, right away, for the sake of simplicity?

stefwalter:
> Or do you think it's worth removing the command line flags and only implementing the automatic approach, right away, for the sake of simplicity?

I think so. Opt in can be configuration in ```"~/.config/sink"``` see ```main()```

dperpeet:
> I think so. Opt in can be configuration in "~/.config/sink" see main()

Yeah exactly, that's what I was thinking of - since the rest is already in there. Note that this pretty much makes the concept of a dry run moot.

dperpeet:
I updated the code to check if pruning is necessary on each startup of the sink.

dperpeet:
I left in the "dry-run" flag (`-d`), I think that's ok.

sink/sink
```diff
@@ -412,17 +417,62 @@ def mkdir_and_chdir(base, identifier):
...
+    entries = os.listdir(base_path)
+    for entry in entries:
+        path = os.path.join(base_path, entry)
+        log_path = os.path.join(path, "log")
+
+        # we only care about directories with our "log" file in them
+        if not os.path.isdir(path) or not os.path.exists(log_path):
+            continue
+
+        # keep release logs
```
stefwalter: This is very Cockpit specific, and this sink is meant to be generic. Can we avoid having these lines? Is there another solution?

sink/sink
```diff
@@ -412,17 +417,62 @@ def mkdir_and_chdir(base, identifier):
...
+        # use the last modified time of the logfile itself, not just the directory
+        last_modified = os.stat(log_path).st_mtime
+
+        diff_days = (reference_time - last_modified) / (60*60*24)
+
+        # keep anything newer than 30 days old
+        if diff_days < 30:
+            continue
+
+        archive_dir = os.path.join(base_path, "archive")
```
stefwalter: Is this meant to delete all attachments and keep the status and log? If so, why don't we just do that? You may have a reason. If so, worth mentioning in the commit log message.

sink/sink
```diff
@@ -412,17 +417,62 @@ def mkdir_and_chdir(base, identifier):
...
+
+        # keep the log summary, ignore nested backup folders
+        archive_file = os.path.join(archive_dir, "{0}-log.tar.xz".format(entry))
+        archive_command = ["tar",
+                           "cJf",
+                           archive_file,
+                           "--directory", base_path,
+                           os.path.join(entry, "log")]
+        # if the log has an html file for parsing, pack that as well for convenience
+        if os.path.isfile(os.path.join(path, "log.html")):
```
stefwalter: Again this is specific to Cockpit, right? This is mentioned in the status file if we need to get it from somewhere more generic.

sink/sink
```diff
@@ -434,6 +484,42 @@ def main():
 
     # Create the directory and chdir
     logs = os.path.expanduser(config.get("Sink", "Logs"))
+
+    # See if we have to prune logs
+    prune_interval = config.get("Sink", "PruneInterval")
+    if prune_interval >= 0:
+        status_file = os.path.expanduser(config.get("Sink", "LogsPruneStatus"))
```
stefwalter: Does this need to be configurable? Can it just be a '.prune-stamp' file in the top level directory?

sink/sink
```diff
@@ -434,6 +484,42 @@ def main():
...
+        prune_now = True
+        # use the last modified time of the logfile itself, not just the directory
```
stefwalter: We should break this stuff out into its own function, because this function is long enough.

dperpeet:
Updated and removed the entire concept of saving/archiving files.
That concept belongs somewhere else.

dperpeet:
to test the sink, I added `-d` to the test-sink script

sink/sink
```diff
@@ -412,17 +416,76 @@ def mkdir_and_chdir(base, identifier):
...
+            reference_time = time.time()
+            last_modified = os.stat(status_file).st_mtime
+            diff_days = (reference_time - last_modified) / (60*60*24)
+            prune_now = diff_days >= prune_interval
+
+        # if we should prune, fork so we don't block the sink
+        if prune_now:
+            if not os.path.isdir(logs):
+                sys.stderr.write("warning: log directory {0} doesn't exist, nothing to do\n".format(logs))
+            elif os.fork() == 0:
```
mvollmer: What of the parent exits before the child?  Wouldn't the child be killed when the ssh session ends?  Do we care?

sink/sink
```diff
@@ -412,17 +416,76 @@ def mkdir_and_chdir(base, identifier):
...
+        if dry_run:
+            sys.stderr.write("Pruning {0} (age: {1} days)\n".format(path, int(round(diff_days))))
+        else:
+            shutil.rmtree(path)
+
+def check_prune(config, logs, dry_run):
+    """ See if we have to prune logs, and if we should, fork and do so
+    """
+    prune_interval = config.get("Sink", "PruneInterval")
+    if prune_interval >= 0:
```
mvollmer: `prune_interval > 0`, no?  Zero means no pruning.

sink/sink
```diff
@@ -412,17 +416,76 @@ def mkdir_and_chdir(base, identifier):
...
 
+def prune_logs(base_path=".", dry_run=False):
+    """ Check all subdirectories of the current path and delete the ones that
+        are either empty or contain only files older than the age threshold.
+
+        For dry runs, print the name of the directory instead.
+    """
+    reference_time = time.time()
+    age_threshold = 30
+    entries = os.listdir(base_path)
```
mvollmer: This will include files, such as the `.prune` stamp file.  I think we should skip non-directories.

sink/sink
```diff
@@ -412,17 +416,76 @@ def mkdir_and_chdir(base, identifier):
...
+
+        if dry_run:
+            sys.stderr.write("Pruning {0} (age: {1} days)\n".format(path, int(round(diff_days))))
+        else:
+            shutil.rmtree(path)
+
+def check_prune(config, logs, dry_run):
+    """ See if we have to prune logs, and if we should, fork and do so
+    """
+    prune_interval = config.get("Sink", "PruneInterval")
```
mvollmer: prune_interval is a string here, and Python will happily compare strings to numbers... so `int(config.get("Sink", "PruneInterval"))`?

sink/sink
mvollmer: The os.walk call below doesn't work for files anyway, I think, and they will always be treated the same as empty directories.

sink/sink
```diff
@@ -412,17 +416,76 @@ def mkdir_and_chdir(base, identifier):
...
+            diff_days = (reference_time - last_modified) / (60*60*24)
+            prune_now = diff_days >= prune_interval
+
+        # if we should prune, fork so we don't block the sink
+        if prune_now:
+            if not os.path.isdir(logs):
+                sys.stderr.write("warning: log directory {0} doesn't exist, nothing to do\n".format(logs))
+            elif os.fork() == 0:
+                # update the file for writing and lock it (prevent concurrent pruning)
+                fd = os.open(status_file, os.O_WRONLY | os.O_CREAT)
```
mvollmer: This doesn't actually update the mtime, so once the prune_interval is reached, we would prune on every run.  Adding O_TRUNC seems to help.

sink/sink
mvollmer: Or even `float(...)`.  Then one can prune twice a day!

mvollmer:
Seems to work, with the proposed changes.

sink/sink
dperpeet: > What of the parent exits before the child? Wouldn't the child be killed when the ssh session ends? Do we care?

You're right, good catch. As we discussed on irc, I think we can live with this. We expect our tests to run longer than the pruning, and even if the pruning gets interrupted, we aren't left with a bad state. I added a comment to that end in the code.

sink/sink
dperpeet: yes

sink/sink
dperpeet: done, skipping everything that's not a directory

sink/sink
dperpeet: float is good

sink/sink
dperpeet: done

dperpeet:
changed

dperpeet:
added a generous exception handler to the pruning loop: if anything goes wrong, silently ignore the error and try the next dir entry

PR #33, opened 2016-08-18
release: install which

larsu:
pbuilder uses which(1) to find debootstrap.

I've already used this to update the base and infra release images.


stefwalter:
Wow. which is not installed.

larsu:
Ya. I don't know how this worked before. Maybe we can update the container at some point? It's still on fedora 22.

stefwalter:
Image is updated to fedora:23 and pushed. You need to ```sudo make base-container``` if you want to reproduce.

PR #37, opened 2016-09-20
Verify: guestfs fixes

dperpeet:


verify/Dockerfile
```diff
@@ -5,11 +5,13 @@ RUN dnf -y install git yum-utils npm rpm-build curl libvirt-client openssl rpmde
          rpm-build krb5-workstation python-lxml expect selinux-policy-devel tar bzip2 python bind-utils \
          nc gnupg freetype fontconfig krb5-workstation expect psmisc mock libvirt \
          libvirt-python sudo python-libguestfs net-tools virt-install
+RUN dnf -y install kernel grubby libguestfs-xfs
```
petervo: Is there a reason not to run all the dnf commands on one line? Will help make the image smaller

dperpeet: Good point. I thought this would be good to leverage layers so people could reuse the older layers, but I realized that with new package versions this layer will change anyway.

dperpeet:
Moved the package installation into a single line and added clean commands

PR #38, opened 2016-10-07
base: update to fedora 24

dperpeet:
Also order the list of included packages, remove duplicates and combine package installation operations into single Docker commands to reduce layers and container image size.


dperpeet:
with these changes, the base image is roughly 30 MiB smaller (4.5%), the verify image 60 MiB (2.5%).

PR #39, opened 2016-10-11
debian: always fetch full package list

larskarlitski:
Fetching Packages.diff (the default) seems to be broken currently.

This should make the debian build work again. I have tested this locally, but haven't pushed a new release container with the change.

### DIFF
--- a/release/release-debian
+++ b/release/release-debian
@@ -72,7 +72,10 @@ prepare()
 
     trace "Creating Debian repository"
 
-    printf "USENETWORK=yes\n" > pbuilderrc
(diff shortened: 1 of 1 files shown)


larskarlitski:
This should also make the ubuntu release work again: [#5093](https://github.com/cockpit-project/cockpit/issues/5093).

dperpeet:
let's wait to test this on the actual release before merging

larskarlitski:
I've just used this to update the debian repository and ubuntu ppa for 119. I think it's good to go.

PR #41, opened 2016-10-14
sink: Update code for Python 2 and 3 usage

ashcrow:


stefwalter:
Merged into master. But had to add a unicode prefix to the DEFAULTS string.

stefwalter:
Thanks @ashcrow

PR #43, opened 2016-10-31
release: Clean up Dockerfile

dperpeet:
Make list of installed packages easier to read and maintain,
combine all install commands into a single layer and clean
up dnf cache when done.


dperpeet:
cockpit/infra-release:26 was built using these changes and the resulting image is smaller (>400 MiB unpacked)

PR #44, opened 2016-11-02
release: Add a release-source and release-srpm commands

stefwalter:
This command generates a patchset that patches an entire tarball including generated files in that tarball, like autoconf/automake or dist/ style generated files.
    
The command can either read from a git repository, or be provided with explicit tarball input.


stefwalter:
A follow on commit will add release/release-srpm which basically takes the non-koji parts of release/release-koji into its own file ... and it'll drive release/release-patchset as necessary. I already have this second part in progress.

stefwalter:
@dperpeet Okay this is done.

The release-source produces patches that are bigger than necessary. This should trend downwards by applying fixes to the Cockpit git repository, such as the LANG fix ... and upcoming .gitattributes fixes.

stefwalter:
I have a simple idea for reducing the diffs somewhat: First try rebuilding for a new commit without clearing out _build. If it then fails, clear out _build and try again.

stefwalter:
Added an optimization by not clearing out the build directory completely for each patch ... see release-source.

PR #45, opened 2016-11-07
release: Recognize multiple source tarballs

stefwalter:
Produce and recognize multiple source tarballs. In particular.
    
 * release-source now creates a directory with output
 * release-github now uploads multiple tarballs
 * RELEASE_SOURCE environment variable
 * Recognize source directory in many places


stefwalter:
Folded into #44

PR #49, opened 2016-11-09
release-source fixes

stefwalter:
* Move tag to patched commit during release-source
 * Complain about missing source in release-source


dperpeet:
Worked when we tested this locally.

PR #52, opened 2016-11-11
push release container images to Docker Hub

dperpeet:
also tag newest images with iso date instead of sequential numbering

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,24 @@ release-container: docker-running
 	docker rm -f cockpit-release-stage
 	@true
 
+release-push: docker-running
+	{ \
+	ID=`docker images -q cockpit/infra-release:latest`; \
+	if [ `echo "$$ID" | wc -w` -ne "1" ]; then \
+		echo "Expected exactly one image matching 'cockpit/infra-release:latest'"; \
+		exit 1; \
+	fi; \
+	TAGS=`docker images --format "table {{.Tag}}\t{{.ID}}" | grep $$ID | awk '{print $$1}'`; \
+	if [ `echo "$$TAGS" | wc -w` -ne "2" ]; then \
+		echo "Expected exactly two tags for the image to push: latest and one other"; \
+		exit 1; \
+	fi; \
+	for TAG in $$TAGS; do \
+		docker push "cockpit/infra-release:$$TAG"; \
+	done \
+	}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,9 +25,16 @@ release-container: docker-running
 	docker rm -f cockpit-release-stage || true
 	docker run --privileged --name=cockpit-release-stage \
 		--entrypoint=/usr/local/bin/Dockerfile.sh cockpit/infra-release:staged
+	$(eval tag := cockpit/infra-release:$(shell date --iso-8601))
```
stefwalter: May I suggest moving this line outside of the rule, above the rule and using an upper case variable name? Typically make variables are upper case:

```
TAG := cockpit/infra-release:$(shell date --iso-8601)

release-container:
       ...
```

release/release-push
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh -euf
+#
+# release-push
```
stefwalter: The scripts named 'release-xxxx' in this directory are meant to be used during the release process. Lets either rename this script, or move it up a level.

It's a bummer that we need such a complicated script for a simple push  task though. Oh well.

Makefile
dperpeet: I didn't want to declare it global since it's only used there, but sure.

dperpeet: done

release/release-push
dperpeet: moved the contents of the script into the Makefile

Makefile
```diff
@@ -20,14 +20,35 @@ release-shell: docker-running
...
+	@true
+
+release-push: docker-running
+	{ \
+	ID=`docker images -q cockpit/infra-release:latest`; \
+	if [ `echo "$$ID" | wc -w` -ne "1" ]; then \
+		echo "Expected exactly one image matching 'cockpit/infra-release:latest'"; \
+		exit 1; \
+	fi; \
+	TAGS=`docker images --format "table {{.Tag}}\t{{.ID}}" | grep $$ID | awk '{print $$1}'`; \
```
dperpeet: this might seem a bit awkward, but I wanted to make sure that we get a known output format when scraping the image info

stefwalter:
Tested and merged into master.

PR #53, opened 2016-11-11
release-debian: generate empty i386 Packages file

larskarlitski:
`apt-get update` tries to find package listings for all architectures on
multi-arch systems. (It probably shouldn't do that when the repository
only advertises one architecture.)

To work around this in the most common case of having i386 configured on
an amd64 system, generate an empty Packages file for i386.


stefwalter:
Will merge this during the next release ... I have a bunch more debian changes that are slated to land.

PR #57, opened 2016-11-19
release: Support multiple distributions in release-debian

stefwalter:
Support multiple distributions in release-debian and start building both unstable and jessie distributions during the release.

Depends on:

* [x] #59 
* [x] #58 
* [x] #52

PR #58, opened 2016-11-20
release: Use a distribution specific basetgz for pbuilder

stefwalter:
Use a distribution specific basetgz as a pbuilder cache. This will be followed up by work on building on various distributions.

### DIFF
--- a/release/release-debian
+++ b/release/release-debian
@@ -59,7 +59,7 @@ changelog_lines()
 
 prepare()
 {
-    local installed
+    local installed basetgz
     local debversion="$TAG-0"
 
     trace "Creating Debian repository"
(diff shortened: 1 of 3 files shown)


stefwalter:
Built a new container based on this and will push it for further debugging of #59

release/release-debian
```diff
@@ -99,11 +99,11 @@ EOF
         reprepro -b $DEBIAN_REPO includedsc unstable cockpit_$debversion.dsc
 
         # build binary package
-        sudo pbuilder clean
-        sudo pbuilder update --configfile pbuilderrc --distribution unstable
-        sudo pbuilder build --configfile pbuilderrc \
-                            --hookdir /etc/pbuilder/hooks \
-                            cockpit_$debversion.dsc
+        basetgz=/var/cache/pbuilder/unstable.tgz
```
dperpeet: If we want to release multiple versions in the future, then we might have two different `basetgz` here - what about calling this one `basetgz_unstable` to match the dist settings in the following lines?

stefwalter: See the following pull request. This goes into a loop.

stefwalter: See the following pull request #57 this goes into a loop. #57 is not working completely yet. But that part with the basetgz= is working.

dperpeet: Ok, that makes sense - this pull request was half-way ready for a loop, which made it appear strange. Improving it in a follow-up is acceptable.

PR #59, opened 2016-11-21
Check out from github on 'make release-shell' if necessary

stefwalter:


Makefile
```diff
@@ -14,6 +14,8 @@ containers: release-container verify-container
 	@true
 
 release-shell: docker-running
+	test -d /home/release/cockpit || git clone https://github.com/cockpit-project/cockpit /home/cockpit/release
```
dperpeet: shouldn't this be `test -d /home/cockpit/release`?

stefwalter: Indeed it should.

stefwalter: Done.

PR #60, opened 2016-11-28
sink: Remove badge support from sink

stefwalter:
This is a workaround for the fact that there was no service hosting
image badges for specific github statuses. We no longer use this.

If anyone ever does need this again, the correct way to solve this
is to host a service that looks up the GitHub status for master
and produces an image based on it.

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -15,34 +15,9 @@ PruneInterval: 0
 Server: chat.freenode.net:6667
 Login: %(user)sbot %(user)sbot %(user)sbot
 Nick: %(user)sbot
-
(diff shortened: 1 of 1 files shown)


stefwalter:
CC @mvollmer @vpodzime @dperpeet  I don't want to have to support the badging code in the sink, since it's the wrong place for it. Since others are starting to use the sink, I'd like to remove this part.

vpodzime:
Fine with me, thanks for letting me know!

mvollmer:
👍

stefwalter:
Merged.

PR #61, opened 2016-12-14
Small fixes in srpm creation script

nmav:


stefwalter:
@dperpeet Do you use this too outside of Cockpit upstream releases? If so, I'd ask for a quick look to make sure it doesn't break you.

I'm testing this during 126

dperpeet:
> @dperpeet Do you use this too outside of Cockpit upstream releases? If so, I'd ask for a quick look to make sure it doesn't break you.

looked but forgot to reply, this doesn't seem to break anything and the changes look good

stefwalter:
Thanks! Merged.

PR #62, opened 2016-12-14
Add the %patch directive

nmav:


nmav:
This applies on top of #61

stefwalter:
This doesn't work with binary patches. Hence in our cockpit spec file we use something like this:

https://github.com/cockpit-project/cockpit/blob/master/tools/cockpit.spec#L147

I like the idea of generalizing this ... but it needs to support binary patches.

nmav:
Is there a bug on fedpkg/rpmbuild for that? I do not think that this component is the right one to address this issue. Alternatively if rpmbuild will not fix that we can do the (quite ugly) hack with git as well in that patch.

stefwalter:
The Fedora RPM macros have a simple macro fix for applying all the listed patches as binary patches. The Fedora RPM macro uses git. The above usage of git in tools/cockpit.spec is simply an expansion of that macro so that the same spec file works on RHEL.

nmav:
How would you suggest to address it? Use git unconditionally?

stefwalter:
> How would you suggest to address it? Use git unconditionally?

That's what we've been forced to do. But I'm open to alternative ideas.

The goal here is to make packaging be an automated activity ... and given a) binary patches and b) lack of RHEL RPM macros for them ... we haven't found an alternative to using git directly in the spec file yet.

stefwalter:
@nmav I wouldn't be against making this work conditional on an optional flag. I generally try to shy away from too many different options or ways to fail ... but we could make an exception here.

nmav:
Applying patches was made conditional

stefwalter:
Merged into git master.

PR #64, opened 2017-01-09
release-koji: fix git ls-tree invocation

nmav:
git ls-tree prints multiple tabs by default, and the current
invocation assumed that only the file name is present. Ensure
that only the file name is present by appending --name-only.


stefwalter:
Merged into git master.

PR #66, opened 2017-01-18
release-debian: Remove PCP build dependencies while pcp is not in testing

martinpitt:
pcp has release-critical bugs and fell out of testing, see https://tracker.debian.org/pkg/pcp. Drop the pcp build dependencies when building for unstable so that the package can be used on testing/stretch.
    
This requires https://github.com/cockpit-project/cockpit/pull/5726 so that the package actually builds without these build dependencies.

### DIFF
--- a/release/release-debian
+++ b/release/release-debian
@@ -121,6 +121,11 @@ EOF
         dpkg-source --extract $DSC $distribution
         sed -i "s/$TAG-0/$TAG-0~$distribution/" $distribution/debian/changelog
         sed -i "s/experimental/$distribution/" $distribution/debian/changelog
+        # Hack: Remove PCP build dependencies while pcp is not in testing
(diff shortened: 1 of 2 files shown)


stefwalter:
I'll test and merge this as part of the 129 release. Does it make sense to have this as WIP or is it ready for use for this one release?

martinpitt:
No, it doesn't work right now -- if you drop the build deps the package will fail to build. This is fixed in https://github.com/cockpit-project/cockpit/pull/5726 . This will hopefully land RSN, it's blocked by some infra problems on the tests (failing to connect to github.com when downloading the JS modules for building the package).

stefwalter:
Okay, #5726 is merged. All related debian-8 tests passed and I merged it. So this is unblocked now, right?

martinpitt:
Ah, correct.

PR #72, opened 2017-01-31
verify: Add initial playbooks from Ari

stefwalter:
These are playbooks that setup a verify machine. And meant to be used
with Jenkins. They may not work perfectly at this point, but bringing
them in here is the first step.


stefwalter:
@dperpeet Any objections to merging? These don't need to be perfect yet, but bringing them here helps us make progress on using them.

dperpeet:
Yeah, let's get these in here and fixup as needed.

PR #73, opened 2017-02-01
Use the ./config/github-token also for GH requests

vpodzime:
Otherwise it always has to be passed as part of the JSON prepended/appended to
the logs which makes it even more exposed.


stefwalter:
Running this on our fedorapeople.org instance. Will run it there until later today.

stefwalter:
Merged into master.

PR #74, opened 2017-02-07
release: Don't clean dsc source package

martinpitt:
With the addition of the dh-systemd build dependency we cannot run the
clean target under Fedora any more, as dh-systemd is not packaged and
debhelper is too old (dh_systemd got merged in 10.1 only). However,
there is no need for it -- we create a clean source anyway, so there is
nothing to clean. So skip cleaning (-nc), and also use -d to skip the
build dependency check to avoid some noise.


stefwalter:
Thanks!

PR #75, opened 2017-02-13
release-srpm: copy additional files in the srpm dir

nmav:
That is, any additional files available in the SOURCE_DIR will be
copied in the working dir of SRPM creation. That allows to copy
files such as gpg keys, or configuration files that are only provided
downstream.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@redhat.com>

### DIFF
--- a/release/release-srpm
+++ b/release/release-srpm
@@ -187,14 +187,20 @@ prepare()
     printf "%%changelog\n" >> $logfile
     changelog_lines $TAG $RELEASE >> $logfile
 
-    # Build up the list of patches
-    count=1
-    find $SOURCE -maxdepth 1 -name '*.patch' | while read patch; do
-        cp "$patch" $workdir
-        changelog_patch "$patch" >> $logfile
-        printf "Patch%s:\t%s\n" $count "$(basename $patch)" >> $specfile
-        count=$(expr $count + 1)
-    done
+    # copy additional files available in source dir
+    # and build up the list of patches
+    pcount=1
+    if test -d "$SOURCE";then
+	    find $SOURCE -maxdepth 1 -type f | while read file; do
+	        cp "$file" $workdir
+
+	        if test "$file" != "${file%patch*}" ; then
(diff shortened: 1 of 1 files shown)


nmav:
Note, that I am unsure about this change and whether it is within the logic of the cockpituous. What I needed in a certain project was to have some downstream configuration files that will be included in the SRPM, and couldn't figure any other way to do that.

release/release-srpm
```diff
@@ -187,6 +187,13 @@ prepare()
     printf "%%changelog\n" >> $logfile
     changelog_lines $TAG $RELEASE >> $logfile
 
+    # copy additional files available in source dir
+    if test -d "$SOURCE";then
+	    find $SOURCE -maxdepth 1 -type f | while read file; do
+	        cp "$file" $workdir
```
stefwalter: Does this mean that the line below to copy the patch file is now superfluous? If so, could you adapt the following lines so that the patches are enumerated from $workdir instead?

nmav: What about this approach, handling them all in one pass?

stefwalter: Yup, that works for me ... as long as we only handle *.patch as patches.

release/release-srpm
```diff
@@ -187,14 +187,20 @@ prepare()
...
+
+	        if test "$file" != "${file%patch*}" ; then
+		    changelog_patch "$file" >> $logfile
+		    printf "Patch%s:\t%s\n" $pcount "$(basename $file)" >> $specfile
+		    pcount=$(expr $pcount + 1)
+	        fi
+	    done
+    fi
```
nmav: That should do it.

stefwalter:
Okay. I'll be testing this with the 133 or later release of Cockpit (at least a week from now). If I forget, please ping me.

stefwalter:
Cockpit 133 was released with these changes. So merging.

PR #78, opened 2017-02-14
release-koji: corrected the way patches are being added

nmav:
Previously 'rpm -q --specfile --queryformat' was used, but this
command does not list the patches in the spec file. The current
approach uses cut and sed.


stefwalter:
Made some changes to the naming of the functions and spacing and will be running this through a cockpit release.

stefwalter:
I had to make some changes so this wouldn't fail when no patches or sources were present. See the commit above.

PR #79, opened 2017-02-24
release: No longer build Fedora 24 releases in Koji

stefwalter:
It's been a while since we've pushed into Fedora 24. There's no reason to do Fedora 24 releases in Koji any more.
    
We still do Fedora 24 releases in COPR.


stefwalter:
@martinpitt Could you review this? There hasn't been a release with it, but it seems pretty straight forward.

PR #80, opened 2017-02-24
release: Remove hack related to cockpit-test-assets

stefwalter:
Cockpit is now bundling more and more tests in a cockpit-tests
package. Lets remove this hack while building the DSC.


stefwalter:
@martinpitt Could you review this? The 132 release was done with this change, so it has been tested.

martinpitt:
Merging manually to add the Closes #

PR #81, opened 2017-03-02
release-srpm: Fix parsing of source directory vs. tarball

stefwalter:
Make sure all the combinations of how these are specified are
handled.
 1. If a directory is specified but no tarball, then find the
    tarball in that directory
 2. If a tarball is specified but no source directory, then
    treat the source directory as empty.
 3. If source is specified then load patches from source directory.

### DIFF
--- a/release/release-srpm
+++ b/release/release-srpm
@@ -13,8 +13,9 @@
 #
 # Arguments are described here. Most arguments have an equivalent envvar.
 #
-# -f tgz     RELEASE_SOURCE=dir     The tarball to include or directory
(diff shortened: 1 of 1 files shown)


stefwalter:
@nmav Could you also take a look and make sure this doesn't break your use case?

stefwalter:
> Not sure if they are supposed to, but if not then the comment at the top should be updated.

Updated documentation.

nmav:

martinpitt:
Thanks @nmav for confirming! Unblocking.

PR #82, opened 2017-03-02
release-srpm: Fix parsing version/tag from tarball

stefwalter:
When -t or $RELEASE_TAG is not specified we parse the version
number from teh tarball. We should be greedy and parse until
the first non-numeric character.


stefwalter:
Okay, this is invalid. The user can specify the tag with -t or RELEASE_TAG. Ideally we would fix this to check that the tarball matches the expected source tarball.

PR #83, opened 2017-03-02
release-srpm: Load spec file from either tarball or working directory

stefwalter:
Often a tarball will include the spec file. We should support loading
it from there and look for it there first.

### DIFF
--- a/release/release-srpm
+++ b/release/release-srpm
@@ -162,18 +162,21 @@ prepare()
         fi
     fi
 
+    workdir=$(mktemp --directory srpm.XXXXXX)
+    specfile=$workdir/$(basename $SPEC)
+    logfile=$workdir/changelog
+
+    # Look for the spec file in the tarball or the cwd
+    (tar -xvf $TARBALL */$SPEC -O 2> /dev/null || cat $SPEC) > $specfile.orig
+
     # If our version is already in the spec file then just bump the version
     # The revision comes from the spec file
-    RELEASE=$(calc_release $TAG $SPEC)
+    RELEASE=$(calc_release $TAG $specfile.orig)
 
     if [ -z "$RELEASE" ]; then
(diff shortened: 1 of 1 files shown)


stefwalter:
@nmav Could you also review this?

nmav:
I am testing that on the new gnutls release. About the patch however, shouldn't that be an option on whether to use the included in the tarball spec? I can think a lot of tarballs which contain a spec which we don't necessarily use for releases.

martinpitt:
@nmav: Out of interest, what would be an use case for that? Wouldn't that essentially mean that your release tarballs ship a broken spec file?

nmav:
@martinpitt think that you are packaging a program which ships opensuse rpms? The upstream test suite may be sufficient to push it to fedora using cockpituous, but you don't want to use the opensuse spec.

martinpitt:
@nmav: So if you have different spec files for suse and for fedora, you would pass the path to the one you want, regardless of whether that's in git or the tar? (Again: I'm not denying that there is an use case for that, I'm just trying to understand it).

Note that you can force this to use the "outside" spec file by specifying an absolute path (or any path which isn't contained in the tar).

nmav:
> @nmav: So if you have different spec files for suse and for fedora, you would pass the path to the one you want, regardless of whether that's in git or the tar? (Again: I'm not denying that there is an use case for that, I'm just trying to understand it).
> Note that you can force this to use the "outside" spec file by specifying an absolute path (or any path which isn't contained in the tar).

My main point is that the semantics of this change are complex (and error prone for the case where the tarball is from a 3rd party and may contain a spec file). It is correct that one can make it work right, however you must know how the script works internally (what you describe above is not the most obvious usage of the script).

nmav:

martinpitt:
Thanks @nmav for testing! Unblocking.

PR #84, opened 2017-03-02
release-srpm: Fix undefined variable tagdate if not a git checkout

stefwalter:


stefwalter:
The ```cockpit/infra-release``` image that was used to release 133 includes this change.

martinpitt:
Not sure why this was "blocked" -- it's quite obvious, not prone to regress non-cockpit use cases, and was successfully tested in production.

PR #85, opened 2017-03-02
release-srpm: Resolve tarball symlinks early

stefwalter:
For tarballs that are symlinks to an actual tarball, we resolve
those before trying to perhaps trying to parse a tag from them.

### DIFF
--- a/release/release-srpm
+++ b/release/release-srpm
@@ -153,6 +153,9 @@ prepare()
 {
     local specfile logfile count workdir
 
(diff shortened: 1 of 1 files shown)


stefwalter:
@nmav Could you also review this?

stefwalter:
The ```cockpit/infra-release``` image that was used to release 133 includes this change.

nmav:

martinpitt:
Thanks @nmav for testing! Unblocking.

PR #86, opened 2017-03-07
release-dsc: Copy patches if requested

martinpitt:
Add a $RELEASE_PATCHES env var and -P option similar to release-srpm.
When enabled, copy our git  patches to debian/git-patches/. Do not create
a quilt series, as quilt is not able to apply git's binary patches.
Instead, debian/rules will be updated to apply them with git, same way
as in cockpit.spec (https://github.com/cockpit-project/cockpit/pull/6052).

Also fix the documentation comment about -s:  This needs to specify the
path to the "control" file, not the debian directory.

### DIFF
--- a/release/release-dsc
+++ b/release/release-dsc
@@ -2,20 +2,22 @@
 #
 # release-dsc
 #
-# A script that creates a Debian DSC based on a tarball.
+# A script that creates a Debian DSC based on a tarball and patches.
 # The path to the debian directory is specified on the command line.
-# The Debian source files are updated with version information as well
-# as changelog entries.
+# The Debian source files are updated with version information, patches,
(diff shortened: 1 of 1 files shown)


martinpitt:
For the record (and public ridicule if it breaks): I tested  this with current master, where `RELEASE_SOURCE=out release-source` produces some 25 patches on top of 233. I then ran `RELEASE_CONTROL=tools/debian/control RELEASE_SOURCE=out RELEASE_PATCHES=1  ../cockpituous/release/release-dsc` and ensured that all patches get applied and the package builds. I also re-ran this without `RELEASE_PATCHES` and made sure that the .dsc did not get any debian/git-patches/ and that the package build still works.

PR #88, opened 2017-03-10
release: Build for fedora-26

dperpeet:
We don't want to push new packages into fedora 25 anymore.

### DIFF
--- a/release/major-cockpit-release
+++ b/release/major-cockpit-release
@@ -18,6 +18,7 @@ job release-srpm
 # Do fedora builds for the tag, using tarball
 job release-koji -k master
 job release-koji f25
+job release-koji -k f26
 
 # Upload release to github, using tarball
 job release-github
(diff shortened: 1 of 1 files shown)


mvollmer:
We don't test yet on Fedora-26, right?  Should we wait with releasing to it until we have at least one green test run?

dperpeet:
> We don't test yet on Fedora-26, right? Should we wait with releasing to it until we have at least one green test run?

Your comment made me realize that we don't want to replace the Fedora 25 builds. That said, we still want to build on Fedora 26 and push our updates. Bodhi is for testing, so nothing goes in automatically. Since Fedora-26 has forked, we need to use Bodhi to actually push newer packages out there for people to consume. It's still pre-alpha.

stefwalter:
@mvollmer is right, ideally we want to test on Fedora 26 before we release there ... although at the point where 26 branched from rawhide whatever build is already there. So the situation doesn't get worse. Not a good place to be though. We're very late on testing on Fedora 26.

stefwalter:
This is blocked on being actually used in at least one release. @dperpeet already uploaded the container, and now we run one release before merging it into master.

dperpeet:
worked during the 136 release

martinpitt:
The [136 release](https://fedorapeople.org/groups/cockpit/logs/release-136/log) produced nice [Fedora 26 packages](https://koji.fedoraproject.org/koji/taskinfo?taskID=18550685), so considering sufficiently tested.

PR #90, opened 2017-03-15
Add a Makefile 'verify-push' target

stefwalter:


Makefile
```diff
@@ -59,15 +60,33 @@ release-install: release-container
 
 verify-shell: docker-running
 	docker run -ti --rm \
-		--privileged \
 		--volume /home/cockpit:/home/user \
 		--volume $(CURDIR)/verify:/usr/local/bin \
 		--volume=/opt/verify:/build:rw \
-		--net=host --pid=host --privileged --entrypoint=/bin/bash \
+		--net=host --pid=host --entrypoint=/bin/bash \
```
martinpitt: Was the intent really to drop *both* `--privileged` options, or just drop the redundant one? When dropping both, I think that at least warrants a detailled commit message or a separate commit. Fine for me to do an en passant cleanup of just one of them.

Makefile
```diff
@@ -59,15 +60,33 @@ release-install: release-container
...
+	if [ `echo "$$ID" | wc -w` -ne "1" ]; then \
+		echo "Expected exactly one image matching 'cockpit/infra-verify:latest'"; \
+		exit 1; \
+	fi; \
+	TAGS=`docker images --format "table {{.Tag}}\t{{.ID}}" | grep $$ID | awk '{print $$1}'`; \
+	if [ `echo "$$TAGS" | wc -w` -ne "2" ]; then \
+		echo "Expected exactly two tags for the image to push: latest and one other"; \
+		exit 1; \
+	fi; \
+	for TAG in $$TAGS; do \
```
martinpitt: Slightly confusing to clobber the global `$TAG` variable here. (Just noticing, keep it if you like).

stefwalter:
This is something that was fixed by Dominik.

PR #92, opened 2017-03-16
release: Make release-source more verbose

stefwalter:
We need to be able to better diagnose build failures.

### DIFF
--- a/release/release-source
+++ b/release/release-source
@@ -37,6 +37,8 @@ VERBOSE=${RELEASE_VERBOSE:-0}
 SOURCE=${RELEASE_SOURCE:-}
 TAG=${RELEASE_TAG:-}
 
+MAKE="make V=1"
+
 # When building tarballs it's important to have a consistent
 # locale and charmap in particular when generating documentation
 LANG=en_US.UTF-8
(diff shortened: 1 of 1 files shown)


stefwalter:
This needs testing during a release run.

release/release-source
```diff
@@ -248,6 +250,8 @@ shift $(expr $OPTIND - 1)
 
 if [ $VERBOSE -eq 1 ]; then
     set -x
+elif [ $QUIET -eq 0 ]; then
+    MAKE="make --silent"
```
martinpitt: This looks the wrong way around -- should't we use `--silent` if `$QUIET` is true, i. e. `-eq 1`, not 0?

release/release-source
```diff
@@ -140,7 +142,7 @@ prepare()
                 trace "Creating first tarball"
 
                 autogen_or_configure $repodir
-                make -C $repodir/_build --silent dist
+                $(MAKE) -C $repodir/_build dist
```
martinpitt: and this needs to be `$MAKE`, this isn't a Makefile (confusingly ☺).

martinpitt:
136 was released with this patch in our container, and the log is verbose: https://fedorapeople.org/groups/cockpit/logs/release-136/log

So I consider this sufficiently tested.

PR #94, opened 2017-04-21
release: Ensure dnf knows builddep

dperpeet:
We need that plugin to install the cockpit build dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,29 @@ all:
 docker-running:
 	systemctl start docker
 
+TAG := $(shell date --iso-8601)
+
 base-container: docker-running
-	docker build -t cockpit/infra-base base
+	docker build -t cockpit/infra-base:$(TAG) base
+	docker tag cockpit/infra-base:$(TAG) cockpit/infra-base:latest
+
+base-push: docker-running
+	{ \
+	ID=`docker images -q cockpit/infra-base:latest`; \
+	if [ `echo "$$ID" | wc -w` -ne "1" ]; then \
+		echo "Expected exactly one image matching 'cockpit/infra-base:latest'"; \
+		exit 1; \
+	fi; \
+	TAGS=`docker images --format "table {{.Tag}}\t{{.ID}}" | grep $$ID | awk '{print $$1}'`; \
+	if [ `echo "$$TAGS" | wc -w` -ne "2" ]; then \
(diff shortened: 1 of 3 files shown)


dperpeet:
I needed to add this to build a new container

Makefile
```diff
@@ -7,8 +7,29 @@ all:
...
+	TAGS=`docker images --format "table {{.Tag}}\t{{.ID}}" | grep $$ID | awk '{print $$1}'`; \
+	if [ `echo "$$TAGS" | wc -w` -ne "2" ]; then \
+		echo "Expected exactly two tags for the image to push: latest and one other"; \
+		exit 1; \
+	fi; \
+	for PUSHTAG in $$TAGS; do \
+		docker push "cockpit/infra-base:$$PUSHTAG"; \
+	done \
+	}
+	@true
```
martinpitt: OOI, why the { } and `@true`?

Makefile
```diff
@@ -68,6 +88,25 @@ verify-shell: docker-running
...
+		TAGS=`docker images --format "table {{.Tag}}\t{{.ID}}" | grep $$ID | awk '{print $$1}'`; \
+		if [ `echo "$$TAGS" | wc -w` -ne "2" ]; then \
+			echo "Expected exactly two tags for the image to push: latest and one other"; \
+			exit 1; \
+		fi; \
+		for PUSHTAG in $$TAGS; do \
+			docker push "cockpit/infra-verify:$$PUSHTAG"; \
+		done \
+		}
+		@true
```
martinpitt: So you have push targets for base and verify, why not for release?

Makefile
dperpeet: I think this was necessary for the ID variable and script to work

Makefile
dperpeet: this code is copy/pasted from the `release-push` target

martinpitt:
Added commit to install rpm-build, which broke the current release attempt:
```
> Starting: release-srpm
rpm: /usr/bin/rpmspec: No such file or directory
release-srpm: could not calculate release number from spec file: tools/cockpit.spec
```

dperpeet:
> Added commit to install rpm-build, which broke the current release attempt:

looks good!

martinpitt:
The 139 release succeeded with these changes, so merging now.

PR #96, opened 2017-04-25
verify: Prune less often

dperpeet:
We don't want the verify script to prune images automatically.

The image store should decide when and how to call the pruning scripts.

### DIFF
--- a/verify/cockpit-verify
+++ b/verify/cockpit-verify
@@ -22,12 +22,18 @@ mkdir -p mock
 
 echo "Starting testing"
 
+iteration=0
 while true; do
     fail=0
     find cockpit -name '*.py?' -delete
     git -C cockpit fetch origin
     git -C cockpit reset --hard origin/master
-    cockpit/test/vm-download --prune
+    if [ "$iteration" -eq 0 ]; then
+        cockpit/test/vm-download --prune
+    else
+        cockpit/test/vm-download
+    fi
+    iteration=$(( (iteration+1) % 25 ))
     if ! cockpit/test/github-task "$@"; then
         exit
     fi


stefwalter:
> The image store should decide when and how to call the pruning scripts.

When and where does this happen?

dperpeet:
> When and where does this happen?

Right now there's no standard way, but I think it's best to cancel this generic one for now.
In general I can imagine having a job in the verify container that runs at least on startup and regularly while the container is up. Since we have to check multiple branches, we don't want to do that for every pull request.

stefwalter:
> Since we have to check multiple branches, we don't want to do that for every pull request.

We need to stop pruning images that are on any branch on origin. 

Before merging this, I'd to know that you have an answer on how to prevent manual or machine specific maintenance on every verify machine. That's something we should be moving towards, not away from.

dperpeet:
> Before merging this, I'd to know that you have an answer on how to prevent manual or machine specific maintenance on every verify machine. That's something we should be moving towards, not away from.

In general I can imagine having a job in the verify container that runs at least on startup and regularly while the container is up. Since we have to check multiple branches, we don't want to do that for every pull request. That said, I want to stop deletion of images first and separately from implementing the "good" pruning.

dperpeet:
We don't want the verify script to prune too often.

dperpeet:
Logic to prune more intelligently is in the cockpit repo: https://github.com/cockpit-project/cockpit/pull/6395

verify/cockpit-verify
```diff
@@ -22,12 +22,19 @@ mkdir -p mock
...
 echo "Starting testing"
 
+iteration=0
 while true; do
     fail=0
     find cockpit -name '*.py?' -delete
     git -C cockpit fetch origin
     git -C cockpit reset --hard origin/master
-    cockpit/test/vm-download --prune
+    let prune=iteration%25
```
martinpitt: The script is `#!/bin/sh`, and this is a bashism. I'd recommend to drop this line and `$prune` altogether, use `$iteration` in the next line, and ..

verify/cockpit-verify
```diff
@@ -22,12 +22,19 @@ mkdir -p mock
...
+    if [ "$prune" -eq 0 ]; then
+        cockpit/test/vm-download --prune
+    else
+        cockpit/test/vm-download
+    fi
+    let "iteration++"
```
martinpitt: ... do `iteration=$(( (iteration+1) % 25 ))` here instead, which is POSIX shell.

verify/cockpit-verify
dperpeet: done...

verify/cockpit-verify
dperpeet: and done

dperpeet:
> I thought you wanted to background the pruning? The tests will download missing images themselves, so having a "prune only" mode for vm-download would allow that. Perhaps we could even change the semantics of --prune to be like that? Then it could just be backgrounded here.
> 
> But I'm fine with incremental improvements and merge this approach, after fixing the bashism.

Thanks for the feedback, updated. The way it works now we don't need to download any additional information or even use `checkout` commands. That said, this doesn't take very long and I didn't want to risk git errors when we traverse the tree while our build process also meddles with the tree.

The tests do download images, but that can take a significant amount of time, so it's better to do that before the tests run. If it's called less often, it's even less of a problem.

dperpeet:
> Thanks, this looks good now. Do you want me to merge, or do we handle the verify container like the release container, i. e. you build/deploy a container and we only merge afterwards once it's confirmed to work?

Thanks. I think it's probably better to try this out first, I'll build the container and do that.

stefwalter:
Container built deployed to cockpit-11

PR #98, opened 2017-04-27
verify: Prevent KVM modprobe usage in the verify container

stefwalter:
This makes test/vm-prep work inside of the verify container.


stefwalter:
Otherwise we use rmmod to remove the module, and then modprobe but cannot add it back because it is missing in the container, or of a different kernel version.

PR #99, opened 2017-04-27
Use an isolated network namespace in verify machines

stefwalter:
This helps things be self-contained, reduces deployment
effort, but also helps us work with socket + mcast qemu networks.

It turns out on RHEL 7 socket + mcast in a container doesn't work
properly if sharing the host network namespace. At least on our
RHEL based verify machines it doesn't.

### DIFF
--- a/verify/README
+++ b/verify/README
@@ -21,11 +21,9 @@ First get the container:
 
 Some bits need to be configured outside the container:
 
-    # yum install libvirt libvirt-client qemu-system-x86
-    # wget https://raw.githubusercontent.com/cockpit-project/cockpit/master/test/vm-prep
-    # mkdir common
-    # wget -O ./common/network-cockpit.xml https://raw.githubusercontent.com/cockpit-project/cockpit/master/test/common/network-cockpit.xml
-    # sh vm-prep
+    # echo "options kvm-intel nested=1" > /etc/modprobe.d/kvm-intel.conf
+    # echo "options kvm-amd nested=1" > /etc/modprobe.d/kvm-amd.conf
+    # ( rmmod kvm-intel && modprobe kvm-intel ) || ( rmmod kvm-amd && modprobe kvm-amd )
 
 Setup a 'cockpit' user:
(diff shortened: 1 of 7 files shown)


stefwalter:
Uploaded a new container including this. Running on cockpit-11

stefwalter:
This built into a container and deployed to all the verify machines.

verify/README
```diff
@@ -21,11 +21,9 @@ First get the container:
...
 Some bits need to be configured outside the container:
 
-    # yum install libvirt libvirt-client qemu-system-x86
-    # wget https://raw.githubusercontent.com/cockpit-project/cockpit/master/test/vm-prep
-    # mkdir common
-    # wget -O ./common/network-cockpit.xml https://raw.githubusercontent.com/cockpit-project/cockpit/master/test/common/network-cockpit.xml
-    # sh vm-prep
+    # echo "options kvm-intel nested=1" > /etc/modprobe.d/kvm-intel.conf
+    # echo "options kvm-amd nested=1" > /etc/modprobe.d/kvm-amd.conf
+    # rmmod kvm-intel && modprobe kvm-intel && rmmod kvm-amd && modprobe kvm-amd || true
```
martinpitt: Technically the `&&` between  the intel and amd parts should be a `||`, with appropriate parentheses, as you want to try amd if intel *fails*, not when it succeeds.

verify/vm-prep
```diff
@@ -0,0 +1,125 @@
+#!/bin/sh
```
martinpitt: This and the above XML are by and large a copy of cockpit's files. Why can't we just use these?

stefwalter: Because stuff in cockpit changes between pull requests. #6334 removes these from the cockpit repo. This cockpit1 network is legacy and in the future will only be used for testing non-master branches in the cockpit repo.

verify/README
stefwalter: Will do.

PR #100, opened 2017-05-03
release: Don't exclude aclocal.m4 from patches

dperpeet:
We don't want to miss the changes.


martinpitt:
I built/uploaded a new release container with this and pulled it onto the release machine, so the next release (141) will run with this.

martinpitt:

PR #101, opened 2017-05-03
tests: Openshift support, atomic invocation and container mount points

stefwalter:
Renamed the container to ```cockpit/tests``` to match the cockpit-tests RPM that we are also trying to make. The container can be invoked with the following command, and it will be ready to run the integration tests.
    
     $ sudo atomic run cockpit/tests
    
It can be installed as a service on a system with the following command:
    
     $ sudo atomic install cockpit/tests
    
In addition we define standard mount points documented in README.
    
     * /secrets: A directory containing SSH or github-token secrets
     * /images: An images directory
    
Added a 'make tests-secrets' target which wraps up the necessary secrets both in tarball and in a Kubernetes Secrets json file.

Added Openshift deployment as an option.


stefwalter:
I've built this container and pushed it. Will deploy it on the various cockpit hosts using the usual pull commands.

stefwalter:
When deploying this container to cockpit-7 - cockpit-11 I'll need to move a files around. I'll do this tomorrow.

Makefile
```diff
@@ -80,10 +80,10 @@ release-install: release-container
 
 verify-shell: docker-running
 	docker run -ti --rm \
-		--privileged \
-		--volume /home/cockpit:/home/user \
+		--privileged --uts=host \
+		--volume /opt/cockpit-verify-secrets:/secrets \
```
martinpitt: IMHO this is a bit of an abuse of [/opt](http://www.pathname.com/fhs/2.2/fhs-3.12.html). `/opt/cockpit-verify/` is somewhat cleaner, with `secrets` and `images` subdirectories, but it's still not a self-sustained third-party app there. This is what `/var` is for. But I won't dwell on this.

Also, files in /opt would normally be root rowned, but the container chowns them to user '1111' which doesn't exist (any more) on the host. This can easily be avoided for the secrets (they can be copied), but not for the images (they are too big and need to be written).

Can the secrets dir be mounted read-only, to ensure that our container doesn't mess around with them? We would never want to ruin a real user's ssh key. The images are less critical as they are public and can always be restored easily (i. e. `/images` is just a cache).

verify/Dockerfile
```diff
@@ -46,28 +46,41 @@ RUN dnf -y update && \
...
 
-RUN chown -R user /build /home/user
 RUN usermod -a -G mock user
 RUN echo 'user ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
 RUN echo "config_opts['basedir'] = '/build/mock/'" >> /etc/mock/site-defaults.cfg
+RUN printf '[user]\n\t\nemail = cockpituous@cockpit-project.org\n\tname = Cockpituous\n' >/home/user/.gitconfig && \
+    ln -s /images /build/images && \
+    mkdir -p /home/user/.config /home/user/.ssh && \
+    ln -snf /secrets/ssh-config /home/user/.ssh/config && \
+    ln -snf /secrets/github-token /home/user/.config/github-token && \
```
martinpitt: Can these be copied at container start time instead of linked at container build time, to ensure that the container doesn't change them? (Like the chown below)

verify/Dockerfile
```diff
@@ -46,28 +46,41 @@ RUN dnf -y update && \
...
+VOLUME /secrets
+VOLUME /images
 
 USER user
 WORKDIR /build
 ENTRYPOINT ["/usr/local/bin/cockpit-verify"]
 CMD ["--publish=fedorapeople.org", "--verbose"]
+
+LABEL INSTALL /usr/bin/docker run -ti --rm --privileged --volume=/:/host:rw --entrypoint=/bin/bash --user=root IMAGE "/usr/sbin/chroot /host /bin/sh -s < /usr/local/bin/install-service"
+LABEL RUN /usr/bin/docker run -ti --rm --privileged --uts=host --volume=/opt/cockpit-verify-secrets:/secrets --volume=/opt/cockpit-verify-images:/images:rw IMAGE /bin/bash -i
```
martinpitt: Same paths/readonly issue here.

verify/README
```diff
@@ -2,77 +2,91 @@
...
+or RHEL 7+.
 
-This is for a single operating system. Choose the operating system to
-test ahead of time.
+Use the following commands to run the verify container as a one off:
 
-## How to deploy
+   $ sudo yum -y install docker atomic
+   $ sudo systemctl start docker
+   $ sudo atomic run cockpit/infra-verify
```
martinpitt: Is it a requirement to start these through `atomic`? What's the main difference to starting them through `docker` directly? Supposedly the difference for installation is that `atomic install` creates a systemd service for you, but for one-off?

verify/build-secrets
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+set -euf
+
+BASE=${1:-./secrets}
+tar -czvf cockpit-verify-secrets.tgz -C "$(dirname $BASE)" "$(basename $BASE)"
```
martinpitt: Where is this tarball being used,  I don't see it anywhere else? Is that ever cleaned up again? (This should be done with `shred -u` as it contains a copy of the user's ssh private key).

verify/README
```diff
@@ -2,77 +2,91 @@
...
+    $ sudo make verify-secrets
+
+Create a service account for use by the verify machines.  Currently the privileged SCC
+is used for scheduling the verify pod. This is because of the requirement to access
+/dev/kvm. Further work is necessary to remove this requirement.
 
-    # git config --global user.email INSERT_EMAIL
-    # git config --global user.name INSERT_NAME
+    $ oc create -f verify/cockpit-verify-account.json
+    $ oc adm policy add-scc-to-user priviliged -z verifier
```
martinpitt: `priviliged` typo.

I'd add `$ shred -u verify/cockpit-verify-account.json` here as it contains a copy of the user's ssh private key and github token.

verify/install-service
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+set -eufx
+
+SECRETS=/opt/cockpit-verify-secrets
+IMAGES=/opt/cockpit-verify-images
+
+echo "options kvm-intel nested=1" > /etc/modprobe.d/kvm-intel.conf
+echo "options kvm-amd nested=1" > /etc/modprobe.d/kvm-amd.conf
+( rmmod kvm-intel && modprobe kvm-intel ) || ( rmmod kvm-amd && modprobe kvm-amd )
```
martinpitt: Can this be guarded with `|| true`, in case `/dev/kvm` or the kernel modules are not available?

verify/Dockerfile
```diff
@@ -46,28 +46,41 @@ RUN dnf -y update && \
...
-VOLUME /build
+VOLUME /secrets
+VOLUME /images
 
 USER user
 WORKDIR /build
 ENTRYPOINT ["/usr/local/bin/cockpit-verify"]
 CMD ["--publish=fedorapeople.org", "--verbose"]
+
+LABEL INSTALL /usr/bin/docker run -ti --rm --privileged --volume=/:/host:rw --entrypoint=/bin/bash --user=root IMAGE "/usr/sbin/chroot /host /bin/sh -s < /usr/local/bin/install-service"
```
martinpitt: I don't understand this -- this runs a command in docker that then escapes back into the host for running `install-service`? Why can't this be run on the host as-is? Could you please add a comment that explains that wrapping?

verify/Dockerfile
stefwalter: These are dangling symlinks. Best practice is to keep secrets out of the container, and certainly not posting secrets to docker hub. These are just references to where the secrets will eventually be mounted, whether by the atomic/docker commands or by kubernetes secrets.

verify/Dockerfile
stefwalter: This is how 'atomic install' generally behaves. I'll explain the wrapping.

verify/README
stefwalter: You can use ```atomic``` stuff to avoid the massive long docker command. You can use ```atomic run``` without ```atomic install```.

verify/build-secrets
stefwalter: This is for copying to other machines.

verify/install-service
stefwalter: I would like the script to fail in that case. Doesn't that make sense?

martinpitt: To avoid trying to run this in an environment without kernel modules, like a chroot or a custom kernel? Not a biggie, and can be changed later if it becomes an issue, so ok.

verify/build-secrets
martinpitt: Hmm. Is that any simpler than scp'ing the `secrets/` directory? If you want/need it, maybe at least chown it to 600?

verify/Dockerfile
martinpitt: Yes, of course - hence my thought of copying them at container startup (not in the Dockerfile), then they can be chowned and chmodded around as needed.

A nice goal would be to support a read-only mount of a user's ~/.ssh directory and copy/chown those into the container's `/run`, to completely avoid any on-disk secret copies. If that's not (easily) possible yet, and the secrets need to be  copied to `/var` or `/opt` on the host, then this approach is fine for now.

Makefile
```diff
@@ -80,10 +80,10 @@ release-install: release-container
 
 verify-shell: docker-running
 	docker run -ti --rm \
-		--privileged \
-		--volume /home/cockpit:/home/user \
+		--privileged --uts=host \
```
stefwalter: Done.

.gitignore
```diff
@@ -2,3 +2,5 @@
 .vimrc
 *.txt
 sink/logs
+/verify/*secrets*
+
```
stefwalter: Done

Makefile
stefwalter: Moving to ```/var/lib/cockpit-tests```. Docker mounts are ro by default, aren't they? I've made this explicit though to assuage any fears.

verify/Dockerfile
stefwalter: > A nice goal would be to support a read-only mount of a user's ~/.ssh directory and copy/chown those into the container's /run, to completely avoid any on-disk secret copies. If that's not (easily) possible yet, and the secrets need to be copied to /var or /opt on the host, then this approach is fine for now.

The things that make this hard in production are:

 * SElinux doesn't allow Docker to access a home directory of a user
 * For Kubernetes/Openshift we need to create a secrets object that contains a copy of these anyway. The secrets will be sent to the Openshift nodes that have access in order to start the appropriate pod, and removed from those nodes afterwards.

verify/Dockerfile
stefwalter: Good catch, docker is rw by default.

verify/build-secrets
stefwalter: Removed this line.

verify/install-service
stefwalter: Moved it to the end of the script, so it will fail the script, but the remainder will have been configured appropriately. Lets see if that works in practice.

Makefile
martinpitt: No, docker manpage says "You can add :ro or :rw suffix to a volume to mount it  read-only or read-write mode, respectively. By default, the volumes are mounted read-write.". If that weren't so, you couldn't chmod the files :-)

.gitignore
```diff
@@ -2,3 +2,4 @@
 .vimrc
 *.txt
 sink/logs
+/verify/*secrets*
```
martinpitt: The commit log still says "both in tarball and in a Kubernetes Secrets json file.", but that can easily be corrected during merging.

stefwalter: Fixed.

verify/Dockerfile
martinpitt: Yeah, I figured something like that. But I'm happy with this now, the documentation explicitly requires the user to put the secrets into `/var/lib/cockpit/`, so they are aware of that and can take appropriate measures.

tests/Dockerfile
```diff
@@ -46,28 +46,44 @@ RUN dnf -y update && \
...
 RUN echo 'user ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
 RUN echo "config_opts['basedir'] = '/build/mock/'" >> /etc/mock/site-defaults.cfg
+RUN printf '[user]\n\t\nemail = cockpituous@cockpit-project.org\n\tname = Cockpituous\n' >/home/user/.gitconfig && \
+    ln -s /images /build/images && \
+    mkdir -p /home/user/.config /home/user/.ssh && \
+    ln -snf /secrets/ssh-config /home/user/.ssh/config && \
+    ln -snf /secrets/github-token /home/user/.config/github-token && \
+    git clone https://github.com/cockpit-project/cockpit /build/cockpit && \
+    cd /build/cockpit && npm install && \
+    chown -R user /build /secrets /images /home/user
```
martinpitt: You can't chown `/secrets` any more as this is a  read-only mount now (that was the point of it). The chown needs to happen on the outside and is already done in `install-service`. I figure it might not actually fail as the files should already have the correct owner, but  I'd take it out for clarity.

tests/Dockerfile
```diff
@@ -46,28 +46,44 @@ RUN dnf -y update && \
...
+VOLUME /secrets
+VOLUME /images
 
 USER user
 WORKDIR /build
-ENTRYPOINT ["/usr/local/bin/cockpit-verify"]
+ENTRYPOINT ["/usr/local/bin/cockpit-tests"]
 CMD ["--publish=fedorapeople.org", "--verbose"]
+
+# We execute the script in the host hence the use of chroot and appropriate piping of the script
```
martinpitt: I'm afraid this comment doesn't really help - the command already makes that clear. I was wondering *why* this runs install-service through that wrapping -- does it make any difference compared to just running it directly on the host?

tests/Dockerfile
```diff
@@ -46,28 +46,44 @@ RUN dnf -y update && \
...
+LABEL INSTALL /usr/bin/docker run -ti --rm --privileged --volume=/:/host:rw --user=root --entrypoint=/bin/bash IMAGE -c \"/usr/sbin/chroot /host /bin/sh -s < /usr/local/bin/install-service\"
```
martinpitt: (sorry, not very familiar with docker): the [documentation](https://docs.docker.com/engine/reference/builder/#label) only describes `LABEL` as taking a key=value pair; is a bare `INSTALL` word valid here?

tests/README.md
```diff
@@ -0,0 +1,104 @@
...
+    $ oc adm policy add-scc-to-user privileged -z tester
+
+Now create all the remaining kubernetes objects:
+
+    $ oc create -f tests/cockpit-tests-secrets.json
+    $ oc create -f tests/cockpit-tests.json
+
+Cleanup the secrets:
+
+    $ shred -u tests/cockpit-tests-secrets.json
```
martinpitt: Requires sudo as well, as `make tests-secrets` was called through sudo.

We could actually avoid both sudos and writing the file in the first place, with

     make tests-secrets | oc create -f -

and adjusting build-secrets to just output to stdout.

tests/install-service
```diff
@@ -0,0 +1,34 @@
...
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+systemctl daemon-reload
+
+echo "options kvm-intel nested=1" > /etc/modprobe.d/kvm-intel.conf
+echo "options kvm-amd nested=1" > /etc/modprobe.d/kvm-amd.conf
+( rmmod kvm-intel && modprobe kvm-intel ) || ( rmmod kvm-amd && modprobe kvm-amd )
```
martinpitt: That will do nicely, good idea to move it to the end!

tests/vm-prep
```diff
@@ -47,6 +47,9 @@ silent()
 
 prepare()
 {
+    # Make sure permissions are correct in containers
+    chown 666 /dev/kvm
```
martinpitt: That won't survive a reboot and /lib/udev/rules.d/80-kvm.rules already sets it up like that, but it can't hurt either.

tests/Dockerfile
stefwalter: The mount is optional. I'm setting this up correctly before hand in case the mount doesn't happen.

tests/Dockerfile
stefwalter: Okay will update the comment.

tests/vm-prep
stefwalter: This happened in practice on openshift nodes.

tests/README.md
stefwalter: Good idea.

tests/Dockerfile
stefwalter: Yup. This is how it's done in other container definitions. You can see these througout the Cockpit codebase as well.

stefwalter:
Fixed and uploaded new containers.

tests/Dockerfile
martinpitt: Ah, ok. Something like `mountpoint /secrets || chown -R user /secrets` would be more robust then. But I'm happy for this to be a followup PR (if/once it's necessary), as the read-only mount now prevents actual damage.

tests/Dockerfile
```diff
@@ -46,28 +46,44 @@ RUN dnf -y update && \
...
+VOLUME /secrets
+VOLUME /images
 
 USER user
 WORKDIR /build
-ENTRYPOINT ["/usr/local/bin/cockpit-verify"]
+ENTRYPOINT ["/usr/local/bin/cockpit-tests"]
 CMD ["--publish=fedorapeople.org", "--verbose"]
+
+# We execute the script in the host, but it doesn't exist on the host. So pipe it in
```
martinpitt: Aah! Yes, that explains it perfectly, thank you!

tests/build-secrets
```diff
@@ -0,0 +1,24 @@
...
+            printf ',\n'
+        fi
+        printf '\t"%s": "%s"' "$(basename $file)" "$(base64 --wrap=0 $file)"
+        first="no"
+    done
+
+    printf '\n} }\n'
+}
+
+secrets > cockpit-tests-secrets.json
```
martinpitt: You now use this in a pipeline, so the file redirection needs to be dropped.

Makefile
```diff
@@ -78,24 +78,24 @@ release-install: release-container
...
+tests-shell: docker-running
 	docker run -ti --rm \
-		--privileged \
-		--volume /home/cockpit:/home/user \
-		--volume $(CURDIR)/verify:/usr/local/bin \
-		--volume=/home/cockpit/verify:/build:rw \
+		--privileged --uts=host \
+		--volume $(CURDIR)/tests:/usr/local/bin \
+		--volume /var/lib/cockpit-tests/secrets:/secrets \
+		--volume=/var/lib/cockpit-tests/images:/images:rw \
```
dperpeet: the previous `--volume` params don't use `=`

PR #104, opened 2017-05-04
test: Make vm-prep double check that cockpit1 doesn't exist

stefwalter:
When run in Kubernetes the Pod can outlast the Container and
so this network may have leaked in from elsewhere. Lets double
check and remove it using brctl.

### DIFF
--- a/tests/vm-prep
+++ b/tests/vm-prep
@@ -64,6 +64,12 @@ prepare()
 		$VIRSH net-undefine $NETWORK_NAME
 	fi
 
+    # In case a network has leaked in from elsewhere
+    if ip address show dev $NETWORK_NAME >/dev/null 2>/dev/null; then
+        ip link set $NETWORK_NAME down || true
+        brctl delbr $NETWORK_NAME || true
+    fi
+
 	# HACK: NetworkManager races with dnsmasq-dhcp on bridge
 	# https://bugzilla.redhat.com/show_bug.cgi?id=1205081
 	cat > /etc/sysconfig/network-scripts/ifcfg-$NETWORK_NAME << EOF


tests/vm-prep
```diff
@@ -64,6 +64,12 @@ prepare()
 		$VIRSH net-undefine $NETWORK_NAME
 	fi
 
+    # In case a network has leaked in from elsewhere
+    if ip address show dev $NETWORK_NAME >/dev/null 2>/dev/null; then
+        ip link set $NETWORK_NAME down || true
+        brctl delbr $NETWORK_NAME || true
+    fi
+
```
martinpitt: The indentation is off.

As the already existing cockpit1 was most likely created by libvirt already, is it right to remove it manually in that way? Does libvirt react to that appropriately, i. e. also remove its own configuration of it? Or will it try to restore it? It looks like this might only remove the actual bridge, but not libvirt's configuration of it?

stefwalter: > The indentation is off.

Various other places in that file are moving to 4 space indentantion. So I figured I would follow suit here.

> We remove it above. It's because a Pod restarts containers inside of it, libvirt is running inside of the containers, but the network namespace is owned by the Pod. And therefore the network outlasts libvirt and the container.

stefwalter:
This is deployed to our openshift cluster.

tests/vm-prep
martinpitt: Right now the indentation looks really broken and confusing (like this new block being outside of the function); can it be consistently changed to 4 spaces instead of a tab then?

Thanks for the explanation about the bridge handling -- and "d'oh!" ☺

stefwalter: > Right now the indentation looks really broken and confusing (like this new block being outside of the function); can it be consistently changed to 4 spaces instead of a tab then?

Done.

martinpitt: No change actually, but I now see that it's already rather garbled, so it doesn't make much of a difference. :-)

martinpitt: Posted PR #106 to clean it up everywhere. Sorry for OCD :-)

PR #105, opened 2017-05-04
tests: Add some notes about what needs config on host machines

stefwalter:
Long term these should go into ansible scripts or some sort of
helpful confugration. But having them documented is a start.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -92,6 +92,17 @@ Some helpful commands:
     $ oc describe pods
     $ oc log -f cockpit-tests-xxxx
 
+The tests need ```/dev/kvm``` to be accessible to non-root users on each node:
+
+    $ sudo chmod 666 /dev/kvm
+
+Some tests need nested virtualization enabled:
+
+    $ sudo -s
+    # echo "options kvm-intel nested=1" > /etc/modprobe.d/kvm-intel.conf
+    # echo "options kvm-amd nested=1" > /etc/modprobe.d/kvm-amd.conf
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -92,6 +92,17 @@ Some helpful commands:
...
+The tests need ```/dev/kvm``` to be accessible to non-root users on each node:
+
+    $ sudo chmod 666 /dev/kvm
+
+Some tests need nested virtualization enabled:
+
+    $ sudo -s
+    # echo "options kvm-intel nested=1" > /etc/modprobe.d/kvm-intel.conf
+    # echo "options kvm-amd nested=1" > /etc/modprobe.d/kvm-amd.conf
+    # ( rmmod kvm-intel && modprobe kvm-intel ) || ( rmmod kvm-amd && modprobe kvm-amd )
```
martinpitt: This is mirroring what the container setup script already does on the host, and what cockpit's test/vm-prep does (except for the chmod). So this is meant for manual setups? Would it be prudent to just refer to vm-prep?

Just a suggestion though, this is (obviously) fine to go in.

stefwalter: Other Openshift clusters need this information. This stuff is for Openshift nodes. 

Longer term @dperpeet is going to come up with a smoother ansible way to make this stuff stick across the Openshift nodes. Cockpit's test/vm-prep is about to be deleted. The container setup script does a bunch more than just this. I'd suggest just waiting until this has a chance to settle before we optimize here.

PR #109, opened 2017-05-05
tests: Use a /cache directory instead of /images

stefwalter:
Also parameterize the place to publish to.

PR #113, opened 2017-05-08
base: Add a push-container script for reuse

stefwalter:
To avoid copying and pasting code in the Makefile for pushing containers, we move the logic there into a shell script.


stefwalter:
> Nice! But can you please rebase to contain only the push-container script commit, not the included "Use /cache directory" commit as well? The latter is already in PR #109.

Done.

PR #114, opened 2017-05-09
release-bodhi: Move to fedora password

martinpitt:
bodhi changed recently to now ask for the Fedora (kinit) passsword, and
the bodhi specific password is not being used any more.  See
https://bugzilla.redhat.com/show_bug.cgi?id=1449045 for details.

Note that this needs the latest python-fedora from updates.

----

The previous 140 release attempt [failed due to this](https://fedorapeople.org/groups/cockpit/logs/release-140/log), and I re-ran the lost steps manually, which finally created a [bodhi update for 140](https://bodhi.fedoraproject.org/updates/cockpit-140-1.fc26). I stopped the release container for the time being, as we need to rebuild it.

### DIFF
--- a/release/README
+++ b/release/README
@@ -36,10 +36,10 @@ update github status:
     /home/cockpit/.ssh/id_rsa.pub
     /home/cockpit/.ssh/known_hosts
     /home/cockpit/.config/bodhi-user
-    /home/cockpit/.config/bodhi-password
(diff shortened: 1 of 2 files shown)


martinpitt:
I built/uploaded a new release container with this and pulled it onto the release machine, so the next release (141) will run with this.

martinpitt:

PR #116, opened 2017-05-10
Add support to sink for verifying GitHub expectations

stefwalter:
This allows the sink to periodically check GitHub to see if state is still as it expects.

### DIFF
--- a/sink/README.md
+++ b/sink/README.md
@@ -119,6 +119,30 @@ current log.
 
 You can use results from the initial status in the final status.
 
+One can also setup an "watches" section in the GitHub status. This
+is a section where the sink will perform the request(s) after 60 seconds
+and check on the results. If the result is not as expected, the sink
+will abort. This is used to prevent collisions or stop in progress
+tasks.
+
+    {
+        "github": {
+            "token": ".......",
+            "watches": [{
+                "method": "GET",
+                "resource": ":repos/statuses/abcdef01234",
+                "result": [
+                    {
+                        "context": "verify/fedora-25",
+                        "state": "pending",
+                        "description": "Testing in progress [host]"
+                    }
+                ]
+            }]
+        }
+    }
+
+
 ### Link field
 
 The `"link"` field will be automatically filled in by the sink. If it
--- a/sink/sink
+++ b/sink/sink
@@ -62,41 +62,6 @@ except:
     MAX_BUFF_SIZE = sys.maxint
 
 
-class GitHubClassic(object):
-    def __init__(self, config):
-        self.token = None
-        try:
-            with open(os.path.expanduser(TOKEN), "r") as gt:
-                self.token = gt.read().strip()
-        except IOError as exc:
-            if exc.errno == errno.ENOENT:
-                 pass
-            else:
-                raise
-
-    def push(self, status):
-        github = status.get("github", { })
-        resource = github.get("resource", None)
-        data = github.get("status", None)
-        if not self.token or not data or not resource:
-            return
-        if "description" not in data and "message" in status:
-            data["description"] = status["message"]
-        if "target_url" not in data:
-            data["target_url"] = status["link"]
-        headers = {"Content-type": "application/json", "User-Agent": "Cockpit Tests" }
-        if self.token:
-            headers["Authorization"] = "token " + self.token
-        conn = HTTPSConnection("api.github.com", strict=True)
-        # conn.set_debuglevel(1)
-        conn.request("POST", resource, json.dumps(data), headers)
-        response = conn.getresponse()
-        output = response.read()
-        conn.close()
-        if response.status < 200 or response.status >= 300:
-            raise RuntimeError("Couldn't update GitHub: {0} {1}\n{2}\n".format(
-                               response.status, response.reason, output.strip()))
-
 # Expansion of strings in JSON objects.
 
 def jpath(val, path):
@@ -285,7 +250,7 @@ class Extras(object):
 
 class Status(object):
     def __init__(self, config, identifier):
-        self.reporters = [ GitHubClassic(config), GitHub(config), Irc(config), Extras(config) ]
+        self.reporters = [ GitHub(config), Irc(config), Extras(config) ]
         self.link = config.get("Sink", "Url").replace('@@', identifier)
         self.data = None
 
--- a/sink/sink
+++ b/sink/sink
@@ -335,6 +335,21 @@ class Buffer(object):
         result += os.read(self.fd, 1024)
         return result
 
+class Tee():
+    def __init__(self, log, std):
+        self.log = log
+        self.std = std
+    def write(self, *args):
+        self.log.write(*args)
+        self.std.write(*args)
+    def flush(self, *args):
+        self.log.flush(*args)
+        self.std.flush(*args)
+    def close(self, *args):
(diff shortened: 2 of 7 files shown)


stefwalter:
CC @mvollmer FYI

stefwalter:
This has been deployed to fedorapeople.org

stefwalter:
Tested, reworked, and deployed to fedorapeople.org

sink/sink
```diff
@@ -89,9 +97,31 @@ def expand(val, env):
...
+# Compare two JSON style values to see if one is contained in the other
+# For lists and dict make sure all items in expect are also present in result
+# For any other values a strict comparison is used
+def contains(expect, result):
+    if isinstance(expect, list) and isinstance(result, list):
+        for i in range(0, len(expect)):
+            res = result[i] if i < len(result) else None
+            if not contains(expect[i], res):
+                return False
+        return True
```
martinpitt: This will say False if the result list does not start with the expect list, but the documentation suggests that expect should merely be *anywhere* in result. I. e. `contains([2], [1, 2, 3])` will be `False` with this. So either the description or the algorithm needs to be changed.

Is there really a use case for comparing GitHub states where the "expected list is a prefix of result" behaviour makes sense? It seems to me that comparing for equality or real "contained in" makes more sense and is more robust against future changes in GitHub's API?

sink/sink
```diff
@@ -118,29 +158,62 @@ class GitHub(object):
...
+        except OSError:
+            pass
+
+    # Check if a given field in GitHub is not as expected
+    def watches(self, token, watches):
+        for watch in watches:
+            try:
+                interval = int(watch.get('interval', None))
+            except TypeError:
+                interval = 60
```
martinpitt: You can just give the `60` as default value  instead of  `None`.

sink/test-logic
```diff
@@ -0,0 +1,77 @@
...
+    { "ret": False, "expect": "x", "result": [ ] },
+    { "ret": False, "expect": { }, "result": [ ] },
+    { "ret": False, "expect": { }, "result": "x" },
+    { "ret": False, "expect": [ ], "result": "x" },
+
+    # Lists
+    { "ret": True,  "expect": [ "a" ], "result": [ "a", "b", "x", "z" ] },
+    { "ret": True,  "expect": [ "a", "b" ], "result": [ "a", "b", "x", "z" ] },
+    { "ret": False, "expect": [ "y" ], "result": [ "a", "b", "x", "z" ] },
+    { "ret": False, "expect": [ "z" ], "result": [ "a", "b", "x", "z" ] },
```
martinpitt: This reflects what I mentioned above -- `z` is clearly contained in the list, so "contains" is a poor name.

sink/sink
```diff
@@ -100,20 +100,30 @@ def expand(val, env):
 # Compare two JSON style values to see if one is contained in the other
 # For lists and dict make sure all items in expect are also present in result
 # For any other values a strict comparison is used
-def contains(expect, result):
+def contains(expect, result, debug=True):
```
martinpitt: The commit message of the 4th commit is identical to the third one. This should be squashed into the 3rd commit.

sink/sink
martinpitt: Ah, so this fourth commit already (attempts to) change that to a real "contains". It still seems a bit weird to me that multiple occurrences count as "false", though? What's the use case for that?

With this algorithm you lose the ordering of `expect` -- i. e. `contains([2, 1], [1, 2, 3])` should now be `True`.
(There are a [couple of recipes on StackExchange]http://stackoverflow.com/questions/3847386/testing-if-a-list-contains-another-list-with-python) for this, BTW). However, I think in this GitHub case this "is a subset" behaviour is actually desirable. You can write that much more succinctly as `if set(expect).issubset(result)`.

martinpitt:
Please check the "show outdated" in my above comment on sink/sink, not sure why GitHub marked it as outdated right away.

sink/sink
stefwalter: What if ```'interval'``` is not a number?

sink/sink
stefwalter: Done

stefwalter:
> Ah, so this fourth commit already (attempts to) change that to a real "contains". It still seems a bit weird to me that multiple occurrences count as "false", though? What's the use case for that?

For when multiple bots try to do something, such as file an issue.

stefwalter:
> However, I think in this GitHub case this "is a subset" behaviour is actually desirable. You can write that much more succinctly as if set(expect).issubset(result).

Unfortunately that code doesn't nest the contains logic for set/issubset as is necessary when interacting with a JSON API.

stefwalter:
> Please check the "show outdated" in my above comment on sink/sink, not sure why GitHub marked it as outdated right away.

Probably because I pushed another commit while you were reviewing? The one that changed it back to proper contains logic?

sink/sink
martinpitt: Then you will get a `ValueError`. IMHO failing explicitly with that is better than silently ignoring the bad configuration value and silently assuming 60.

sink/sink
```diff
@@ -89,9 +97,38 @@ def expand(val, env):
...
 
+# Compare two JSON style values to see if one is contained in the other
+# For lists and dict make sure all items in expect are also present in result
+# For any other values a strict comparison is used
+def contains(expect, result, debug=True):
+    if isinstance(expect, list) and isinstance(result, list):
+        result = list(result)
+        for exp in expect:
+            while len(result):
+                res = result.pop(0)
```
martinpitt: This will eat away all elements which don't match the first one in `expect` from the `result` copy. But they could still be matches for the second, third, etc. element in `expect`. Also, you now don't assert that each element in expect occurs only once in `list`, I thought this was important in case the same action was being done more than once in `result`?

For simple elements this would be correct even for repetitions of elements in `expect`:
```python
for i in expect:
    if  expect.count(i) != result.count(i):
        return False
else:
    return True
```
If `i` is an aggregate (list or dict) it still needs recursion of course, and then the case of "multiple instances of the same value" is not very well defined, as it's not clear whether an element which is a subset of the expected one should count as "same" or "different".

Can we somehow avoid the case that multiple identical entries in a list are significant? Perhaps by always appending the container name as a suffix to action labels, or something such?

stefwalter: I've moved back to the old code since this was a step backwards. The intent is to ensure that exactly one of the expected item is contained in the outer container. 

Please see accompanying pull requests in cockpit for cases where this is used.

> Can we somehow avoid the case that multiple identical entries in a list are significant? Perhaps by always appending the container name as a suffix to action labels, or something such?

It's hard. Because the entire point of this is to detect collisions and duplicates.

sink/sink
stefwalter: Fixed.

sink/README.md
```diff
@@ -119,6 +119,30 @@ current log.
 
 You can use results from the initial status in the final status.
 
+Oen can also setup an "watches" section in the GitHub status. This
```
martinpitt: "Oen" typo

stefwalter: Fixed.

stefwalter:
I also changed the logic so that unless an "interval" is specified, we don't repeat the request.

PR #118, opened 2017-05-12
release-source: Don't generate patches for non-tarball commits

stefwalter:
In release-source we generate patches for every git commit
after the tag. However some commits don't affect the tarball
and therefore should be skipped and not generate patches
for the tarball.


martinpitt:

PR #120, opened 2017-05-12
tests: Run or load the .tasks file in the target repository

stefwalter:
This file produces the possible set of bot tasks to be executed in the
current repository by bots. Empty lines and commented lines are ignored
and commented lines are removed. If this file is executable it will be
run, and the output will be used as the list of tasks.

The tasks are sorted descending alphabetically and a task is chosen from
near the top of the sorted result. The task will be run in a shell by the
Cockpit bots.

Depends on: 

 * [x] https://github.com/cockpit-project/cockpit/pull/6509
 * [x] https://github.com/cockpit-project/cockpit/pull/6552

### DIFF
--- a/tests/cockpit-tests
+++ b/tests/cockpit-tests
@@ -31,16 +31,11 @@ if [ ! -d cockpit ]; then
     git clone https://github.com/cockpit-project/cockpit
 fi
 
+# TODO: This needs to move into tests bot
 if [ ! -f $TMPDIR/vm-prep-was-run ]; then
     sudo /usr/local/bin/vm-prep
     touch $TMPDIR/vm-prep-was-run
 fi
-
-if [ ! -d cockpit/node_modules ]; then
-    ( cd cockpit && npm install )
-fi
-
-mkdir -p mock
 
 # If interactive arguments then just exec
 if [ $loop != "yes" ]; then
@@ -53,16 +48,41 @@ echo "Starting testing"
 iteration=0
 while true; do
     fail=0
-    find cockpit -name '*.py?' -delete
     git -C cockpit fetch origin
     git -C cockpit reset --hard origin/master
-    if [ "$iteration" -eq 0 ]; then
-        cockpit/test/vm-download --prune
-    else
-        cockpit/test/vm-download
-    fi
+
+    # TODO: Soon we'll move this to an images bot
     iteration=$(( (iteration+1) % 25 ))
-    if ! cockpit/test/github-task "$@"; then
-        sleep 60
+    if [ -x cockpit/test/vm-download ]; then
+        if [ "$iteration" -eq 0 ]; then
+            cockpit/test/vm-download --prune
+        fi
     fi
+
+    (
+        # If the tasks file is executable, run it
+        if [ -x cockpit/.tasks ]; then
+            cockpit/.tasks
+
+        # If it exists then cat it
+        elif [ -e cockpit/.tasks ]; then
+            cat cockpit/.tasks
+
+        # No tasks found
+        else
+            echo "echo 'No tasks found in cockpit repo' >&2; exit 1"
+        fi
+
+    # Remove empty lines, comments, and reverse sort.
+    # Then randomly pick one of the first 10 commands
+    ) | sed -e 's/#.*$//' -e '/^$/d' | sort -r | head -n 10 | shuf | while read line; do
+
+        # Execute the task in question
+        if ! /bin/sh -c "cd cockpit/; set -ex; $line"; then
+            sleep 60
+        fi
+
+        # Start the loop again up above
+        break
+    done
 done


tests/cockpit-tests
```diff
@@ -31,17 +31,12 @@ if [ ! -d cockpit ]; then
     git clone https://github.com/cockpit-project/cockpit
 fi
 
+# TODO: This needs to move into tests bot
 if [ ! -f $TMPDIR/vm-prep-was-run ]; then
     sudo /usr/local/bin/vm-prep
     touch $TMPDIR/vm-prep-was-run
 fi
 
-if [ ! -d cockpit/node_modules ]; then
```
petervo: Not saying this is wrong, just wondering what changed that makes this no longer required.

stefwalter: A long time ago, we wouldn't load node_modules during autogen.sh. We started looping at the end of this script. That means the first autogen.sh in the container installs the node_modules anyway, without efficiency problems.

stefwalter:
Deployed this to all machines, slowly the containers will start to use it.

stefwalter:
Almost all containers have now moved over to this.

tests/Dockerfile
```diff
@@ -53,6 +53,7 @@ RUN echo 'user ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
 RUN echo "config_opts['basedir'] = '/build/mock/'" >> /etc/mock/site-defaults.cfg
 RUN printf '[user]\n\t\nemail = cockpituous@cockpit-project.org\n\tname = Cockpituous\n' >/home/user/.gitconfig && \
     ln -s /cache/images /build/images && \
+    ln -s /cache/images /build/virt-builder && \
```
martinpitt: This will throw virt-builder and cockpit images into the same directory. They won't collide on names, but does it really make sense to replicate and distribute them across all of our image stores? Keeping that image around locally seems sufficient.

Are we sure that the virt-builder images never change? I. e. Debian 8 or Ubuntu 16.04 regularly get point releases, and the virt-builder images might get refreshed under the same name. TBH I'd treat them as an implementation detail of virt-builder and not expose them to our mirroring infra.

tests/Dockerfile
```diff
@@ -69,6 +69,7 @@ RUN touch /etc/modprobe.d/kvm-amd.conf && touch /etc/modprobe.d/kvm-intel.conf
 # Continue to work with a host defined cockpit1 network if necessary
 RUN chmod -v u+s /usr/libexec/qemu-bridge-helper && echo 'allow cockpit1' >> /etc/qemu/bridge.conf
 
+ENV LIBGUESTFS_BACKEND direct
```
martinpitt: Niiiiiice! I wasn't aware that this is possible. \o/

tests/cockpit-tests
```diff
@@ -53,16 +48,41 @@ echo "Starting testing"
...
+        if [ -x cockpit/.tasks ]; then
+            cockpit/.tasks
+
+        # If it exists then cat it
+        elif [ -e cockpit/.tasks ]; then
+            cat cockpit/.tasks
+
+        # No tasks found
+        else
+            echo "echo 'No tasks found in cockpit repo' >&2; exit 1"
```
martinpitt: We don't set pipefail, so this should work. But I wonder why this doesn't just exit cleanly with 0 in this case?

stefwalter: Because we want the code below to wait for 60 seconds.

tests/Dockerfile
stefwalter: > This will throw virt-builder and cockpit images into the same directory. They won't collide on names, but does it really make sense to replicate and distribute them across all of our image stores? Keeping that image around locally seems sufficient.

They are not replicated, unless someone asks for them. They are placed in a cache directory so other containers sharing that cache directory can use the images.

> Are we sure that the virt-builder images never change? I. e. Debian 8 or Ubuntu 16.04 regularly get point releases, and the virt-builder images might get refreshed under the same name. 

The file names are incremented with a number here: http://libguestfs.org/download/builder/index.asc  virt-builder itself caches these images and would fail if it didn't change the names. 

> TBH I'd treat them as an implementation detail of virt-builder and not expose them to our mirroring infra.

They are part of our cache, not part of any mirroring.

martinpitt: > The file names are incremented with a number here:

Ah, ok. So as there's no debian-8.1 (for example) I suppose that these images never actually change (which is a pity, as it will need an ever-increasing dist upgrade for each usage, but that's not under our control).

tests/cockpit-tests
martinpitt: OK, now I'm confused; if you want the `exit 1` to have any actual meaning, you need to set pipefail in the outer shell. `false | while read f; do echo $f; done` succeeds, doesn't actually iterate over anything, and does exactly the same as `true | while ...`. So in this caes there would be no sleep (and we don't want any either), but it would be the same (but cleaner) result with `exit 0`.

So, in summary, I think this is just a stylistic issue (and a bit confusing), I just want to make sure that you don't expect this `exit 1` to actually have a visible effect.

tests/Dockerfile
stefwalter: Yup @martinpitt. Noticed that too. I don't think those images have ever been updated.

martinpitt:
Needs a rebase, as this includes the commit from #119. As usual, please either force-push to this PR (and I press the button) or land it yourself.

PR #121, opened 2017-05-16
images: Run cockpit/images container on HTTP as well

stefwalter:
To make life simpler for folks accessing the images, expose the images over HTTP and HTTPS as well.
    
If we need security for downloads we should verify the hashes during the download.


stefwalter:
CC @mvollmer 

This is deployed to verifymachine4 verifymachine5 and all e2e test machines. In addition the kubernetes service changes have been deployed.

martinpitt:
LGTM. Apparently I still can't push to origin myself (`error: GH006: Protected branch update failed for refs/heads/master.`). Can you please add the "Closes #121" bits, and force-push or just land yourself?

PR #124, opened 2017-05-18
tests: Wait for 60 seconds when no task is available

stefwalter:
Wait for 60 seconds when no task is available or the last task
failed.

### DIFF
--- a/tests/cockpit-tests
+++ b/tests/cockpit-tests
@@ -45,20 +45,7 @@ fi
 
 echo "Starting testing"
 
-iteration=0
-while true; do
-    fail=0
-    git -C cockpit fetch origin
-    git -C cockpit reset --hard origin/master
-
-    # TODO: Soon we'll move this to an images bot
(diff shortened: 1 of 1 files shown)


martinpitt:
An idle bot certainly shouldn't go into that busy loop where it runs sed | head all the time, but waiting an entire minute sounds like they would become unnecessarily unresponsive. That's not a hard objection though, I'm okay with landing this especially if it's already deployed.

stefwalter:
> An idle bot certainly shouldn't go into that busy loop where it runs sed | head all the time, but waiting an entire minute sounds like they would become unnecessarily unresponsive. That's not a hard objection though, I'm okay with landing this especially if it's already deployed.

The responsiveness goes up with the number of bots. If there really was nothing to do, the chances that in less than 60 seconds there *is* something to do is really low. In addition with multiple bots checking this, I think things remain reasonably responsive.

PR #125, opened 2017-05-19
release: Enforce success on Fedora 26, drop Fedora 25

martinpitt:
We don't regularly release to Fedora 25 any more, so drop koji build.
    
Stop allowing failures on Fedora 26 as that is now our primary target.
See https://github.com/cockpit-project/cockpit/pull/6602

### DIFF
--- a/release/major-cockpit-release
+++ b/release/major-cockpit-release
@@ -17,8 +17,7 @@ job release-srpm
 
 # Do fedora builds for the tag, using tarball
 job release-koji -k master
-job release-koji f25
-job release-koji -k f26
+job release-koji f26
 
 # Upload release to github, using tarball
 job release-github


release/major-cockpit-release
```diff
@@ -17,8 +17,8 @@ job release-srpm
 
 # Do fedora builds for the tag, using tarball
 job release-koji -k master
-job release-koji f25
-job release-koji -k f26
+job release-koji -k f25
```
stefwalter: I don't think we should commit to dist-git here automatically at all. Because the next time (if) we do, we would want to then backport a patch, and a bunch of other useless dist-git commits would be in the way. I think we should just remove this line.

martinpitt: Good point, updated.

PR #127, opened 2017-05-20
tests: Only do 60 tasks before exiting the container

stefwalter:
This avoids all sorts of disk usage, and temp file problems and also allows automatically upgrading to the latest tagged containers.
    
We can tweak the exact numbers here if they need to be different.

 - [x] land/rebase on #124


stefwalter:
Rebased.

PR #131, opened 2017-05-24
tests: Add cleanup hacks for container networking

stefwalter:
The first hack is because we don't want to see this error message
when libvirt tries to write bridge related stuff:

    error: Unable to set bridge cockpit1 forward_delay: Read-only file system

Secondly, when run in Kubernetes the Pod can outlast the Container and
so this network may have leaked in from elsewhere. Lets double
check and remove it using brctl.

Neither of these hacks work in the tasks code because it may be
executing older branches which do not account for these things.

### DIFF
--- a/tests/cockpit-tests
+++ b/tests/cockpit-tests
@@ -26,6 +26,16 @@ if [ -n "$TMPDIR" ]; then
     ln -snf $TMPDIR/libvirt /tmp/libvirt
 fi
 
+# HACK: /sys is read-only in containers. And libvirt tries
+# to write there, but only if a specific subtree exists here
+sudo -n mount -o bind /run /sys/devices/virtual/net/ || true
+
+# In case a network has leaked in from elsewhere
+if sudo -n ip address show dev cockpit1 >/dev/null 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


tests/cockpit-tests
```diff
@@ -26,6 +26,15 @@ if [ -n "$TMPDIR" ]; then
     ln -snf $TMPDIR/libvirt /tmp/libvirt
 fi
 
+# HACK: In case of read-only /sys avoid libvirt writing there
+mount -o bind /mnt /sys/devices/virtual/net/
```
martinpitt: This seems a bit dangerous when running on someone's production machine. Does libvirt insist on a writable directory? If so, the comment  is confusing, as this would *allow* libvirt to write there instead of  avoiding it.

stefwalter:
Deployed to all test machines.

tests/cockpit-tests
martinpitt: I. e. maybe `mkdir -p /run/libvirt` and bind-mount that?

stefwalter: This is inside a container, but i'll make it more clear.

PR #132, opened 2017-05-25
tests: Add a volume to container for tmp

stefwalter:
Add a scratch empty volume to a container for the /tmp directory
and start to do the builds and running of VMs there.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -45,7 +45,7 @@ RUN dnf -y update && \
 ADD https://raw.githubusercontent.com/cockpit-project/cockpit/master/tools/cockpit.spec /tmp/cockpit.spec
 RUN dnf -y builddep /tmp/cockpit.spec && dnf clean all
 
-RUN mkdir -p /usr/local/bin /home/user /build /build/tmp /secrets /cache/images /cache/github
+RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/github
 ADD cockpit-tests install-service /usr/local/bin/
 
 RUN usermod -a -G mock user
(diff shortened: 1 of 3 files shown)


tests/Dockerfile
```diff
@@ -55,6 +55,7 @@ RUN printf '[user]\n\t\nemail = cockpituous@cockpit-project.org\n\tname = Cockpi
     ln -s /cache/images /build/images && \
     ln -s /cache/images /build/virt-builder && \
     ln -s /cache/github /build/github && \
+    ln -s /tmp /build/tmp && \
```
martinpitt: OOI, is /build/tmp/ still being used anywhere, or is this just a safety net in case we forgot a spot?

tests/cockpit-tests
```diff
@@ -37,7 +30,7 @@ if sudo -n ip address show dev cockpit1 >/dev/null 2>/dev/null; then
 fi
 
 if [ ! -d cockpit ]; then
-    sudo chown -R user /build
+    sudo chown -R user .
```
martinpitt: The script doesn't change directory by itself, so it's not clear that '.' is always the desired location to change. Could this be more explicit?

tests/Dockerfile
stefwalter: ```/build``` is set to ```$TEST_DATA``` ... and we start to use ```$TEST_DATA/tmp``` in the cockpit repo. There's one open #6733 and one closed #6701 pull request.

tests/cockpit-tests
stefwalter: But we checkout cockpit to the current directory. Would a ```./``` be more explicit?

martinpitt: Not really, `./` is exactly the same as `.`. It's rather blunt, but oh well,  this is all inside a container, so the risk of breaking something permanently is low.

PR #134, opened 2017-06-03
tests: Add Zanata to tests container for translations

stefwalter:


stefwalter:
Built container and pushed to all bots.

PR #135, opened 2017-06-06
release: Check expiration of copr config

martinpitt:
Abort release-copr in the "prepare" phase already if ~/.config/copr is
expired. Previously it failed too late in the "commit" phase on an
expired token, breaking the release.

### DIFF
--- a/release/release-copr
+++ b/release/release-copr
@@ -55,8 +55,28 @@ readpath_or()
     fi
 }
 
+check_config() {
+    local expiration
+
+    trace "Validating ~/.config/copr"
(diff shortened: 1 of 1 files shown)


stefwalter:
Ready for a container to be built. Lets merge after this successfully is used in a release.

martinpitt:
Release 142 was successfully done with that (other than the Fedora Rawhide perl-Git breakage): https://copr.fedorainfracloud.org/coprs/g/cockpit/cockpit-preview/packages/

So I think this is ready to land.

PR #138, opened 2017-07-17
Add american-fuzzy-lop to things we want to use while testing

stefwalter:
This is a prerequisite to the bots doing some fuzzing.


stefwalter:
Pushing now.

martinpitt:
Arrived at https://hub.docker.com/r/cockpit/tests/tags/ now

PR #139, opened 2017-07-18
tests: Document doing modprobe /dev/kvm on boot

stefwalter:
Unless libvirt is used on the Openshift host, the kvm module
is often not loaded by itself.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -104,6 +104,8 @@ Some helpful commands:
 
 The tests need ```/dev/kvm``` to be accessible to non-root users on each node:
 
+    $ sudo modprobe kvm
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -104,6 +104,8 @@ Some helpful commands:
 
 The tests need ```/dev/kvm``` to be accessible to non-root users on each node:
 
+    $ sudo modprobe kvm
+    $ printf 'kvm\n' | sudo tee /etc/modules-load.d/kvm.conf
```
martinpitt: A simple "echo kvm" would do, but as this usually gets copy&paste'd, it makes little difference.

PR #140, opened 2017-07-20
tests: Make the qemu logs persistent

stefwalter:
I'd like to consume these logs in talks and discussions about our
setup. These show the number of VMs that were launched, which operating
systems, on which hosts, and for how long they ran.

I would like to gather these for many months without cleaning up the
data.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -48,7 +48,7 @@ RUN dnf -y update && \
 ADD https://raw.githubusercontent.com/cockpit-project/cockpit/master/tools/cockpit.spec /tmp/cockpit.spec
 RUN dnf -y builddep /tmp/cockpit.spec && dnf clean all
 
-RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/github
+RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/github /build/libvirt/qemu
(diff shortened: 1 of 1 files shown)


stefwalter:
Image build uploaded and tested on verifymachine5.

tests/Dockerfile
```diff
@@ -57,6 +57,7 @@ RUN echo "config_opts['basedir'] = '/build/mock/'" >> /etc/mock/site-defaults.cf
 RUN printf '[user]\n\t\nemail = cockpituous@cockpit-project.org\n\tname = Cockpituous\n' >/home/user/.gitconfig && \
     ln -s /cache/images /build/images && \
     ln -s /cache/images /build/virt-builder && \
+    ln -s /cache/images /build/libvirt/qemu/log && \
```
martinpitt: Are you sure that you don't want to put them into the /cache/images/log/ subdirectory? Seems nicer to keep the images and qemu logs apart. Not a blocker of course, just a question.

stefwalter: We could ... that's more complicated to work out. I'd like to consume them via http, but having them in a subdirectory isn't *that* bad. It does make the whole symlinking stuff more complicated though, having to create directories with correct permissions during startup etc.

stefwalter: libvirt uses the $XDG_CACHE_HOME for qemu logs. Mixing of cache and logs is not a surprise.

PR #142, opened 2017-07-31
sink: The sink can make arbitrary HTTP requests

stefwalter:
This code change allows the sink to make arbitrary HTTP requests.
It always code do this, but was all framed within the GitHub
moniker. This is mostly cleanup.

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -150,7 +150,7 @@ class GitHub(object):
         method = request.get('method', "GET")
         resource = expand(request.get('resource', None), self.results)
         data = expand(request.get('data', None),  self.results)
-        token = token or self.token
+        token = request.get('token', token or self.token)
         headers = {"Content-type": "application/json", "User-Agent": "Cockpit Tests" }
         if token:
             headers["Authorization"] = "token " + token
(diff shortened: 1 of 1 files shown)


sink/sink
```diff
@@ -195,22 +195,23 @@ class GitHub(object):
 
     def push(self, status):
         github = status.get("github", { })
+        http = status.get("github", { })
```
martinpitt: I suppose you mean something else here, as otherwise this would just duplicate the entries in `requests`.

sink/sink
```diff
@@ -195,22 +195,23 @@ class GitHub(object):
...
         token = github.get("token", self.token)
-        requests = github.get("requests", [])
+        requests = github.get("requests", []) + http.get("requests", [])
 
         for r in requests:
             self.results['link'] = status['link']
             result_name = r.get('result', None)
 
-            result = self.req(token, r)
+            result = self.req(r.get('token'), r)
```
martinpitt: Is there an `or token` missing, or will the request now always have the token due to the changes in `req()` above? The `token =` assingment from above is still being used further down in `self.watches(token, watches)`, does that not need to use the request specific token(s) as well?

sink/sink
stefwalter: Thanks. Fixed.

sink/sink
stefwalter: Thanks. Good catch. Fixed.

stefwalter:
I rolled it out on fedorapeople.org, and it seems to be working since a few machines are still posting there.

PR #143, opened 2017-07-31
tests: Add basic users to container

stefwalter:
This allows people to log in if they get that far using the
cockpit/tests container in this way.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -51,8 +51,8 @@ RUN dnf -y builddep /tmp/cockpit.spec && dnf clean all
 RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/github
 ADD cockpit-tests install-service /usr/local/bin/
 
-RUN usermod -a -G mock user
-RUN echo 'user ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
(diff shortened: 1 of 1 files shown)


tests/Dockerfile
```diff
@@ -51,8 +51,8 @@ RUN dnf -y builddep /tmp/cockpit.spec && dnf clean all
 RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/github
 ADD cockpit-tests install-service /usr/local/bin/
 
-RUN usermod -a -G mock user
-RUN echo 'user ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
+RUN usermod -a -G mock user && useradd -a -G wheel admin
+RUN printf 'user ALL=(ALL) NOPASSWD: ALL\nadmin ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
```
martinpitt: I suppose it's not relevant, but this now lacks a terminating newline in the `admin` line. It was there before as echo automatically terminates. But fine for me if you tested this.

martinpitt:
This has bitrotted for a while, and I never actually noticed that this was necessary for anything. We are generally moving towards fixing our containers to work in unprivileged OpenShift environments with random UIDs, so I vote for actually not adding more hardcoded users.

PR #144, opened 2017-07-31
WIP: images: Generalize to sharing of state

stefwalter:
The bots can share state. Previously this state was just created
images, but we generalize this to represent all state that
should be shared. This state is still in named files, just like
images.


martinpitt:
Closing this now, as it is very stale. I think this more generic renaming is a nice cleanup if someone feels inclined to re-do it, but it's no functional change (and thus not that important).

PR #145, opened 2017-08-08
release: Wait for copr build to finish

petervo:
That way we can be sure the rpms will be available to
the docker images.

### DIFF
--- a/release/release-copr
+++ b/release/release-copr
@@ -100,7 +100,7 @@ prepare()
 
 commit()
 (
-    if ! copr-cli build --nowait $COPR "$URL"; then
+    if ! copr-cli build $COPR "$URL"; then
         if [ $CONTINUE -eq 0 ]; then
             exit 1
         fi


martinpitt:
I built and deployed a new release container with this. After a successful cockpit release (planned for tomorrow morning) I'll merge this and send the container dockerhub-wards.

martinpitt:
I don't think that actually worked - the release script just finished (), but on the [copr page](https://copr.fedorainfracloud.org/coprs/g/cockpit/cockpit-preview/builds/) the 150 build is currently "running". As expected there is no new build on https://hub.docker.com/r/cockpit/ws/tags/ - I guess we should kick that manually after the copr build is done?

martinpitt:
Today's copr build for 150 failed because F24 is not supported any more. This just got fixed in the docker file, so the next round actually should work. Let's re-check this for 151 next week, in the meantime this will stay deployed on releasemachine.

martinpitt:
With today's 151 release this now worked perfectly. I observed `copr-cli` block in the release container until the [COPR build](https://copr.fedorainfracloud.org/coprs/g/cockpit/cockpit-preview/build/605826/) finished, and only afterwards the [container version bump commit](https://github.com/cockpit-project/cockpit-container/commit/2dff107fbfae04) was done. Then the [new container got auto-built](https://hub.docker.com/r/cockpit/ws/tags/).

martinpitt:
Release container pushed to dockerhub.

PR #146, opened 2017-08-21
bodhi: update for F26 bodhi

nmav:
That tool changed all its options, and provides no backwards compatibility.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@redhat.com>

### DIFF
--- a/release/release-bodhi
+++ b/release/release-bodhi
@@ -86,9 +86,9 @@ commit()
         fednvr="${NVR/%fc*/fc$number}"
     fi
 
-    echo "$PASSWORD" | setsid -w -- bodhi --new --type=enhancement --close-bugs --notes="$NOTES" \
-        --stablekarma=2 --unstablekarma=-1 --request=stable --release="$FEDORA" \
-        --username=$USER ${fednvr}
+    echo "$PASSWORD" | setsid -w -- bodhi updates new --type=enhancement --close-bugs --notes="$NOTES" \
+        --stable-karma=2 --unstable-karma=-1 --request=stable --user $USER  \
+        ${fednvr}
 )
 
 while getopts "p:qvxz" opt; do


nmav:
Usage verified with gnutls 3.5.15 release in F25.

martinpitt:
Our release container is Fedora 25 based. To clarify: I suppose this will be necessary once we upgrade our release container to Fedora 26. Or does it (somehow) affect the cockpit release/upload *to* Fedora 26?

stefwalter:
@bowlofeggs Do you know why the bodhi client was updated to be incompatible with previous versions in Fedora 26? Any mitigation we can use to work across old and new versions?

bowlofeggs:
@stefwalter Bodhi 2 is not compatible with Bodhi 0.9. Bodhi 2 was added to Rawhide before Fedora 26 branched, so it is the first release to have it. Since the Fedora [update policy](https://stg.fedoraproject.org/wiki/Updates_Policy#Stable_Releases) states not to make backwards incompatible updates in any given Fedora release, Fedora 25 is stuck on the old Bodhi 0.9. There is a [Copr](http://copr.fedorainfracloud.org/coprs/bowlofeggs/bodhi/) that provides the Bodhi 2 client to Fedora 25 users.

bowlofeggs:
Oh, and for history - I don't know why the CLI was rewritten, but it was completely redone from the ground up before I started working on Bodhi.

As for mitigation, there is a ```--version``` flag on both versions of the client that will print its version. You could use that to detect whether you are dealing with 0.9 or 2.x and use its options accordingly.

bowlofeggs:
Keep in mind that F25 only has a few months of life left. It is currently scheduled to go EOL on November 31.

martinpitt:
Setting blocked, as this depends on us updating the containers to F26.

martinpitt:
I ran the 156 release with this, and https://bodhi.fedoraproject.org/updates/cockpit-156-1.fc27 got created successfully.

PR #147, opened 2017-09-15
tests: Add 12 hour timeout for any task

stefwalter:
Various tasks have timeouts of their own such as running tests
or building images. But if any task hangs for more than 12 hours
we kill it.

We do want to allow for tasks that take a long time such as
machine learning tasks, but they should be broken up to take
significantly less than half a day.


stefwalter:
Pushed a new container based on this

stefwalter:
Running this container locally as a test

PR #149, opened 2017-10-02
Fix typos in release-srpm

ueno:


ueno:
Sorry, I was too dumb and didn't realize that updating ```%changelog``` is done in two steps: first by release-srpm and later by release-koji.  So here release-srpm works as expected.  I am withdrawing the changes.

PR #152, opened 2017-10-07
tests: Add sklearn dependency

stefwalter:
We want to be able to use machine learnin from the python sklearn
library in our bots.


stefwalter:
New image has been pushed, and all cockpit-tasks pods are cycling.

martinpitt:
Landing this as you already pushed it into production, and looks harmless.

PR #154, opened 2017-11-09
images: Expose http on standard port 80

martinpitt:
Cockpit expects to talk to cockpit-11 on the standard HTTP port [1], and
apparently the created cockpit-tests.service had been manually changed
for that until the reinstall. As we don't need port 80 for anything
else, change the port in the Dockerfile instead of changing REDHAT_PING,
to keep the recent behaviour.

[1] https://github.com/cockpit-project/cockpit/blob/master/bots/tests-scan#L58


martinpitt:
This caused all RHEL related test requests to get ignored by the bot. I deployed the change on cockpit-11, now the tests are running.

martinpitt:
If this is not desirable  for some reason, I'm also happy to send a PR against cockpit to change `REDHAT_PING` to port 8090. This works equally well.

PR #155, opened 2017-11-12
tests: Install chromium-headless

martinpitt:
This allows us to experiment with moving Cockpit tests from PhantomJS to
the Chrome Debugging Protocol, using headless Chromium
(https://github.com/cockpit-project/cockpit/pull/8069)

----

Note that there has not yet been a final decision between Electron and Chromium, but unlike electron it is impractical to pull in Chromium via NPM.  There are no pre-built chromium-headless packages on NPM, and https://www.npmjs.com/package/chromium is the full version which requires X and its libraries (which we don't have in the tests container).

Thus it might be that we revert this again soon, but this unblocks https://github.com/cockpit-project/cockpit/pull/8069

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -7,6 +7,7 @@ RUN dnf -y update && \
         bind-utils \
         bodhi-client \
         bzip2 \
+        chromium-headless \
         copr-cli \
         curl \
         expect \


martinpitt:
I built and uploaded a new cockpit-tests container with this change, rolled out to infra now. If this breaks anything, I'll immediately revert the container to the previous :latest tag. Adding the package itself seems fairly harmless, but it could be that some Fedora updates have changed behaviour.

martinpitt:
Tests in https://github.com/cockpit-project/cockpit/pull/8067 and https://github.com/cockpit-project/cockpit/pull/8070 ran successfully with this container, and chromium-headless tests in https://github.com/cockpit-project/cockpit/pull/8069 at least start to run  in general, so this seems fine.

PR #156, opened 2017-11-14
Update containers to Fedora 26, optimize layering

martinpitt:
Now that Fedora 27 is released, Fedora 25 will be EOL really soon. Update our containers to Fedora 26.

While I was at it, I also optimized the tests container's layering to save half a Gigabyte.

I locally ran a build and test/verify/run-tests successfully.

TODO:

 - [x] rebuild and verify release container (also take PR #146 into account)
 - [x] rebuild and verify images container
 - [x] roll out new images and tests containers to OpenShift

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:25
+FROM fedora:26
 MAINTAINER "Stef Walter" <stefw@redhat.com>
 
 RUN dnf -y update && \
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,8 +1,7 @@
 FROM cockpit/infra-base
 MAINTAINER "Stef Walter" <stefw@redhat.com>
 
-RUN dnf -y update
-RUN dnf -y install nginx
+RUN dnf -y update && dnf -y install nginx && dnf clean all
 
 RUN mkdir -p /usr/local/bin && \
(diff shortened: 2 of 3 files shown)


martinpitt:
Local cockpit/release build succeeded, and from a coarse inspection (running release-* with `--help` and such) it seems to work fine. I'll handle the actual rollout of this in PR #146, and keep this for cockpit/tests.

martinpitt:
As Fedora 27 is released now, I tried to bump the containers to 27 right away, but `npm install -g phantomjs-prebuilt` fails with some weird linking error. Chances are that we don't need to debug this any more when F26 goes EOL, so I keep this on Fedora 26 for now.

I also did the same layering optimization for the images container. It makes quite some difference: Previous cockpit/images container is > 1 GB, this one is now 380 MB. [compressed size](https://hub.docker.com/r/cockpit/images/tags/) is 469 MB → 130 MB.

I now also verified that cockpit/images works fine, it's running now on https://cockpit-11.e2e.bos.redhat.com:8493/

I now pushed the base, images, and tests container. As said, I'll update release through PR #146.

martinpitt:
https://github.com/cockpit-project/cockpit/pull/8099 completely and successfully ran on the new container (a few others as well). So this is ready to land, the changes are all deployed and working.

PR #157, opened 2017-11-15
tests: Increase SHM size

martinpitt:
Docker containers only get a 64 MB /dev/shm by default, which cannot be
increased through OpenShift [1]. The recommendation is to use volumes
instead, so just use the host's /dev/shm. This will enable us to use
chromium-headless or electron in our tests docker containers, which use
shared memory to exchange data between processes. Some of these, like
HTML dumps or screenshots are rather big, and multiple parallel
processes often exceed the 64 MB limit.

For the standalone cockpit-tests.service, bump the SHM size to 500MB.

[1] https://docs.openshift.com/container-platform/3.6/dev_guide/shared_memory.html

### DIFF
--- a/tests/cockpit-tasks.json
+++ b/tests/cockpit-tasks.json
@@ -49,6 +49,11 @@
                                         "readOnly": false
                                     },
                                     {
+                                        "name": "shm",
+                                        "mountPath": "/dev/shm",
+                                        "readonly": false
+                                    },
+                                    {
                                         "name": "tmp",
                                         "mountPath": "/tmp",
                                         "readonly": false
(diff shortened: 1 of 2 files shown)


martinpitt:
I rolled this out on the infrastructure together with PR #156 and verified that /dev/shm has plenty of space now in the cockpit/tests containers.

martinpitt:
This did the trick, the chromium-headless test now [completed to run](https://fedorapeople.org/groups/cockpit/logs/pull-8069-20171115-100202-d47049ae-verify-fedora-27/log.html).

martinpitt:
[Spike: Evaluate electron and chromium-headless as PhantomJS replacement](https://trello.com/c/agwNN89E/559-spike-evaluate-electron-and-chromium-headless-as-phantomjs-replacement)

PR #158, opened 2017-11-15
Modernize dockerfiles

TomasTomecek:
image: apply dockerfile good practices
tests: invoke container in exec form
tests: chain metadata instructions
tests: chain run instructions
tests: don't install present packages
base: chain all run instructions
base: tar, util-linux are in base image
dockerfile: label maintainer > MAINTAINER instr
base: use latest stable Fedora: 26

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -15,5 +15,5 @@ RUN dnf -y update && \
         util-linux \
         && \
     dnf clean all
-RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user
-RUN mkdir -p /home/user/.ssh && mkdir -p /usr/local/bin
+RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user && \
+    mkdir -p /home/user/.ssh && mkdir -p /usr/local/bin
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -1,5 +1,5 @@
 FROM fedora:26
-MAINTAINER "Stef Walter" <stefw@redhat.com>
+LABEL maintainer='"Stef Walter" <stefw@redhat.com>'
 
 RUN dnf -y update && \
     dnf -y install \
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,5 +1,5 @@
 FROM cockpit/infra-base
-MAINTAINER "Stef Walter" <stefw@redhat.com>
+LABEL maintainer='"Stef Walter" <stefw@redhat.com>'
 
 RUN dnf -y update && dnf -y install nginx && dnf clean all
 
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,7 +1,9 @@
 FROM cockpit/infra-base
 LABEL maintainer='"Stef Walter" <stefw@redhat.com>'
 
-RUN dnf -y update && dnf -y install nginx && dnf clean all
+RUN dnf -y update && \
+    dnf -y install nginx && \
+    dnf clean all
 
 RUN mkdir -p /usr/local/bin && \
     ln -sf /dev/stdout /var/log/nginx/access.log && \
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -13,8 +13,7 @@ RUN mkdir -p /usr/local/bin && \
 ADD install-service /usr/local/bin/
 ADD nginx.conf /etc/nginx/
 
-VOLUME /secrets
-VOLUME /cache
+VOLUME /secrets /cache
 
 EXPOSE 80 443
 STOPSIGNAL SIGQUIT
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -12,8 +12,8 @@ RUN mkdir -p /usr/local/bin && \
     ln -snf /cache/images /usr/share/nginx/html && \
     chown -R user /home/user
 
-ADD install-service /usr/local/bin/
-ADD nginx.conf /etc/nginx/
+COPY install-service /usr/local/bin/
+COPY nginx.conf /etc/nginx/
 
 VOLUME /secrets /cache
 
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -90,7 +90,7 @@ VOLUME /secrets /cache /tmp
 
 USER user
 WORKDIR /build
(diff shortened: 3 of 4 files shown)


martinpitt:
@TomasTomecek : Thanks for the cleanup! However, can you please rebase on #156 ?  This is going  in today, and already moves to Fedora 26 and optimzies the layering wrt. the  dnf commands.

tests/Dockerfile
```diff
@@ -6,17 +6,13 @@ RUN dnf -y update && \
         american-fuzzy-lop \
         bind-utils \
         bodhi-client \
-        bzip2 \
```
martinpitt: There is no harm if the specified packages already exist, and it's always better to state explicitly what we need than making assumptions on what the base image provides.  This is fine for required packages like util-linux, but not for optional packages like curl. Please drop this commit.

TomasTomecek:
Will update tomorrow.

martinpitt:
#156 landed, so this can and needs to be rebased now.

TomasTomecek:
updated

base/Dockerfile
```diff
@@ -14,6 +14,6 @@ RUN dnf -y update && \
         tar \
         util-linux \
         && \
-    dnf clean all
-RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user
-RUN mkdir -p /home/user/.ssh && mkdir -p /usr/local/bin
+    dnf clean all && \
+    groupadd -g 1111 -r user && useradd -r -g user -u 1111 user && \
+    mkdir -p /home/user/.ssh && mkdir -p /usr/local/bin
```
martinpitt: What's the rationale for this? Chaining the groupadd and mkdir stuff is fine, but IMHO it makes sense to keep the "quick & small" setup bits in a separate RUN command (and thus layer). That way, it's much faster to iterate and build a new container when changing those, as the docker layer caching can re-use the expensive "install package" step.

tests/Dockerfile
```diff
@@ -47,16 +45,17 @@ RUN dnf -y update && \
         tar \
         virt-install \
         zanata-client && \
-        dnf -y builddep /tmp/cockpit.spec && dnf clean all
-RUN npm -g install phantomjs-prebuilt
+    npm -g install phantomjs-prebuilt && \
+    curl -s -o /tmp/cockpit.spec https://raw.githubusercontent.com/cockpit-project/cockpit/master/tools/cockpit.spec && \
+    dnf -y builddep /tmp/cockpit.spec && dnf clean all && \
+    mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/github /build/libvirt/qemu
```
martinpitt: The `mkdir` should move into the "quick & cheap" second RUN command below, while we are at it.

images/Dockerfile
```diff
@@ -1,20 +1,20 @@
 FROM cockpit/infra-base
 LABEL maintainer="'Stef Walter' <stefw@redhat.com>"
 
-RUN dnf -y update && dnf -y install nginx && dnf clean all
-
-RUN mkdir -p /usr/local/bin && \
+RUN dnf -y update && \
+    dnf -y install nginx && \
+    dnf clean all && \
+    mkdir -p /usr/local/bin && \
```
martinpitt: OK, installing one package is cheap enough to do it in one RUN step.

images/Dockerfile
```diff
@@ -1,20 +1,20 @@
...
     ln -sf /dev/stdout /var/log/nginx/access.log && \
     ln -sf /dev/stderr /var/log/nginx/error.log && \
     rm -rf /usr/share/nginx/html && \
     ln -snf /cache/images /usr/share/nginx/html && \
     chown -R user /home/user
 
-ADD install-service /usr/local/bin/
-ADD nginx.conf /etc/nginx/
+COPY install-service /usr/local/bin/
+COPY nginx.conf /etc/nginx/
```
martinpitt: Can you please add some explanation to the commit message why COPY is preferrable over ADD? The [documentation](https://docs.docker.com/engine/reference/builder/#add) does not discuss that.

TomasTomecek: I can. COPY is more straightforward -- it only operates on files, while ADD performs magic (downloads files, unpacks archives), therefore it's preferred to use COPY and use ADD only when needed.

TomasTomecek:
Seems that you don't like the simplification and chaining of the run commands -- I'll remove it. If that's your preferred workflow, iterate quickly locally, then the current solution is just fine. My point with this was that I watched Stef pulling `cockpit/tests` and the amount of layers just scared me.

martinpitt:
@Tomas: Thanks for clarifying ADD vs. COPY, makes sense! Also, I *do* like layer reduction in general, so please don't remove all of it - I was just pointing out why splitting into two layers generally makes sense - one which just installs packages and dnf clean (expensive, worth caching) and one which does all other setup (but is quick and doesn't pull in much data).

TomasTomecek:
@martinpitt I hope I addressed all your comments.

tests/Dockerfile
```diff
@@ -47,16 +45,20 @@ RUN dnf -y update && \
...
         virt-install \
         zanata-client && \
-        dnf -y builddep /tmp/cockpit.spec && dnf clean all
-RUN npm -g install phantomjs-prebuilt
+    npm -g install phantomjs-prebuilt
+
+# conserving layers, that's why we are not using ADD
+RUN curl -s -o /tmp/cockpit.spec https://raw.githubusercontent.com/cockpit-project/cockpit/master/tools/cockpit.spec && \
+    dnf -y builddep /tmp/cockpit.spec && \
+    dnf clean all
```
martinpitt: This is actually a case where you do need to merge the `RUN` commands. Running `dnf clean all` in a separate one makes it useless, as all the yum caching is already in the previous image layer. That's why I previously rearranged it like that. Calling curl in the middle instead of the `ADD` is fine too, though.

TomasTomecek:
Thanks Martin!

martinpitt:
I built, tested, pushed, and rolled out all four containers. Looking good, thanks!

TomasTomecek:
Thank you! Please let me know if you run into some issues.

martinpitt:
@TomasTomecek : The bots have run with the new container for almost a day now, and seem happy.

PR #162, opened 2017-12-01
tests: Install latest chromium-headless from updates-testing

martinpitt:
Current chromium 61 has a bug in `Network.setCookie()` which is hard to
work around and fixed in 62. Install chromium-headless from
updates-testing, which gets current upstream versions much more timely.


martinpitt:
I built, tested, pushed, and rolled out a new tests container with this.

PR #163, opened 2017-12-01
tests: Include python3 in the tests container

stefwalter:
This blocks https://github.com/cockpit-project/cockpit/pull/8193

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -26,6 +26,7 @@ RUN dnf -y update && \
         libvirt \
         libvirt-client \
         libvirt-python \
+        libvirt-python3 \
         mock \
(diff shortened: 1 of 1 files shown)


stefwalter:
Rebased

stefwalter:
Container is up ... we really do need a better way to try out contianers in the infra. They cycle by themselves eventually.

martinpitt:
I tested mine this morning with starting it with /bash, building master, an image, and running a few verify tests. Indeed it's a bit laborious, but fortunately doesn't happen that often to become a real nuisance. Once it does, we could teach the semaphore test for cockpituous to do these steps for us, and even do the upload too?

PR #164, opened 2017-12-06
tests: Install full chromium and Xvfb

martinpitt:
Headless Chrome does not handle requests for client SSL certificates
(https://bugs.chromium.org/p/chromium/issues/detail?id=757181) and there
is no known workaround. These are issued by OpenShift's OAuth during
Cockpit's container tests. Install the full chrome and Xvfb for the
time being so that we can run the container tests under "full" chrome.
(Other tests can continue to use `--headless`).

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -46,8 +46,9 @@ RUN dnf -y update && \
         sudo \
         tar \
         virt-install \
+        xorg-x11-server-Xvfb \
         zanata-client && \
-    dnf -y install chromium-headless --enablerepo=updates-testing && \
+    dnf -y install chromium --enablerepo=updates-testing && \
     npm -g install phantomjs-prebuilt && \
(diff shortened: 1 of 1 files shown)


martinpitt:
As expected this drags in quite a lot of additional stuff:
```
docker.io/cockpit/tests                         2017-12-06          435647b8a9df        About an hour ago   2.799 GB
docker.io/cockpit/tests                         latest              1c149f238bfd        4 days ago          2.573 GB
```
I. e. +250 MB uncompressed, but less than 100 MB compressed (still 1 GB on https://hub.docker.com/r/cockpit/tests/tags/).

I verified that tests work in the new container, and with additional changes to use xvfb-run and full chromium the container tests now generally work as well. I pushed a new container with this change to dockerhub, and rolled it out into our infrastructure.

martinpitt:
This has been running in production for 5 days and got approved, so I merge this now.

PR #165, opened 2017-12-07
systemd: Avoid container startup failure on stale Docker containers

martinpitt:
Clean up a stale `cockpit-{tests,images,release}` at boot before
starting a new one, to avoid failed units and thus non-running image
containers after an unclean reboot.

---

This often happened on verifymachine4/5 and thus public image servers didn't exist until someone cleaned up manually. I tested and rolled out the changes there.


martinpitt:
This has been running in production for 5 days and got approved, so I merge this now.

PR #166, opened 2017-12-08
sink: Fix TypeError in GitHub/HTTP logic

stefwalter:
This fixes the following error when using watches:

TypeError: unsupported operand type(s) for +: 'NoneType' and 'list'


stefwalter:
Deployed to verifymachine4

PR #168, opened 2018-01-08
release-source: Generate separate patch for build system changes

martinpitt:
The previous approach of including the build system changes into every
single patch has several drawbacks:

 - It is susceptible to file timestamps being out of order, i. e.
   sometimes generated files get patched before source files. This
   causes build failures as webpack et al are not available in
   downstream build environments.

 - It makes patches unnecessarily big (due to duplicated patches to
   generated files) and unreadable.

 - It makes patch generation unnecessarily slow, as each new patch
   requires a full build cycle. This previously mis-motivated us in the 
   past to merge several upstream patches into one downstream patch
   instead of using clean cherry-picks.

Change and simplify the approach to use the upstream patches verbatim,
and generate a new 0999-Build-system-generated-changes-from-patches.patch
on top of them. This will ensure that build system changes always get 
applied after source changes, and thus avoid timestamp sorting problems
that confuse make.


martinpitt:
Tested downstream release with patches like this:
```
$ git checkout rhel-7.5
$ mkdir ~/patches
# download 154 release and cache tarballs to ~/patches/, and munge broken paths in cockpit-cache-154.tar.xz
$ ../cockpituous/release/release-source -t 154 -f ~/patches/
$ ls -l ~/patches/
-rw-rw-r-- 1 martin mock       2240  8. Jan 12:02 0001-test-Fix-test-dbus-to-work-without-a-system-bus-7952.patch
-rw-rw-r-- 1 martin mock       8899  8. Jan 12:02 0002-test-Adjust-for-RHEL-7.5-image.patch
-rw-rw-r-- 1 martin mock       2988  8. Jan 12:02 0003-test-Disable-LVM-related-storaged-tests-on-RHEL-7.5.patch
-rw-rw-r-- 1 martin mock      17793  8. Jan 12:02 0004-kubernetes-Updates-to-work-with-3.7.patch
-rw-rw-r-- 1 martin mock     171279  8. Jan 12:02 0005-po-Import-Japanese-translation-updates-from-RHEL-7.5.patch
-rw-rw-r-- 1 martin mock        841  8. Jan 12:02 0006-bridge-Forward-kill-messages.patch
-rw-rw-r-- 1 martin mock     379948  8. Jan 12:02 0999-Build-system-generated-changes-from-patches.patch
-rw-rw-r-- 1 martin martin 12244828  8. Jan 10:27 cockpit-154.tar.xz
-rw-rw-r-- 1 martin martin 21886392  8. Jan 10:32 cockpit-cache-154.tar.xz
```
Inspecting the *.patches to make sure they look alright.

I'm going to replace the current rhel-7.5 downstream patches with these generated ones. In the current downstream ones I had to carefully dissect the upstream ja.po changes as they caused build failures due to timestamp issues. But this confuses the downstream PO file scanner, so I'd rather have them back.

Tested an upstream release (without patches) like this:
```
$ git tag -s 159    # don't push this!
$ ../cockpituous/release/release-source -f ~/rel
$ ls -l ~/rel
-rw-rw-r-- 1 martin mock 13486872  8. Jan 12:11 cockpit-159.tar.xz
-rw-rw-r-- 1 martin mock 21370464  8. Jan 12:11 cockpit-cache-159.tar.xz
```
and inspecting the tarballs to make sure they look alright.

stefwalter:
I'm fine with this, seems simpler and easier to understand. It may need further revision if/when you get push back from release engineering or others in the delivery chain.

release/release-source
```diff
@@ -135,101 +136,70 @@ prepare()
...
+    git -C $stagedir config core.autocrlf false
+    git -C $stagedir config core.safecrlf false
+    git -C $stagedir config gc.auto 0
+
+    # Mark all sorts of extra files as binary for diffing purposes
+    printf "*.min.* binary\n*.map binary\n/dist/** binary\n" > $stagedir/.git/info/attributes
+
+    # HACK: Exclude node_modules/ which was in cockpit tarballs
+    tar -C $stagedir --exclude="node_modules" \
+        --strip-components=1 -xf "$archive"
```
stefwalter: Is this still relevant? I think our node_modules in cockpit tarballs is now a small manageable size, right?

release/release-source
```diff
@@ -135,101 +136,70 @@ prepare()
...
+        git format-patch --stdout $TAG.. | git -C $stagedir am -
+
+        git -C $repodir checkout -q --detach "$head"
+        trace "Cleaning out build and rebuilding with commits on top of $TAG"
+
+        rm -rf $repodir/_build
+        autogen_or_configure $repodir
+        $MAKE -C $repodir/_build dist-gzip distdir=release-patched
+
+        trace "Committing tarball"
```
stefwalter: Is this line still correct?

martinpitt:
Force-pushed to adjust the comments to new reality.

release/release-source
martinpitt: It's not, it's still 385 MB. We still have phantomjs in there, as the unit tests still need that (see https://github.com/cockpit-project/cockpit/pull/8314)

martinpitt:
Adding "needswork", I'm still testing/verifying this with downstream builds.

release/release-source
martinpitt: Right, updated.

martinpitt:
This is still not perfect. E. g. when we patch po/ja.po, our upstream build system does not change files like dist/pcp/po.ja.js, as they are empty and have no translations. Thus their mtime does not get bumped either, and the downstream build system tries to rebuild it. But this is incrementally better than what we have now.

I tested this now with a local mockbuild and a brew scratch build, using some more `touch`ing to work around the above issue. Thus removing "needswork" and added "blocked", as this needs to go through a downstream release first.

martinpitt:
I built a new release container with this PR, will land once the next Cockpit release succeeds.

martinpitt:
https://github.com/cockpit-project/cockpit/releases/tag/159 was successfully done with this change. I pushed the new image to [docker](https://hub.docker.com/r/cockpit/release/tags/).

PR #169, opened 2018-01-18
release: Clone specified repository instead of cockpit

martinpitt:
This allows release-runner to work for projects other than cockpit.
    
release/cockpit-release.service already passes in that repository
through `-r`, so for cockpit this works exactly as before.

### DIFF
--- a/release/README
+++ b/release/README
@@ -21,7 +21,7 @@ This is the container for the Cockpit release runner.
 
 First get the container:
 
-    # docker pull cockpit/infra-release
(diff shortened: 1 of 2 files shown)


martinpitt:
I tested this on welder-web, see https://github.com/weldr/welder-web/pull/187

martinpitt:
https://github.com/cockpit-project/cockpit/releases/tag/161 was done with this, so I pushed the new container to dockerhub.

PR #171, opened 2018-02-08
tests: Clean the tree after resetting

martinpitt:
This will guarantee that we don't have any built leftovers from previous
runs. This becomes even more important when testing third-party modules.


martinpitt:
I'm building and deploying a new container with this now, let's merge after it survived a few bot runs.

martinpitt:
Our cluster runs the new container, and correctly triggers and runs tests, so from my POV this is good to go.

martinpitt:
Approved and rolled out already, so I'll just self-land this.

PR #172, opened 2018-02-15
Update containers to Fedora 27, remove cruft

martinpitt:
See individual commits.  @stefwalter, this includes dropping the pbuilder bits (we talked about that in Brno).

I locally tested the new tests container with a full build/image-prepare/test run, deployed it on verifymachine4, and re-ran it there. Same for cockpit/images, it's running on [verifymachine4](https://209.132.184.41:8493/).

I also gave the rebuilt cockpit/release container a shallow testing. The full run will happen on Wednesday when we release Cockpit 162. Until then it's deployed on verifymachine, but not pushed to dockerhub yet. Contingency plan is to revert to the previous version (i. e. current on dockerhub), and re-run the release script.


martinpitt:
Tests in https://github.com/cockpit-project/cockpit/pull/8620 ran with this successfully (modulo some flakes due to https://github.com/cockpit-project/cockpit/issues/8605, but this affected the previous container too). Same for https://github.com/cockpit-project/cockpit/pull/8548. In addition, tests in https://github.com/cockpit-project/cockpit-examples/pull/6 now actually run instead of failing due to missing async/await. So I think this is sufficient proof.

PR #173, opened 2018-02-27
release-runner: Fix -t option

martinpitt:
Declare -t option to have an argument, otherwise it fails with

    release/release-runner: line 153: OPTARG: unbound variable


martinpitt:
We (obviously) don't use this option in production, but I stumbled over this when running this on a local archive.

PR #175, opened 2018-03-06
release: Replace release polling with GitHub trigger

martinpitt:
So far cockpit-release.service restarted itself every 30 minutes to
poll if a new release is pending. This is a waste most of the time, and
an annoying delay when a release actually happens.

Make this unit start via a HTTP request instead, so that it becomes
suitable as a GitHub "create" web hook, and generalize the mechanism to
work with more than just cockpit:

Add a releasetrigger.socket that listens on port 8090 and activates
releasetrigger@.service. This receives a HTTP request and uses the
path to select the project (sanitizing it to just alphanumeric
characters). If `<project>-release.service` exists, start that and
report success, otherwise report a 404.

There is no user input aside from the sanitized HTTP path, so at most a
remote can try to DoS. Mitigate this by limiting cockpit-release.service
to three starts a day, and if it is already running trying to start it
again is a no-op.

Drop the `make release-install` dependency to `release-container`. There
is no need to build a container image, we want to user the one from
dockerhub, as documented in README.

Update the documentation accordingly. Refer to `make release-install`
instead of replicating the setup commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,10 +54,10 @@ release-container: docker-running
 release-push: docker-running
 	base/push-container docker.io/cockpit/release
 
-release-install: release-container
-	cp release/cockpit-release.service /etc/systemd/system/
+release-install:
+	cp release/releasetrigger.socket release/releasetrigger@.service release/cockpit-release.service /etc/systemd/system/
 	systemctl daemon-reload
-	systemctl enable cockpit-release
+	systemctl enable --now releasetrigger.socket
 
 tests-shell: docker-running
 	docker run -ti --rm \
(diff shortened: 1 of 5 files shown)


martinpitt:
While testing this, I rolled this out on our Cockpit release machine and I added the webhook. I can test this on tomorrow's release, but it can be trivially tested with curl (and adding the webhook in github also correctly launches cockpit-release.service - it just immediately exits again as there are no new tags, as expected).

Setting to blocked until after a successful upstream release, as usual.

This will be doubly useful once I set up a release service for welder-web, which is this →← close.

martinpitt:
I just pushed the tag for 163, and the release container got triggered properly. So this has proven itself in production, removing blocked tag.

PR #176, opened 2018-03-09
tests: Make deployable in restricted Openshift SCC

stefwalter:
Many Openshift clusters don't give us anything but the restricted
SCC, which doesn't allow host mounts. We typically use host mounts
to acheive high density of bots, but when these are not available
we have to fall back to other methods.

Includes:

 * A new pod definition for the restricted SCC case
 * Don't use sudo when unpredictable Openshift UIDs are used
 * Setup $HOME directory properly with unpredictable UIDs
 * Give access to the right files with unpredicable UIDs

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -74,7 +74,7 @@ RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/gith
     ln -snf /secrets/zanata.ini /home/user/.config/zanata.ini && \
     git clone https://github.com/cockpit-project/cockpit /build/cockpit && \
     cd /build/cockpit && npm install && \
-    chown -R user /build /secrets /cache /home/user
+    chmod -R ugo+w /build /secrets /cache /home/user
 
 # Prevent us from screwing around with KVM settings in the container
 RUN touch /etc/modprobe.d/kvm-amd.conf && touch /etc/modprobe.d/kvm-intel.conf
@@ -87,7 +87,8 @@ ENV LIBGUESTFS_BACKEND=direct \
     TMPDIR=/tmp \
     TEMPDIR=/tmp \
     TEST_DATA=/build \
-    TEST_PUBLISH=
+    TEST_PUBLISH= \
+    HOME=/home/user
 
 VOLUME /secrets /cache /tmp
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -83,7 +83,19 @@ The testing machines can run on Openshift cluster(s).
 
 Create a service account for use by the testing machines. Make sure to have the
 ```oci-kvm-hook``` package installed on all nodes.  This is because of the requirement
-to access ```/dev/kvm```. Further work is necessary to remove this requirement.
+to access ```/dev/kvm```.
+
+This creates all the remaining kubernetes objects. The secrets are created from the
+```/var/lib/cockpit-tests/secrets``` directory as described above.
+
+    $ sudo make tests-secrets | oc create -f -
+    $ oc create -f tests/cockpit-tasks-restricted.json
+
+## High Density Openshift Deployment
+
+In order to deploy on Openshift at high density, shared host mounts between pods
+are necessary. To do this we need to have additional privileges, and the Openshift
+setup is different:
 
     $ oc create -f tests/cockpituous-account.json
     $ oc adm policy add-scc-to-user anyuid -z cockpituous
--- a/tests/cockpit-tasks-restricted.json
+++ b/tests/cockpit-tasks-restricted.json
@@ -1,0 +1,146 @@
+{
+    "apiVersion": "v1",
+    "kind": "List",
+    "items": [
+        {
+            "kind": "ReplicationController",
+            "metadata": {
+                "name": "cockpit-tasks"
+            },
+            "spec": {
+                "replicas": 1,
+                "selector": {
+                    "infra": "cockpit-tasks"
+                },
+                "template": {
+                    "metadata": {
+                        "labels": {
+                            "infra": "cockpit-tasks"
+                        }
+                    },
+                    "spec": {
+                        "containers": [
+                            {
+                                "name": "cockpit-tasks",
+                                "image": "docker.io/cockpit/tests:2018-03-08",
+                                "env": [
+                                    {
+                                        "name": "TEST_JOBS",
+                                        "value": "8"
+                                    },
+                                    {
+                                        "name": "TEST_PUBLISH",
+                                        "value": "sink"
+                                    }
+                                ],
+                                "volumeMounts": [
+                                    {
+                                        "name": "secrets",
+                                        "mountPath": "/secrets",
+                                        "readOnly": true
+                                    },
+                                    {
+                                        "name": "cache",
+                                        "mountPath": "/cache",
+                                        "readOnly": false
+                                    },
+                                    {
+                                        "name": "shm",
+                                        "mountPath": "/dev/shm",
+                                        "readonly": false
+                                    },
+                                    {
+                                        "name": "tmp",
+                                        "mountPath": "/tmp",
+                                        "readonly": false
+                                    }
+                                ]
+                            },
+                            {
+                                "name": "cockpit-images",
+                                "image": "docker.io/cockpit/images",
+                                "ports": [
+                                    {
+                                        "containerPort": 80,
+                                        "protocol": "TCP"
+                                    },
+                                    {
+                                        "containerPort": 443,
+                                        "protocol": "TCP"
+                                    }
+                                ],
+                                "volumeMounts": [
+                                    {
+                                        "name": "secrets",
+                                        "mountPath": "/secrets",
+                                        "readOnly": true
+                                    },
+                                    {
+                                        "name": "cache",
+                                        "mountPath": "/cache",
+                                        "readOnly": false
+                                    },
+                                    {
+                                        "name": "hack-hide-virtual-net",
+                                        "mountPath": "/sys/devices/virtual/net/",
+                                        "readOnly": true
+                                    }
+                                ]
+                            }
+                        ],
+                        "volumes": [
+                            {
+                                "name": "secrets",
+                                "secret": {
+                                    "secretName": "cockpit-tests-secrets"
+                                }
+                            },
+                            {
+                                "name": "cache",
+                                "emptyDir": { }
+                            },
+                            {
+                                "name": "shm",
+                                "emptyDir": {
+                                    "medium": "Memory"
+                                }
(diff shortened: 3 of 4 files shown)


tests/Dockerfile
```diff
@@ -74,7 +74,7 @@ RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/gith
     ln -snf /secrets/zanata.ini /home/user/.config/zanata.ini && \
     git clone https://github.com/cockpit-project/cockpit /build/cockpit && \
     cd /build/cockpit && npm install && \
-    chown -R user /build /secrets /cache /home/user
+    chmod -R ugo+w /build /secrets /cache /home/user
```
martinpitt: This should not have any effect for /cache and /secrets as get bind-mounted from the host/from a volume (even in your new cockpit-tasks-restricted.json). These directories should be completely empty during container creation.

/home/user is not a volume, but why does the  home dir need to be world-writable? Shouldn't the original `user` ownership of it be enough?

tests/Dockerfile
```diff
@@ -87,7 +87,8 @@ ENV LIBGUESTFS_BACKEND=direct \
     TMPDIR=/tmp \
     TEMPDIR=/tmp \
     TEST_DATA=/build \
-    TEST_PUBLISH=
+    TEST_PUBLISH= \
+    HOME=/home/user
```
martinpitt: OOI, how is this related to running the container on a different OpenShift? The change seems fine to me, I just wonder if I'm missing the impact of this.

tests/README.md
```diff
@@ -83,7 +83,19 @@ The testing machines can run on Openshift cluster(s).
...
-to access ```/dev/kvm```. Further work is necessary to remove this requirement.
+to access ```/dev/kvm```.
+
+This creates all the remaining kubernetes objects. The secrets are created from the
+```/var/lib/cockpit-tests/secrets``` directory as described above.
+
+    $ sudo make tests-secrets | oc create -f -
+    $ oc create -f tests/cockpit-tasks-restricted.json
+
+## High Density Openshift Deployment
```
martinpitt: This part would apply to non-restricted environments like our current cockpit openshift cluster? It's not entirely clear whether this is mutually exclusive with the above -restricted.json.

tests/cockpit-tasks-restricted.json
```diff
@@ -0,0 +1,146 @@
+{
```
martinpitt: For easier review, this is the [diff compared to tests/cockpit-tasks.json](https://paste.fedoraproject.org/paste/6l2fBfSuMXfEFScFGTaS4g). If we can make this service work without the `uid == 1111` assumption, is there any reason to not do this on our current Cockpit openshift cluster as well? I. e. also apply the `securityContext` and `hack-hide-virtual-net` changes to tests/cockpit-tasks.json?

AFAICS the only significant change is about not mounting the host's tests/cockpit-tasks.json, but using a regular docker volume for that as well. Likewise, this would also simplify things on our current OpenShift cluster, as we have to jump through quite a lot of hoops to make this work (remember to do the chown, apply the appropriate SELinux context, etc.). It seems to me that we should just use regular volumes there as well.

So in summary: I don't see anything in that diff that shouldn't apply to our current cluster as well - i. e. WDYT about patching tests/cockpit-tasks.json  instead and we roll this out on our current cluster? (I'm fine with hand-holding this and reverting if necessary).

tests/cockpit-tests
```diff
@@ -16,21 +16,21 @@ if [ $loop = "yes" ]; then
...
-sudo mkdir -p /cache/github /cache/images
-sudo chown user:user /cache /cache/github /cache/images
+mkdir -p /cache/github /cache/images
 
 # HACK: /sys is read-only in containers. And libvirt tries
 # to write there, but only if a specific subtree exists here
-sudo -n mount -o bind /run /sys/devices/virtual/net/ || true
+if [ -d /sys/devices/virtual/net/lo ]; then
+    sudo -n mount -o bind /run /sys/devices/virtual/net/ || true
+fi
```
martinpitt: To clarify, this will only actually run if you manually run a container on e. g. your own laptop. In OpenShift /net already gets bind-mounted to an empty directory, so this hack isn't necessary there, right?

tests/cockpit-tests
```diff
@@ -63,7 +63,8 @@ function task() {
 }
 
 # Perform N tasks or waits, then restart
-for i in $(seq 1 30); do
+TASKS=300
```
martinpitt: This also seems unrelated. This PR keeps the cache volume, so there's no precious state in these containers. It seems useful to not accumulate too much cruft in them, and update to the latest cockpit/tests tag relatively often. Was there a particular reason to change this?

tests/Dockerfile
stefwalter: When run in Openshift, the pod is run as a random uid. This means we don't have permissions to write to those directories. In order to cover *both* the bind-mounted use case and the random-uid use case this is a way to solve this.

The ```user``` account does not play in Openshift at all.

tests/Dockerfile
stefwalter: The ```user``` account does not work in Openshift without the ```anyuid``` SCC. In the ```restricted``` SCC the pod is run as a random user. However we have all our credentials in ```/home/user``` and the random user needs to be able to read them. So we use this  trick to point to the location of the credentials.

tests/README.md
stefwalter: I've run them both on the same cluster successfully. One simply uses less efficient mechanisms when we don't have the right SCC to efficiently cache images across pods.

tests/cockpit-tasks-restricted.json
stefwalter: It's just *way* less efficient. We can run the restricted one in our current cluster, but we could never achieve the density we do today.

tests/cockpit-tests
stefwalter: Right.

tests/cockpit-tests
stefwalter: Yes, it's related.

In addition to efficiency, we use the host caching to prevent constantly re-downloading images. However in the ```restricted``` case we cannot use that host caching. So the interval between downloading *all* the images again starts to be annoying. 

Hence the bumping up of this number, and I believe it's a reasonable tradeoff between the two factors: 
 * Avoiding cruft
 * Downloading everything constantly

tests/README.md
stefwalter: You can deploy the same stuff in the same cluster. Obviously the names conflict, but you can put it in another project.

tests/cockpit-tasks-restricted.json
martinpitt: So I'm obviously missing something here. Docker volumes get shared between all containers that use them. I. e. they have much the same behaviour as with our current host bind mount, except that the cached data isn't (easily) visible on the host (but that's not very important either). What is the principal difference in density here? Of course for this it should use an actual volume, not an `emptyDir`.

My understanding is that [OpenShift volumes](https://docs.openshift.com/container-platform/3.7/dev_guide/volumes.html) are pretty much exactly docker volumes.

Or is the thing that I'm missing that centos CI does not have volumes? I can't believe that, as (1) you are already using a secrets volume, and (2) such an OpenShift environment would be almost useless. I'm certainly not willing to move off to such a cluster with non-shared caches between pods, that would be insanely inefficient (not just for storage, also for test runtime).

tests/Dockerfile
martinpitt: I understand the random uid and the permission problem. I don't understand how this chown command has any visible effect. This happens on the *unmounted* mount point directory during container build, doesn't  it? Once a bind mount is done onto that dir, the original permissions don't matter at all - what's relevant is the permissions of the root directory of the *mounted* device. I. e. what matters is the permissions of the files in the mounted volume, not of this mount point.

tests/cockpit-tests
martinpitt: I think this is just a follow-up from the discussion above, so let's ignore that for now. (But to reiterate: pods not sharing volumes sounds like a really bad idea IMHO)

tests/cockpit-tasks-restricted.json
stefwalter: CentOS Ci has volumes, just no /dev/kvm support (like oci-kvm-hook yet).

Openshift volumes have semantics that require strange fsGroup or a number of optional techniques in order to get containers to write to them. 

Openshift went and made docker containers run under a random UID, which has *all* sorts of strange implications. In this case we don't need volumes, since we're just using this stuff as a cache. Booting a VM image from a Gluster or NFS volume is a fail in any case. We really just want the container to pull images to an emptyDir and use that.

tests/cockpit-tests
stefwalter: I don't see why they would need to share volumes. The various pods should be organic, distributed and able to operate on their own. The fact that they do share volumes was purely an optimization for density.

tests/cockpit-tasks-restricted.json
martinpitt: I don't understand - why wouldn't we want to use a shared volume? Your intent with the `chmod a+w` is obviously to make the cache world-writable (except that happens in the wrong place - it needs to happen when the cache dirs are written, not in the static container image). With random UIDs you can indeed not do any better than making volumes world-writable (but also, it's fairly irrelevant in this environment).

tests/cockpit-tests
martinpitt: IMHO not having ten copies of our 54 GB (!) cache directory on one and the same node is not an "optimziation" - it would be a giant waste to not do it. Also note that this would make pods prone to multiple parallel downloads, which we've just recently [seen to fall over horribly](https://github.com/cockpit-project/cockpit/pull/8740#issuecomment-369581186). This got [fixed](https://github.com/cockpit-project/cockpit/pull/8741), but not sharing a cache directory would bring back this problem.

So I'd ask it the other way around: why would we *nod* use a volume for the cache? That doesn't preclude them from being organic or independent in any way.

martinpitt:
This looks like the last remaining blocker for running pods on CentOS CI: https://bugs.centos.org/view.php?id=14915

tests/Dockerfile
martinpitt: Bwah, of course this is because it is now *chmod*, not *chown*. "Wer lesen kann, ist klar im Vorteil!" Sorry! :-)

martinpitt:
I took the necessary container changes from this PR, added a few extra fixes (like setting up a passwd entry so that ssh actually works) and cleanups, and sent it as PR #186. We can then rebase this PR and use it for the openshift  YAML resources only, without having to rebuild the container.

PR #179, opened 2018-03-16
release: Add welder-web-release.service

martinpitt:
This is fairly similar to cockpit-release.service, but configures the
git repository and cockpituous script path for welder-web.


martinpitt:
This is already deployed on our release machine and was used to release https://github.com/martinpitt/welder-web/releases/tag/0.0.1 . So this already passed the "production" check.

PR #183, opened 2018-06-04
release: Add resources and webhook for running in OpenShift

martinpitt:
Add a script to turn a home directory with all the necessary credentials
into a Kubernetes secrets volume.

Add a GitHub webhook (using only Python 3 included modules) that can 
release any project with the Cockpituous release script. The git repo
URL is taken from the requests's GitHub metadata, and the path to the 
release script is taken from the URL path.

Document how to deploy this.

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -29,7 +29,7 @@ This is the container for the Cockpit release runner. It normally gets
 activated through a HTTP request: <http://host:8090/cockpit>. The "/cockpit"
 path specifies the systemd service name to start (<name>-release.service).
 
-## How to deploy
+## Deploying on a host
 
 Setup a 'cockpit' user:
 
@@ -70,4 +70,36 @@ Follow the logs of a running release:
 Start the container manually (without a webhook):
 
     # systemctl start cockpit-release
+
+# Deploying on OpenShift
+
+On some host `$SECRETHOST`, set up all necessary credentials as above, plus an
+extra `~/.config/github-webhook-token` with a shared secret.
+
+Then build a Kubernetes secret volume definition on that host, copy that to a
+machine that administers your OpenShift cluster, and deploy that secret volume:
+
+    ssh $SECRETHOST release/build-secrets | oc create -f -
+
+Then deploy the other objects:
+
+    oc create -f release/cockpit-release.yaml
+
+This will create a POD that runs a simple HTTP server that acts as a
+[GitHub webhook](https://developer.github.com/webhooks/). Set this up as a
+webhook in GitHub, using an URL like
+
+   http://release-cockpit.apps.ci.centos.org/bots/major-cockpit-release
+
+using the path to the release script of the corresponding project's git tree
+(the git repository URL will be taken from the POST data that GitHub sends).
+Use the same secret as in `~/.config/github-webhook-token` above. Make sure to
+change the Content Type to `application/json`.
+
+To remove the deployment:
+
+    oc delete service cockpit-release
+    oc delete route release
+    oc delete pod release
+    oc delete secrets cockpit-release-secrets
 
--- a/release/cockpit-release.yaml
+++ b/release/cockpit-release.yaml
@@ -1,0 +1,48 @@
+---
+apiVersion: v1
+kind: List
+items:
+- kind: Pod
+  apiVersion: v1
+  metadata:
+    name: release
+    labels:
+      infra: cockpit-release
+  spec:
+    containers:
+      - name: release
+        image: cockpit/release
+        ports:
+          - containerPort: 8080
+            protocol: TCP
+        command: [ "webhook" ]
+        volumeMounts:
+        - name: secrets
+          mountPath: /run/secrets/release
+          readOnly: true
+    volumes:
+    - name: secrets
+      secret:
+        secretName: cockpit-release-secrets
+
+- kind: Service
+  apiVersion: v1
+  metadata:
+    name: cockpit-release
+  spec:
+    clusterIP: None
+    selector:
+      infra: cockpit-release
+    ports:
+    - name: webhook
+      port: 8080
+      protocol: TCP
+
+- kind: Route
+  apiVersion: v1
+  metadata:
+    name: release
+  spec:
+    to:
+      kind: Service
+      name: cockpit-release
--- a/release/Dockerfile
+++ b/release/Dockerfile
@@ -21,12 +21,14 @@ gnupg \
 hardlink \
 krb5-workstation \
 nc \
-npm tar \
+npm \
+origin-clients \
 psmisc \
 python \
 python-irclib \
 reprepro \
 rpm-build \
+tar \
 which \
 yum-utils \
     && \
--- a/release/Dockerfile
+++ b/release/Dockerfile
@@ -42,6 +42,11 @@ yum-utils \
 
 ADD * /usr/local/bin/
 
+# HACK: Working around Node.js screwing around with stdio
+ENV NODE_PATH /usr/bin/node.real
+RUN mv /usr/bin/node /usr/bin/node.real
+ADD node-stdio-wrapper /usr/bin/node
+
 VOLUME /home/user /build
 WORKDIR /build
 USER user
--- a/release/Dockerfile
+++ b/release/Dockerfile
@@ -35,6 +35,7 @@ yum-utils \
     dnf -y builddep /tmp/cockpit.spec && \
     dnf clean all && \
     mkdir -p /usr/local/bin /home/user /build/rpmbuild && \
+    chmod g=u /etc/passwd && \
     chown -R user /build /home/user
 
 ADD * /usr/local/bin/
--- a/release/build-secrets
+++ b/release/build-secrets
@@ -1,0 +1,16 @@
+#!/bin/sh
+# Run this in a home directory with all credentials; write OpenShift secrets
+# volume JSON definition to stdout
+# https://docs.openshift.com/container-platform/3.9/dev_guide/secrets.html
+set -eu
+
+printf '{ "apiVersion": "v1", "kind": "Secret", "metadata": { "name": "cockpit-release-secrets" }, "data": {\n'
+
+first=yes
+cd ~
+for f in $(find .ssh .gnupg .config .pki .fedora* .gitconfig -type f); do
+    [ -n "$first" ] || printf ',\n'
+    printf '\t"%s": "%s"' "$(echo $f | sed "s!/!--!g")" "$(base64 --wrap=0 $f)"
+    first=''
+done
+printf '\n} }\n'
--- a/release/node-stdio-wrapper
+++ b/release/node-stdio-wrapper
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# This is certified, A-grade, high quality turd polish. Node leaves
+# stdio file descriptors in non-blocking mode when exiting. This has
+# been reported, fixed, unfixed, ad nauseum.
+#
+# https://github.com/nodejs/node/issues/14752
+# https://github.com/nodejs/node/pull/17737
+# https://github.com/nodejs/node/pull/20592
+# https://github.com/nodejs/node/pull/21257
+#
+# It's starting to fester.
+
+OUT=$(/usr/bin/node.real "$@" 2>&1 </dev/null)
+ret=$?
+echo -n "$OUT"
+exit $ret
--- a/release/webhook
+++ b/release/webhook
@@ -1,0 +1,152 @@
+#!/usr/bin/python3
+
+import os
+import hmac
+import logging
+import json
+import subprocess
+import shutil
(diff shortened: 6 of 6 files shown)


martinpitt:
I rebuilt the [release container](https://hub.docker.com/r/cockpit/release/tags/) with these changes. I tested it on my own [martinpitt/cockpit](https://github.com/martinpitt/cockpit) project, i. e. I added the webhook there, and deployed it on CentOS CI. After pushing a tag, the release runner started, and it got as far as doing a koji scratch build. ssh to the sink also works, so the credentials generally work. ([log from a second run](https://fedorapeople.org/groups/cockpit/logs/release-169/log) which I aborted very early).

I will use this for the next Cockpit release and see how it goes. I deactivated the current webhook that triggers releasemachine (with the old systemd activation), and added a hook to the new CentOS CI deployment. If this breaks, I can trivially trigger the release on the old infrastructure (with a simple curl).

martinpitt:
Force-pushed to amend the documentation (the webhook content is expected in JSON format).

martinpitt:
Note: This does not use SSL for the route. I didn't find a way to enable that on CentOS CI. On the OpenShift web console you can manually provide a certificate, but that would require getting a custom one, or controlling DNS. Maybe in the future we can use our own domain name (like release-ci.cockpit-project.org) and a LetsEncrypt cert.

But this isn't a blocker really. The information that goes on the wire is not private, and it gets cryptographically signed by GitHub with the shared secret, so we do have authentication and authorization. Our current GitHub has no validation at all, everyone in the world can trigger it (although with almost no input and much stronger validation).

release/cockpit-release.yaml
```diff
@@ -0,0 +1,48 @@
...
+    containers:
+      - name: release
+        image: cockpit/release
+        ports:
+          - containerPort: 8080
+            protocol: TCP
+        command: [ "webhook" ]
+        volumeMounts:
+        - name: secrets
+          mountPath: /run/secrets/release
```
stefwalter: Does this need container need access to all the secrets? Do you feel that the webhook script is sufficiently secure to prevent leaking the secrets? Should we have the webhook in its own container, which triggers a Kube/Openshift "Job" in order to do its work?

I know this is significantly different, and this problem may already exist with the current deployment.

martinpitt: Thanks for pointing out [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)! In fact this is exactly what I was looking for, when googling something like "kubernetes service create pod on demand" (all existing hits I found just said "there is no such concept"). I don't particularly mistrust the HTTP server implementation, as it does not have any responses that are composed of file system items, but nevertheless this sounds like a better approach all around. I'm not very happy to have the release container run all the time - it should spin up, do *one* release, and get killed. `Job` is what I was missing :-)

martinpitt:
@stefwalter: I've spent about three hours on trying to get some service account be able to create jobs. There is a plethora of rather unhelpful and confusing documentation, about 
[service accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/), [SCCs](https://blog.openshift.com/understanding-service-accounts-sccs/) (which a non-admin user doesn't have access to, AFAICS),  [authorization and role bindings](https://docs.openshift.com/container-platform/3.6/architecture/additional_concepts/authorization.html), 
[RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/), [pod security policies](https://kubernetes.io/docs/concepts/policy/pod-security-policy/), but I keep hitting walls:

```
$ oc auth can-i create jobs
no - User "system:serviceaccount:cockpit:deployer" cannot create jobs.batch in project "cockpit"

$ oc create role psp:createjob --verb create --resource=jobs
Error from server (Forbidden): User "stefw" cannot create roles.rbac.authorization.k8s.io in project "cockpit"

$ oc create role psp:createjob --verb create --resource=jobs
Error from server (Forbidden): User "stefw" cannot create roles.rbac.authorization.k8s.io in project "cockpit"

$ oc policy who-can create jobs
Namespace: cockpit
Verb:      create
Resource:  jobs.batch

Users:  bstinson
        stefw
        system:admin
        system:serviceaccount:kube-system:cronjob-controller
        system:serviceaccount:openshift-infra:template-instance-controller

Groups: system:cluster-admins
        system:masters
```

The `system:serviceaccount:openshift-infra:template-instance-controller` sounds interesting, but this again is a service account which cannot be specified in a pod's `serviceAccountName:`. Do you happen to have any clue how to create or modify a service account to allow `create jobs`?

martinpitt:
I found some [more documentation](https://medium.com/@lestrrat/configuring-rbac-for-your-kubernetes-service-accounts-c348b64eb242), and tried my luck with this:
```yaml
---
apiVersion: v1
kind: List
items:
- kind: ServiceAccount
  apiVersion: v1
  metadata:
    name: create-job
    namespace: cockpit

- kind: Role
  apiVersion: rbac.authorization.k8s.io/v1beta1
  metadata:
    namespace: default
    name: job-creator
  rules:
    apiGroups: [""] # "" indicates the core API group
    resources: ["jobs"]
    verbs: ["create", "list", "delete"]

- kind: RoleBinding
  apiVersion: rbac.authorization.k8s.io/v1beta1
  metadata:
    name: job-creator-binding
    namespace: default
  subjects:
  - kind: ServiceAccount
    name: create-job
    namespace: cockpit
  roleRef:
    kind: Role
    name: job-creator
    apiGroup: rbac.authorization.k8s.io
```

but doesn't work either:
```
Error from server (Forbidden): User "stefw" cannot create roles.rbac.authorization.k8s.io in project "default"
Error from server (Forbidden): User "stefw" cannot create rolebindings.rbac.authorization.k8s.io in project "default"
```

So it seems non-admins just cannot delegate permissions to a new service account?

martinpitt:
I reset this PR back to the previous approach, as apparently at least on CentOS CI there doesn't seem to be a way to create a restricted user with just the `create jobs` permission. I don't want to put the real user's credentials into the secrets volume, as that would make things worse - as that, a pod can create an arbitrary other pod that leaks data to an arbitrary place. But I kept the approach on the shelf, in case we get access to another openshift instance at some point.

@stefwalter, does that work for you?

stefwalter:
> @stefwalter, does that work for you?

Works for me. And is no worse than the status quo.

martinpitt:
The 170 release started off nice, but then it [failed](https://fedorapeople.org/groups/cockpit/logs/release-170/backup.XIA_my/backup.Ohecwl/log) with `make[2]: write error: stdout`. Maybe we need some kind of careful-cat hack there.

martinpitt:
I adopted the [turd fix from Cockpit](https://github.com/cockpit-project/cockpit/commit/ef50d97cf4), let's see how that works out today.

martinpitt:
https://fedorapeople.org/groups/cockpit/logs/release-171/ failed at *exactly* the same location, so neither the recent fixes in Cockpit itself nor that "turd polish" thing helped.

martinpitt:
But now the run on the old infrastructure fails in exactly the same way. I tried to run the container manually without the sink, same error. The one thing that helped was to start the container interactively (`docker run -it`) and manually run the release-runner inside it.

martinpitt:
Yesterday's 171 release was done in the new CentOS CI container with these changes, although without the sink and in "foreground terminal" mode (as anything else triggers the stdout write error).

martinpitt:
Debugging notes: I'm stripping away step by step the container, cockpituous, and the sink. I'll update this comment as I progress.

I set up a local "sink" user with just ~/sink pointing to the cockpituous checkout's sink, and a `~/public_html/logs` user, and installed the container's and my ssh key into its `authorized_keys`. With that, it is enough to run this in a checked out and configured tree locally (as my normal user):

```
$ make clean; (stdbuf -oL -eL make -j4 dist-gzip 2>&1) | stdbuf -oL -eL ssh -tt sink@localhost -- python sink release-171
[...]
 make[1]: Leaving directory '/home/martin/upstream/cockpit'
make[1]: write error: stdout
```
This is essentially what `release-runner` does with a sink.

The `make clean` is necessary as otherwise the failing build builds enough for the next attempt to start building the tarballs, which then succeeds.

When replacing the ssh+sink entirely, then this *sometimes* works and *sometimes* fails with the write error:

    make clean; (stdbuf -oL -eL make -j4 dist-gzip 2>&1) | stdbuf -oL -eL cat

It's much more reliable with the ssh sink.

I then wrapped `/usr/bin/node` into the "non-blocking" assertion, which confirms that webpack (or sometimes tests) switch stdout to non-blocking, and tries to reset it:

```
#!/bin/sh
/usr/bin/node.real "$@"
r=$?
python -c 'import fcntl, os; map(lambda fd: fcntl.fcntl(fd, fcntl.F_SETFL, fcntl.fcntl(fd, fcntl.F_GETFL) &~ os.O_NONBLOCK), [0, 1, 2])'
python -c 'import fcntl, os; assert fcntl.fcntl(0, fcntl.F_GETFL) & os.O_NONBLOCK == 0, "stdin"; assert fcntl.fcntl(1, fcntl.F_GETFL) & os.O_NONBLOCK == 0, "stdout"; assert fcntl.fcntl(2, fcntl.F_GETFL) & os.O_NONBLOCK == 0, "stderr"'
exit $r
```
Wit that the assertions *still* fail, proving that setting stdout to nonblocking does not affect the parent process. So that "turd polish" fix looks like it couldn't really fix things, and as the release still failed with it it can't be sufficient. Then again with this I can't get the make build to fail with it, so it does help a bit.
I noticed that in failing cases I also get assertion errors on stdin.

martinpitt:
I adjusted the turd polish commit to take an even stronger hammer. Instead of trying to clean up after it (which apparently doesn't work) it prevents it from ever touching the parent process' stdio. This has two important consequences:

 * You can't tell apart stdout and stderr. I tried with leaving stderr as-is, but it still fails with that.
 * You don't get live output, just the whole output as a single flush after node exited.
 * This approach does not work for *all* invocations of node in Cockpit, particularly not for `make check`. `cdp-driver.js` crucially depends on having proper stdin/stdout to its caller.

These limitations don't matter much for the release container, so I think it's fine there. I tried to cherry-pick the node invocations to be wrapped in cockpit, but that's difficult to track down. So for now I'd go with this simplistic, but effective hack.

https://fedorapeople.org/groups/cockpit/logs/release-999/log is a release from my martinpitt/cockpit.git with a major-cockpit-release limited to release-{source,srpm} (these were the crucial bits).

martinpitt:
The stdio turd polish worked \o/ The [172 release](https://fedorapeople.org/groups/cockpit/logs/release-172/log) was done successfully and completely unattended on the new CentOS CI pod.

release/README.md
```diff
@@ -61,6 +61,38 @@ Install the systemd services:
...
+On any host, set up all necessary credentials as above, plus an extra
+`~/.config/github-webhook-token` with a shared secret.
+
+    release/build-secrets > /tmp/secrets.json
+
+to build a Kubernetes secret volume definition. Copy that to a machine that
+administers your OpenShift cluster, and deploy that secret volume and the other
+objects with
+
+    oc create -f /tmp/secrets.json
```
stefwalter: I would like to suggest piping the secrets rather than storing yet another copy in a file. Can we do:

```
release/build-secrets | oc create -f -
```

martinpitt:
> LGTM ... although the Webhook invoking a local script is really scary security wise.

I know, but I find that better than hardcoding the list of projects and their release script paths. I think all around this is actually more secure than the old webhook, as this one now  requires a valid cryptographic signature, while the other one didn't. That said, I really appreciate ideas how to isolate this  better.

> I think we should have the webhook touch a file, and the release-runner script loops and looks for when that file exists.

Out of interest, how would that improve isolation? Do you mean run the webhook and the release-runner script in two different pods, and thus with two different UIDs? Even with that, the webhook caller still has full control over which script to execute, so I don't think this would significantly change anything. I. e. someone who breaks the request signature or acquires the token can leak our secrets (but then again with the github token you can already do a lot anyway).

release/README.md
martinpitt: Nice icea. Done and pushed: https://github.com/martinpitt/cockpituous/blob/release-openshift/release/README.md#deploying-on-openshift

martinpitt:
I pushed again, I moved the "Troubleshooting" section up as it applies to a local host installation, not openshift.

martinpitt:
It's still on my todo list to ask CentOS CI for a service account that can create jobs. With that we could run the webhook permanently in one pod (with only the shared secret for that), and let it spawn completely isolated release containers on demand. This would be both much safer and also much more robust. But we don't have permission to create jobs right now, as above.

martinpitt:
I asked about the Job creation service account here: https://bugs.centos.org/view.php?id=15071 

Once we get that, we can split the webhook and release container again, and make this much nicer.

PR #184, opened 2018-06-06
images: Fix container to work on OpenShift, and add deployment resources

martinpitt:
I rebuilt the container with this changes, rolled it out onto both CentOS CI *and* our existing image stores on verifymachine4/5 (to demonstrate backwards compatibility). The new image store is here: https://images-cockpit.apps.ci.centos.org/

This needs some fixes to Cockpit's image-{upload,download}, in https://github.com/cockpit-project/cockpit/pull/9321 .

### DIFF
--- a/images/README.md
+++ b/images/README.md
@@ -1,0 +1,29 @@
+# Cockpit image server
+
+This is an image store for hosting VMs for Cockpit integration tests.
+
+# Deploying on a host
+
+Secrets need to be set up in the same way as for the
+[tests container](../tests/README.md). This container particularly needs the CA
+and server SSL certificates and the `htpasswd` file for authenticating users
+that are allowed to upload.
+
+    $ sudo docker pull cockpit/images
+    $ sudo atomic install cockpit/tests
+    $ sudo systemctl start cockpit-tests
+
+# Deploying on OpenShift
+
+Again, secrets need to be set up in the same way as for the OpenShift
+deployment for [tests](../tests/README.md), in particular the
+`cockpit-tests-secrets` secret volume.
+
+This needs a large persistent volume to hold the images. Create a claim for it
+with
+
+    oc create -f images/images-claim.yaml
+
+then wait until it gets provisioned, and create the remaining objects with
+
+    oc create -f images/cockpit-images.yaml
--- a/images/cockpit-images.yaml
+++ b/images/cockpit-images.yaml
@@ -1,0 +1,79 @@
+---
+apiVersion: v1
+kind: List
+items:
+- kind: Pod
+  apiVersion: v1
+  metadata:
+    name: images
+    labels:
+      infra: cockpit-images
+  spec:
+    containers:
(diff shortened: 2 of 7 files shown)


images/README.md
```diff
@@ -0,0 +1,29 @@
...
+# Deploying on a host
+
+Secrets need to be set up in the same way as for the
+[tests container](../tests/README.md). This container particularly needs the CA
+and server SSL certificates and the `htpasswd` file for authenticating users
+that are allowed to upload.
+
+$ sudo docker pull cockpit/images
+$ sudo atomic install cockpit/tests
+$ sudo systemctl start cockpit-tests
```
stefwalter: These lines need to be indented with 4 spaces to be formatted right.

images/README.md
```diff
@@ -0,0 +1,29 @@
...
+# Deploying on OpenShift
+
+Again, secrets need to be set up in the same way as for the OpenShift
+deployment for [tests](../tests/README.md), in particular the
+`cockpit-tests-secrets` secret volume.
+
+This needs a large persistent volume to hold the images. Create a claim for it
+with
+
+   oc create -f images/images-claim.yaml
```
stefwalter: This needs another space indent.

images/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+then wait until it gets provisioned, and create the remaining objects with
+
+   oc create -f images/cockpit-images.yaml
```
stefwalter: This needs another space indent.

images/cockpit-images.yaml
```diff
@@ -0,0 +1,79 @@
...
+          mountPath: /secrets
+          readOnly: true
+        - name: images
+          mountPath: /cache/images
+        - name: httpd-log
+          mountPath: /var/log/nginx
+        - name: httpd-state
+          mountPath: /var/lib/nginx
+        - name: httpd-state-tmp
+          mountPath: /var/lib/nginx/tmp
```
stefwalter: Can we have just one mount for the entire ```/var/lib/nginx``` directory? If there's a good reason to have separate that's fine ... but just thought I'd ask.

martinpitt: That would be nicer indeed, but unfortunately does not work. Just creating /var/lib/nginx/ causes nginx to fail with a permission error to mkdir the /tmp dir below it, as the parent dir (/var/lib/nginx) is not writable to the random uid. For the same reason just creating /var/lib/nginx//tmp also fails because the parent dir is not accessible to the random uid.

martinpitt:
I pushed the indentation fixes, thanks for spotting! https://github.com/martinpitt/cockpituous/tree/images-openshift/images gets rendered correctly now. I answered to the volume mount questions above.

images/cockpit-images.yaml
stefwalter: Makes sense.

PR #186, opened 2018-06-12
tests: Fix container to work with unprivileged users

martinpitt:
In some restricted OpenShift environments pods get started with a random
UID:

 - Don't chown files and use sudo, use world-writable directories
   instead.
 - Explicitly set $HOME, as the random UID does not have a passwd entry.
 - Ensure we have a passwd entry for a random UID, so that ssh works.

All of that is backwards compatible with the current container version
on privileged OpenShift.

Based on patch by Stef Walter, with some simplifications and adding the 
passwd fix.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -62,7 +62,6 @@ RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/gith
     printf '[user]\n\t\nemail = cockpituous@cockpit-project.org\n\tname = Cockpituous\n' >/home/user/.gitconfig && \
     ln -s /cache/images /build/images && \
     ln -s /cache/images /build/virt-builder && \
-    ln -s /cache/images /build/libvirt/qemu/log && \
     ln -s /cache/github /build/github && \
     ln -s /tmp /build/tmp && \
     mkdir -p /home/user/.config /home/user/.ssh /home/user/.rhel && \
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -79,9 +79,6 @@ RUN mkdir -p /usr/local/bin /home/user /build /secrets /cache/images /cache/gith
 # Prevent us from screwing around with KVM settings in the container
 RUN touch /etc/modprobe.d/kvm-amd.conf && touch /etc/modprobe.d/kvm-intel.conf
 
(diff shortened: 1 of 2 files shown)


martinpitt:
This is based on @stefwalter 's work in PR #176, but yet without the OpenShift resources. I'd like to land these separately, as we cannot yet fully deploy this due to https://bugs.centos.org/view.php?id=14915 . However, I did deploy a simple container on CentOS CI:

```yaml
---
kind: Pod
apiVersion: v1
metadata:
  name: tests
spec:
  containers:
    - name: tests
      image: cockpit/tests
      volumeMounts:
      - name: secrets
        mountPath: /secrets
        readOnly: true
      - name: images
        mountPath: /cache/images
      - name: cache
        mountPath: /cache
  volumes:
  - name: secrets
    secret:
      secretName: cockpit-tests-secrets
  - name: images
    persistentVolumeClaim:
      claimName: cockpit-images
  - name: cache
    emptyDir: {}
  nodeSelector:
    oci_kvm_hook: allowed
```

It starts up correctly and can talk to GitHub:
```
❱❱❱ oc logs tests
+ bots/tests-scan
tests-scan: No /dev/kvm access, not running tests here
+ bots/tests-scan --repo cockpit-project/starter-kit -c cockpit/centos-7 -c cockpit/fedora-27
tests-scan: No /dev/kvm access, not running tests here
[...]
```

images work:

```
❱❱❱ oc rsh tests sh -c 'ls -l $TEST_DATA/images/'
total 66472760
-rw-r--r--. 1 nobody nobody 1192349184 Jun  6 14:44 candlepin-b079be0009a06018009076f834a2806579b2335c90c125ca41162e5f0d4cc979.qcow2
[...]
```

ssh to the sink works:
```
❱❱❱ oc rsh tests ssh fedorapeople.org hostname
people02.fedoraproject.org
```
So for now it's useful for little more than image pruning, but I think it sufficiently demonstrates that the directory, volume, and permission structure work, and that setting up the passwd entry is sufficient to make ssh work.

I also built a new container with this, tested it on verifymachine4 (in "manual" mode), and rolled it out to dockerhub and our Cockpit OpenShift cluster. [This test](https://fedorapeople.org/groups/cockpit/logs/pull-9361-20180612-070808-c932daa3-verify-rhel-7/log.html) (and others) run fine on the new container.

PR #188, opened 2018-07-12
release: Don't force ssh TTY

martinpitt:
This causes eternal hangs, as an EOF on stdin is then not sufficient any
more to shut down ssh.  This can easily be demonstrated with

   echo whoami | ssh -tt somehost

This was introduced in commit 3f2c177abfff for release-debian. But this
has been gone for a  while, and there is no real justification to do
that -- nothing in our release scripts ought to need a PTY, and if there
is anything that should be wrapped in `script` or similar.

### DIFF
--- a/release/release-runner
+++ b/release/release-runner
@@ -217,5 +217,5 @@ else
         else
             printf '\n{ "message": "Release failed: %s", "irc": { "channel": "#cockpit" } }' "$RELEASE_TAG"
         fi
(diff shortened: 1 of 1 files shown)


martinpitt:
The cockpit-ostree 173 release was done in a container with this patch; the log looks alright, and the hang is gone, only the webhook is running. So this works in production.

PR #190, opened 2018-07-12
tests: Properly reset HEAD of cockpit checkout

martinpitt:
`git reset --hard origin/master` still leaves the current checkout in
"detached HEAD" state. This sometimes confuses git when testing a new
PR, it then tries to re-apply all commits of the entire history when
cockpit's tests-invoke tries to rebase a checked out PR onto current
master.

Do what we actually mean: First reset and clean up all changes from the
current checkout and then checkout the current origin/master.

### DIFF
--- a/tests/cockpit-tests
+++ b/tests/cockpit-tests
@@ -65,8 +65,9 @@ function task() {
 # Perform N tasks or waits, then restart
 for i in $(seq 1 30); do
     git -C cockpit fetch origin
-    git -C cockpit reset --hard origin/master
+    git -C cockpit reset --hard
     git -C cockpit clean -dxff
+    git -C cockpit checkout origin/master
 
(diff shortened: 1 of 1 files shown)


martinpitt:
> LGTM ... are we running this container now?

I just built it (over the meeting) and tested it on bots1. I'm uploading right now, will restart our pods as soon as this is done. Until that works we should not land this yet.

martinpitt:
New container with this fix deployed to our openshift. Let's watch them for a while, there can always be unrelated things due to updates of software in the container itself (newer browser and such).

martinpitt:
This has survived countless test runs now. I think I may have seen some rebase errors still, but far less. This isn't making things worse, so let's land this for now (we might need to refine later).

PR #191, opened 2018-07-16
tests: Do a clean clone if origin is shallow

martinpitt:
starter-kit and similar project tests run

    git fetch --depth=1 https://github.com/cockpit-project/cockpit.git 172

to do a shallow checkout of cockpit's bots/ directory for running their
test. git remembers this and does not fetch the full origin again even
after `git fetch origin`. While there are ways to un-shallow a checkout,
this is brittle and also does not happen too often (as the majority of
our tests are still from cockpit itself).

To avoid "rebase" errors for subsequent test runs, do a clean re-clone
of cockpit if there are any "grafted" commits, i. e. the oldest commit
from a previous shallow clone.

### DIFF
--- a/tests/cockpit-tests
+++ b/tests/cockpit-tests
@@ -2,6 +2,8 @@
 
 set -e
 
+GITHUB_REPO=https://github.com/cockpit-project/cockpit
+
 # If first argument starts with a path no loop
 loop=yes
 case ${1-} in
@@ -31,7 +33,7 @@ if ! whoami && [ -w /etc/passwd ]; then
 fi
 
 if [ ! -d cockpit ]; then
-    git clone https://github.com/cockpit-project/cockpit
+    git clone "$GITHUB_REPO" cockpit
 fi
 
 # If interactive arguments then just exec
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested this manually on a "broken" pod, and it restored the checkout from a clean checkout as expected. After that, it uses the normal "reset" method without another re-clone.

I rebuilt and pushed the new container to dockerhub, rolled it out, and restarted failing tests.

tests/cockpit-tests
```diff
@@ -31,7 +33,7 @@ if ! whoami && [ -w /etc/passwd ]; then
 fi
 
 if [ ! -d cockpit ]; then
-    git clone https://github.com/cockpit-project/cockpit
+    git clone "$GITHUB_REPO"
```
larskarlitski: Since the project name is now encoded in `$GITHUB_REPO`, please change this to `git clone "$GITHUB_REPO" cockpit`. (Same further down.)

martinpitt: Fixed.

martinpitt:
The rebuilt container (with the above change for pinning down the checkout dir) has run several tests now, successfully.

PR #193, opened 2018-07-16
release: Make release-source work for plain Makefile projects

martinpitt:
welder-web, starter-kit, cockpit-ostree, and similar don't use
autotools, but just a plain Makefile. Support these in release-source:
if autogen.sh or configure exist, do as before (i. e. for cockpit
itself), otherwise try `make dist` or `make dist-gzip`.

Disable patch generation for non-autotools projects, as there is no
predictable way to do out-of-tree builds for them.

### DIFF
--- a/release/release-source
+++ b/release/release-source
@@ -142,14 +142,32 @@ prepare()
     # Find archives in the source directory
     archive="$(find_and_extract_tarballs $TAG $SOURCE $repodir)"
 
+    if [ -x ./autogen.sh ] || [ -x ./configure ]; then
+        is_autotools=1
+    else
+        is_autotools=
+    fi
+
     # If it is missing rebuild the tarball from latest tag and copy into place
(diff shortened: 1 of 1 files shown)


martinpitt:
I successsfully tested this with cockpit, starter-kit, cockpit-ostree, and welder-web. From a clean git checkout I ran

    RELEASE_SOURCE=$PWD/_release/source ~/upstream/cockpituous/release/release-source

and got a valid release tarball created in _release/source. During this I found a bug in starter-kit's build system ([fix](https://github.com/cockpit-project/starter-kit/commit/3109cad8ca34fa)), which I will also apply to ostree and podman.

martinpitt:
This replaces and generalizes welder-web's [cockpituous release script](https://github.com/weldr/welder-web/blob/master/utils/cockpituous-release) hack with a local `release_source()` function.

PR #194, opened 2018-07-17
release: Support spec files with pre-filled Version:

martinpitt:
starter-kit based projects do release tarballs with spec files that
already have a correct `Version:` instead of `0`. This makes them more
self-contained and able to build working (s)rpms without the help of
cockpituous.

Support this mode with a new `release-srpm -V` option. By default,
release-srpm would generate a spec file with `Release: 2` as it assumes
that `XXX-1` was already released. `-V` tells it that the `Version:` is
already correct and it should leave the `Release:` at 1.

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -72,13 +72,13 @@ this:
 
 ## Spec file requirements
 
-For a spec file to work with the scripts, it should be setup like this:
+For a spec file to work with `release-srpm`, it should have `Version: 0`.
+If your build system already puts the target release version into `Version:`,
+call `release-srpm` with the `-V` option instead, otherwise it will assume that
+this version was already released in a previous srpm and bump the `Release:`.
 
-    Version: 0
-
-And not have any content after the following line:
-
-    %changelog
+The spec file must not have any content after `%changelog`. It will generate a
+changelog from the git tag description.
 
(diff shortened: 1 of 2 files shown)


martinpitt:
Setting blocked as usual, until we do a successful Cockpit release with this.

martinpitt:
Tested with starter-kit with
```
git clean -ffdx
RELEASE_SOURCE=$PWD/_release/source  ~/upstream/cockpituous/release/release-source
RELEASE_SOURCE=$PWD/_release/source  RELEASE_SPEC=cockpit-starter-kit.spec RELEASE_SRPM=_release/srpm ~/upstream/cockpituous/release/release-srpm -V
```
This also needs a few fixes in starter-kit and related projects itself, which I'll post soon.

martinpitt:
[cockpit-ostree 173](https://github.com/cockpit-project/cockpit-ostree/releases/tag/173) was successfully released with this. But for good measure, let's still wait until after the next Cockpit release, to test the case with `Version: 0` again as well.

martinpitt:
[Cockpit 173 release](https://fedorapeople.org/groups/cockpit/logs/release-173/log) completed successfully with this.

PR #198, opened 2018-08-21
release: Force autoreconf in patch creation

martinpitt:
If ./autogen.sh is available, prefer that over calling ./configure when
building patches. Otherwise changes to autotools input files
(configure.ac) will not end up in generated patches.

### DIFF
--- a/release/release-source
+++ b/release/release-source
@@ -68,10 +68,10 @@ message()
 autogen_or_configure()
 {
     mkdir -p $1/_build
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested this when generating patches from Cockpit's current rhel-8.0 branch for https://bugzilla.redhat.com/show_bug.cgi?id=1615508 . Before it didn't pick up the libssh configure.ac patch properly, now it does.

PR #200, opened 2018-08-27
release: Adjust to moved Cockpit release script

martinpitt:
Cockpit's cockpituous release script was moved into tools [1], so adjust
the documentation accordingly.

[1] https://github.com/cockpit-project/cockpit/pull/9914

 - [x] Land https://github.com/cockpit-project/cockpit/pull/9914


martinpitt:
This does not affect our container, as the path to the release script is set in the web hook. But we should land https://github.com/cockpit-project/cockpit/pull/9914 first, thus setting to blocked.

PR #201, opened 2018-09-07
release: Enable automatic landing in bodhi

martinpitt:
In the past two years there has never been a case where we had to recall
a Cockpit update, and pushing it to stable is an extra manual step that
one must remember. Let this happen automatically now.

Note that a bad update can still be blocked by a single -1.


martinpitt:
This corresponds to the "Autopush" flag that one sees on bodhi update pages. See https://fedoraproject.org/wiki/Bodhi#Karma_Automatism

Issue #196, opened 2018-08-03
image pod needs to auto-respawn


martinpitt:
We currently [create](https://github.com/cockpit-project/cockpituous/blob/master/images/cockpit-images.yaml) only one image pod to serve https://images-cockpit.apps.ci.centos.org/ . The pod just failed as the underlying node went away. This needs to use a `ReplicationController` to automatically provide a replacement in this situation.

PR #202, opened 2018-09-07
tests: Add pod definition for centos ci

Gundersanne:
Use yaml over json + added centos ci pod definition.


tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,83 @@
...
+      metadata:
+        labels:
+          infra: cockpit-tasks
+      spec:
+        containers:
+        - name: cockpit-tasks
+          image: docker.io/cockpit/tests
+          securityContext:
+            runAsUser: 1111
+            hostNetwork: false
```
martinpitt: Please drop this on CentOS CI, it doesn't work there.

In fact, I think we should also work towards fixing the UID on our BOS cloud. The necessary pieces should all be there now, and there's also PR #176. But this is clearly a follow-up.

tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,83 @@
...
+    replicas: 1
+    selector:
+      infra: cockpit-tasks
+    template:
+      metadata:
+        labels:
+          infra: cockpit-tasks
+      spec:
+        containers:
+        - name: cockpit-tasks
```
martinpitt: Can we please name this "centosci-tasks", and rename the -bos one "bos-tasks"? This makes it possible to see which cloud they were running on, and where to log in for "oc rsh". I'm not entirely sure that it's *this* `name:` which determines that, it could also be the `ReplicationController` name (but I think that's unlikely).

tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,83 @@
...
+            value: '8'
+          - name: TEST_PUBLISH
+            value: sink
+          volumeMounts:
+          - name: secrets
+            mountPath: "/secrets"
+            readOnly: true
+          - name: cache
+            mountPath: "/cache"
+            readOnly: false
```
martinpitt: `readOnly: false` is the default, let's drop it.

tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,83 @@
...
+          - name: images
```
martinpitt: Wrongly indented, this starts a new  "container:" definition. Please let's not do this here. The /cache/images volume is already shared between all pods, there is no need to deploy umpteen nginx services that all serve from the same directory. We already have an image server in images/cockpit-images.yaml. (This doesn't auto-respawn, but that's issue #196 - let's fix that separately).

So please drop this container.

tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,83 @@
...
+          - name: cache
+            mountPath: "/cache"
+            readOnly: false
+        volumes:
+        - name: secrets
+          secret:
+            secretName: cockpit-tests-secrets
+        - name: cache
+          hostPath:
+            path: "/var/cache/cockpit-tests"
```
martinpitt: This  only applies to the BOS cloud. Here we need:
```
  - name: cache
    emptyDir: {}
  - name: images
    persistentVolumeClaim:
      claimName: cockpit-images
```

tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,83 @@
...
+        - name: shm
+          emptyDir:
+            medium: Memory
+        - name: tmp
+          emptyDir: {}
+        serviceAccountName: cockpituous
+- kind: Service
```
martinpitt: Like above, let's not do Images here, that's a separate service in the images/ directory.

tests/cockpit-tasks-centosci.yaml
Gundersanne: Ok, Should I drop it on bos too?

martinpitt: No, let's keep that for now on BOS, this works differently as it uses host paths, not a shared volume. We might make this more efficient in the future, but I'd rather not touch it without thinking hard about it :)

tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,51 @@
+---
+apiVersion: v1
+kind: List
+items:
+- kind: ReplicationController
```
martinpitt: For my similar PR #206 it complained that I need to set `apiVersion: v1` here as well. Apparently this wasn't necessary at the time when we rolled this out on our bos cluster, but CentOS CI's version needs it.

That's the kind of fine-tuning that you can do when actually rolling this out. Try without, and if it complains about something like

    error: unable to get type info from the object "*unstructured.Unstructured": Object 'apiVersion' is missing in 'object has no apiVersion field'

then add it here.

tests/cockpit-tasks-centosci.yaml
```diff
@@ -0,0 +1,51 @@
...
+    template:
+      metadata:
+        labels:
+          infra: cockpit-tasks
+      spec:
+        containers:
+        - name: centosci-tasks
+          image: docker.io/cockpit/tests
+          securityContext:
+            hostNetwork: false
```
martinpitt: I don't think we need this on CentOS CI either - our ad-hoc pods seemed to work well without it, so I'm in favor of completely dropping the `securityContext` section.

tests/cockpit-tasks-centosci.yaml
Gundersanne: Hm centos ci does need the api version, it won't create one otherwise.

PR #203, opened 2018-09-13
release: Add option to release from a non-master branch

martinpitt:
This is useful for cutting releases from stable branches, which just
receive some cherry-picks. This is more practical and robust for
downstream packaging than having to carry huge patches against dist/.

It's also very difficult to carefully tweak all file time stamps to not
trigger webpack when patching - this approach avoids this entirely.

### DIFF
--- a/release/release-runner
+++ b/release/release-runner
@@ -23,6 +23,7 @@
 # -v         RELEASE_VERBOSE=1      Make output more verbose
 # -z         RELEASE_CHECK=1        Check access and configuration
 # -l         RELEASE_SINK=1         Host to sink logs to
+# -b         RELEASE_BRANCH=stable  Branch to release (default: master)
 # -t         RELEASE_TAG=0.X        Tag to release
 #
 
@@ -34,6 +35,7 @@ set -m
 
 SINK=${RELEASE_SINK:-}
 RELEASE_TAG=${RELEASE_TAG:-}
+RELEASE_BRANCH=${RELEASE_BRANCH:-master}
 RELEASE_VERBOSE=0
 RELEASE_TRANSACTION=1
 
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested this on my cockpit-ostree fork. The latest tag there is 176, and I created 173.1 with a modified cockpituous release script. I then invoked the release like this:

    ~/upstream/cockpituous/release/release-runner -r https://github.com/martinpitt/cockpit-ostree.git -b stable -v ./cockpituous-release

This succeeded, there is a [GitHub release](https://github.com/martinpitt/cockpit-ostree/releases) and [successful COPR build](https://copr.fedorainfracloud.org/coprs/martinpitt/welder-web/build/797681/).

I was a bit afraid that the "XXX.Y" tag format would break something, but it seems it's going alright.

Note that this won't play well with our webhook yet - this does not yet know about the new `--branch` option. But this at least gives us a means to manually trigger a stable point release, which is already very useful.

martinpitt:
FTR: When pushing a tag to a non-master branch, this will currently trigger our webhook, which will try to do a release on the master branch. This is a bit ugly, but it behaves sensible and fails fast:
```
fatal: No tags can describe 'c76b49c840a7ac8ecd87d80bfc10643e8fc5043b'.
Try --always, or create some tags.
release-runner: No release tag set
```

martinpitt:
I now pushed a 177 release tag to my master branch and ran with the defaults:

    ~/upstream/cockpituous/release/release-runner -r https://github.com/martinpitt/cockpit-ostree.git -v ./cockpituous-release

That again succeeded to do a [GitHub release](https://github.com/martinpitt/cockpit-ostree/releases) and a [COPR build](https://copr.fedorainfracloud.org/coprs/martinpitt/welder-web/build/797685/), so I think this proves that I didn't screw up the "master" default.

PR #206, opened 2018-09-14
images: Use ReplicationController

### LINKED ISSUE #196: image pod needs to auto-respawn

martinpitt:
Pods can fail at any time, e. g. our solitary "images" pod got killed
the other day when the underlying node went away. So wrap it in a
replication controller with one replica (which is sufficient for our
purposes).

Fixes #196

### DIFF
--- a/images/cockpit-images.yaml
+++ b/images/cockpit-images.yaml
@@ -2,50 +2,56 @@
 apiVersion: v1
 kind: List
 items:
-- kind: Pod
+- kind: ReplicationController
   apiVersion: v1
   metadata:
     name: images
-    labels:
-      infra: cockpit-images
   spec:
-    containers:
-      - name: images
-        image: cockpit/images
-        ports:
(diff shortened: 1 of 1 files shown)


martinpitt:
I removed our old static images pod, and rolled this out on CentOS CI. The usual [image server URL](https://images-cockpit.apps.ci.centos.org/) continues to work as usual, and it created one pod:

    images-zxp3r   1/1       Running   0          55s

I then killed this (`oc delete pod images-zxp3r`), and as desired a new one was immediately created:

   images-902nm   1/1       Running   0          1m

So this works as intended.

PR #209, opened 2018-09-30
tests: Make sure we run on Centos CI nodes that expose /dev/kvm

martinpitt:
This was lost in translation between my original PoC and bouncing it between Sanne and me.

Right now it seems to work anyway, as apparently oci-kvm-hook got deployed everywhere. But let's be explicit about it anyway.


martinpitt:
I haven't yet deployed this and thus syntax-checked it. Just quick-filing this as a reminder.

martinpitt:
Deployed and verified that containers have /dev/kvm. @Gundersanne , can you please review/land?

PR #210, opened 2018-10-02
Updates to tests container

martinpitt:
See individual commits.


martinpitt:
I rebuilt the base and tests containers, pushed them to dockerhub, and rolled them out onto our bos and centosci clouds.

martinpitt:
I also rebuilt/repushed the release and images containers, to make sure they also work with Fedora 28. https://images-cockpit.apps.ci.centos.org/ proves that for images. The release container will test itself on the next release.

PR #213, opened 2018-10-11
learn: Add simple limit to amount of training data

stefwalter:
The current DBSCAN model tries to do a full matrix of all
distances in the training data. Lets just limit to the most
recent N data items. Otherwise the size of the model
and runtime becomes unwieldly.

Hopefully we can eventually migrate to a more efficient
clustering algorithm or implementation.

 * [x] https://github.com/cockpit-project/cockpit/pull/10276

### DIFF
--- a/learn/cluster.py
+++ b/learn/cluster.py
@@ -184,7 +184,7 @@ class Model():
         self.features = None
 
     # Perform the unsupervised clustering
-    def train(self, items):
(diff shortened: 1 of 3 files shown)


stefwalter:
This is deployed in both clusters and being tested by https://github.com/cockpit-project/cockpit/pull/10276

PR #214, opened 2018-10-12
learn: Add a expose Route for testing

stefwalter:
This is to be used temporarily to interact with the learn
service. However it's not meant for long term usage since
there's no authentication on the service.

### DIFF
--- a/learn/cockpit-expose.yaml
+++ b/learn/cockpit-expose.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: v1
+kind: List
+items:
+- kind: Route
+  apiVersion: v1
+  metadata:
+    name: cockpit-expose
+  spec:
+    to:
+      kind: Service
+      name: cockpit-learn
+    port:
+      targetPort: 8080
--- a/learn/README.md
+++ b/learn/README.md
@@ -60,6 +60,17 @@ with curl should work:
 
     $ kubectl run -it test --image=docker.io/cockpit/learn --restart=Never -- sh
 
+Alternatively you can expose the service temporarily. By doing the following:
+
+    $ kubectl create -f learn/cockpit-expose.yaml
+    $ kubectl describe route cockpit-expose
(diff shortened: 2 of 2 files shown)


learn/cockpit-expose.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: v1
+kind: List
+items:
+- kind: Route
+  apiVersion: v1
+  metadata:
+    name: cockpit-expose
```
martinpitt: This ends up as `cockpit-expose-cockpit.apps.ci.centos.org`, which is both redundant and also non-descriptive. Could this be "learn" instead?

stefwalter: I'd like to somehow penalise and remind people not to leave this "open" ... this is purely a diagnostic tool, because there is no authentication on cockpit-learn. I thought that by giving it a general name, and having multilpe such things compete for it, we may be able to accomplish that.

Do you know of another way to accomplish this? 

Or should I work to make the only the RO parts of cockpit-learn accessible via this Route?

martinpitt: The route still exists on CentOS CI. I'm not entirely sure why it's necessary (we don't seem to use it in Cockpit's bots), but this approach sounds like "we leave open an unsafe port and hope that someone stumbles upon it because of the weird name"?

AFAICS there are just two scenarios:

 - This is just a debugging tool, but allows write access;  this should then be done via `oc rsh` and never be exposed. It's fine to add the route to README or so, but not into the officially deployed YAML.
 - This is a long-term useful read-only debug service which remote services want to talk to; then it should get a proper name.

stefwalter: Yes, sorry. everything that has been going on has caused me to neglect this.

I've done work to transfer the data gathering tasks into the learn pod. Which will allow us to leave this open in read-only mode.

stefwalter:
Will be done correctly in #232

PR #216, opened 2018-10-12
release: Use Jobs to spawn release-runner in separate containers

martinpitt:
See https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

This provides several benefits:

 - Allow multiple releases to happen in parallel. Right now, pushing a
   release tag to one project while another release is running (or 
   hanging), will be silently ignored.
 - Simplify/robustify the structure of our webhook, as a release request
   does not need to be handled synchronously.
 - Separate the credentials for the webhook and all the release secrets,
   providing better isolation in case the web server gets compromised.

In cockpit-release.yaml, drop the release secrets and use the new 
"create-job" service account.

Copy webhook's home directory initialization logic to release-runner, so
that this can run in a freshly spawned cockpit/release container with
/run/secrets/release and a random UID.

Simplify webhooks's setup() to drop the passswd setup (now unnecessary,
as this doesn't do any actual work aside from `oc create`), and only set 
up the temporary $HOME with the webhook secret. In theory, the latter
could also be done statically in Dockerfile with a symlink; this is a
simplification that can be done in a follow-up PR. 

In webhook, replace the direct invocation of release-runner with a job 
that creates a new cockpit/release pod. This is fast, so it can happen
synchronously in the HTTP handler.

### DIFF
--- a/release/release-runner
+++ b/release/release-runner
@@ -217,5 +217,5 @@ else
         else
             printf '\n{ "message": "Release failed: %s", "irc": { "channel": "#cockpit" } }' "$RELEASE_TAG"
         fi
-    ) | stdbuf -oL -eL ssh "$SINK" -- python sink release-$RELEASE_TAG
+    ) | stdbuf -oL -eL ssh "$SINK" -- python sink "release-$(basename $REPO .git)-$RELEASE_TAG"
 fi
--- a/release/release-runner
+++ b/release/release-runner
@@ -60,6 +60,39 @@ message()
     echo "release-runner: $@" >&2
 }
 
+# check if we have a secrets volume mounted and the /home/user directory
+# from the cockpit/release container; if so, initialize HOME=/home/user
+init_container_home()
+{
+    local secrets_dir=/run/secrets/release
+    local home_dir='/home/user'
+
+    [ -d "$secrets_dir" ] && [ -d "$home_dir" ] || return 0
+    export HOME=$home_dir
+
+    trace "Initializing $HOME"
+
+    # ensure we have a passwd entry for random UIDs
+    # https://docs.openshift.com/container-platform/3.7/creating_images/guidelines.html
+    if ! whoami &> /dev/null && [ -w /etc/passwd ]; then
+       echo "randuser:x:$(id -u):0:random uid:${HOME}:/sbin/nologin" >> /etc/passwd
+    fi
+
+     # install credentials from secrets volume; copy to avoid world-readable files
+     # (which e. g. ssh complains about), and to make them owned by our random UID.
+     local old_umask=$(umask -p)
+     umask 077
+     # ignore files starting with .., these are secrets volume internal files
+     find "$secrets_dir" -mindepth 1 ! -path '*/..*' | while read f; do
+        name=$(basename $f)
+        # secrets volume files encode / as --, replace them back
+        dest="$HOME/${name//--//}"
+        mkdir -p "$(dirname "$dest")"
+        cp "$f" "$dest"
+     done
+     $old_umask
+}
+
 export_all()
 {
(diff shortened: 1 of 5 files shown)


martinpitt:
I rebuilt the container, [rolled it out to dockerhub](https://hub.docker.com/r/cockpit/release/tags/), and deployed this on CentOS CI. I tried it once again with my [cockpit-ostree fork](https://github.com/martinpitt/cockpit-ostree), and I got a successful release: [GitHub](https://github.com/martinpitt/cockpit-ostree/releases), [COPR](https://copr.fedorainfracloud.org/coprs/martinpitt/welder-web/package/cockpit-ostree/), [log](https://fedorapeople.org/groups/cockpit/logs/release-177/log).

Logs from webhook container:
```
$ oc logs release-jtlkm  -f
DEBUG:root:event: create, path: /cockpituous-release
DEBUG:root:{"ref":"177","ref_type":"tag","master_branch":"master", [...]
INFO:root:Releasing project https://github.com/martinpitt/cockpit-ostree.git, tag 177, script ./cockpituous-release
job "release-job" created
```

Logs from spawned Job pod:
```
$ oc logs -f release-job-5plk0
> Initializing /home/user
Cloning into '.'...
HEAD is now at 238af8d disable fedora
http://fedorapeople.org/groups/cockpit/logs/release-177/
> Initializing /home/user
> Starting: release-source
[...]
  09:04:36 Build 808612: running
  09:09:09 Build 808612: succeeded
> Completed job
```

Unfortunately the job object stays around after success:

```
$ oc get jobs
NAME          DESIRED   SUCCESSFUL   AGE
release-job   1         1            17m

$ oc get pods | grep job
release-job-5plk0      0/1       Completed   0          16m
```

The `ttlSecondsAfterFinished: 0` option in the job ([docs](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/)) is supposed to clean this up, but it's apparently not yet implemented in the relatively old OpenShift 3.6 setup on CentOS CI. So for now these completed job containers stay around, and we have to clean them up from time to time. But I don't see this as a big issue, they don't take up significant resources. As a workaround, a follow-up PR could regularly remove these. I filed issue #218 about this.

martinpitt:
@stefwalter: Are you interested in reviewing this? (This might also be beneficial for the learn container; happy to help and explain things) If not, I'll ask someone else.

martinpitt:
Cockpit 180 release also went through without a hitch: https://fedorapeople.org/groups/cockpit/logs/release-180/log

martinpitt:
I fixed the ambiguous job object name, and also fixed the release log name on the sink while I was at it. Everything re-rolled out again. I updated the above comment for the new URLs, cockpit-ostree 177 release was successful.

release/release-runner
```diff
@@ -60,6 +60,39 @@ message()
...
 }
 
+# check if we don't have a proper home directory, but a secrets volume mounted;
+# if so, initialize HOME=/home/user
+init_container_home()
+{
+    local secrets_dir=/run/secrets/release
+    local home_dir='/home/user'
+
+    [ -d "$secrets_dir" ] && [ -d "$home_dir" ] || return 0
```
Gundersanne: This isn't 100% clear to me, doesn't this clash with the comment above? This will only continue if both the `secrets_dir` and `home_dir` exist.

Or is `home_dir` assumed to be always present but it needs to be initialized from the secrets dir?

martinpitt: Indeed I went back and forth with this a bit, so the comment and the code now don't coincide. /home/user is created in Dockerfile, but it's empty. I'll update the comment.

martinpitt: Pushed.

PR #212, opened 2018-10-11
learn: Well known path to latest log of training

stefwalter:
This adds a well known HTTP REST path to the log of the latest training.

 * [x] https://github.com/cockpit-project/cockpit/pull/10276
 * [x] #213

PR #219, opened 2018-10-12
learn: Gather statistics for the tests

stefwalter:
In addition make the cluster json output have the same format as the statistics.

 * [x] https://github.com/cockpit-project/cockpit/pull/10276
 * [x] #212 
 * [x] #213 
 * [x] #215

### DIFF
--- a/learn/analyze.py
+++ b/learn/analyze.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2017 Stef Walter
+#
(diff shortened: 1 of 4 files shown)


learn/analyze.py
```diff
@@ -0,0 +1,89 @@
...
+
+class Group(dict):
+    total = 0
+
+    def count(self, field, value, factor=1):
+        def add(value):
+            values = self.get(field)
+            if values is None:
+                self[field] = values = { }
+            count = values.get(value, 0)
```
martinpitt: A better pattern for this is `self.setdefault(field, {}).get(value, 0)`

PR #220, opened 2018-10-18
sink: Add script to clean up old logs

martinpitt:
Our sinks accumulate an impressive amount of log data (roughly 70 GB in
the last month on fedorapeople.org). These are prone to hit the quota
limit or ENOSPC.

Add a script which cleans up obsolete ones. This takes into account that
we don't yet do machine learning on non-Cockpit tests, so keep all of
these external project logs around. These don't have any artifacts like
core dumps etc. right now, so they don't take up much space anyway.

### DIFF
--- a/sink/clean-logs.sh
+++ b/sink/clean-logs.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+# clean up obsolete logs on our sink
+
+set -ec
+cd /srv/groups/cockpit/logs
+
+# remove all logs older than two months
+# keep logs for external projects for now, until we get them into machine learning
+# ironically, the external projects have "-cockpit-" in the name (from a test
+# like cockpit/rhel-7-6/chrome@weldr/welder-web), while cockpit's own tests are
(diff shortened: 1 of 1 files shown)


stefwalter:
How does this run?

martinpitt:
It's not cron'ed or anything right now, I just ran it once yesterday to clean up. I mostly just want a place where it is accessible and maintainable.

stefwalter:
This seems cockpit specific. Others are using the sink and release code, and we should be trending away from including cockpit specific stuff here. Any way to make it less cockpit specific, or include it elsewhere?

If no other better option, then I'm fine with merging it here.

martinpitt:
It's also fedorapeople.org specific due to the path. I'm also happy with just adding it to fedorapeople.org/~cockpit, I'm not that fussed about adding it to cockpituous. Just "somewhere else than just my laptop" :)

I added it here: https://fedorapeople.org/groups/cockpit/

PR #223, opened 2018-11-09
tests: Copy bots from their own checkout

mvollmer:
So that they are protected from whatever happens inside the cockpit/
checkout.  For example, the cockpit/ checkout might run tests on a
branch with a version of the bots that are too old.


mvollmer:
I haven't tested this at all yet.

mvollmer:
Tested (outside a container) and it seems to work.

PR #224, opened 2018-11-15
Add weldr/welder-web

mvollmer:


martinpitt:
Deployed. Let's give them a few spins and then land this.

PR #225, opened 2018-11-22
tasks: Rename tests to tasks

Gundersanne:
- [x] rename the directory on bos cluster and add a symlink for old name
 - [x] move the images container over to cockpit-tasks-secrets and re-deploy it


Makefile
```diff
@@ -9,8 +9,8 @@ docker-running:
 	systemctl start docker
 
 TAG := $(shell date --iso-8601)
-TEST_SECRETS := /var/lib/cockpit-tests/secrets
-TEST_CACHE := /var/cache/cockpit-tests
+TEST_SECRETS := /var/lib/cockpit-tasks/secrets
+TEST_CACHE := /var/cache/cockpit-tasks
```
martinpitt: Let's rename the variables to `TASKS_*`, too. When deploying this container, we'll need to adjust this path on our bos cluster. We can do this now and add a tests → tasks symlink. (Creating a TODO item for that)

README.md
```diff
@@ -6,5 +6,5 @@ The Cockpit Continuous Delivery Tools
  * learn: Machine learning for test results
  * release: container to run releases
  * sink: log sink used by testing infrastructure
- * tests: container used to select and run integration testing
+ * tasks: container used to select and run integration testing
```
martinpitt: This does way more these days; how about "run cockpit bots tasks like integration tests or image rebuilds"

martinpitt:
The images/ container also uses the same tests/tasks volume and dir, so the renaming (and re-building/deploying) needs to happen there as well.

martinpitt:
I renamed the directories and symlinked them on all our bos machines.

martinpitt:
I built and deployed cockpit-tasks-secrets to both our clouds. Once you switched over and everything works, the old cockpit-tests-secrets should be removed.

As next steps, I suggest:

 - Uploading cockpit/tasks to dockerhub
 - Setting the replicationcontrollers to 0 on both clouds
 - deploying the new tasks yaml, and starting the pods
 - watch them for a while
   * if everything is good, remove the cockpit-tests related resources on both clouds
   * if something breaks, dial the tasks RCs to 0 and bring the tests RCs back up

martinpitt:
Note that the images container also needs to be updated, see above. Therefore the "requested changes" remains for now.

images/.#cockpit-images.yaml
```diff
@@ -0,0 +1 @@
+sanne@worky.17978:1542204001
```
martinpitt: We don't want that bit :)

Gundersanne: woops, guess I didn't setup my info/exclude on this one :)

Gundersanne:
Rolled out on bos now too; only cockpit-11 needs a refresh of cockpit-images

images/install-service
```diff
@@ -15,12 +15,12 @@ Requires=docker.service
...
 [Service]
-Environment="TEST_CACHE=$CACHE"
-Environment="TEST_SECRETS=$SECRETS"
+Environment="TASK_CACHE=$CACHE"
+Environment="TASK_SECRETS=$SECRETS"
 Restart=always
 RestartSec=60
 ExecStartPre=-/usr/bin/docker rm -f cockpit-images
-ExecStart=/bin/sh -xc "/usr/bin/docker run --name=cockpit-images --publish=8090:8080 --publish=8493:8443 --volume=\$TEST_SECRETS:/secrets:ro --volume=\$TEST_CACHE:/cache:rw cockpit/images"
+ExecStart=/bin/sh -xc "/usr/bin/docker run --name=cockpit-images --publish=8090:8080 --publish=8493:8443 --volume=\$TASK_SECRETS:/secrets:ro --volume=\$TASK_CACHE:/cache:rw cockpit/images"
```
martinpitt: The source ports are wrong, but I'll send a separate PR about that.

martinpitt:
I removed the old cockpit-tests-secrets everywhere.

PR #226, opened 2018-11-26
tasks: Move webhook from release to tasks

Gundersanne:
- [x] rebuild & repush tasks container


tasks/Dockerfile
```diff
@@ -60,7 +61,9 @@ RUN dnf -y update && \
     dnf -y builddep /tmp/cockpit.spec && \
     dnf clean all
 
-COPY cockpit-tasks install-service /usr/local/bin/
+RUN pip install pika
```
martinpitt: What's wrong with installing the `python3-pika` package?

tasks/cockpit-tasks-webhook.yaml
```diff
@@ -5,20 +5,20 @@ items:
 - kind: ReplicationController
   apiVersion: v1
   metadata:
-    name: release
+    name: tasks-webhook
```
martinpitt: How about just naming that `webhook`? It should be an implementation detail of the RC, service, and pod that this is running in the tasks container.

tasks/cockpit-tasks-webhook.yaml
```diff
@@ -5,20 +5,20 @@ items:
...
   spec:
     replicas: 1
     selector:
-      infra: cockpit-release
+      infra: cockpit-tasks
```
martinpitt: And this would be particularly confusing, as we have a separate RC for the actual task execution. So let's just name the yaml file, and all labels like "webhook"?

tasks/README.md
```diff
@@ -137,3 +139,26 @@ We can scale the number of testing machines in the openshift cluster with this
 command:
 
     $ oc scale rc cockpit-tasks --replicas=3
+
+## Automatic releases through a GitHub webhook
```
martinpitt: I know it doesn't do that yet, but it will soon - so maybe let's already call that "GitHub webhook triggered actions"?

tasks/README.md
```diff
@@ -137,3 +139,26 @@ We can scale the number of testing machines in the openshift cluster with this
...
+
+The intention is that the webhook pod automatically starts a release runner
+whenever a new release tag gets pushed to a project. This can be done with a
+[GitHub webhook](https://developer.github.com/webhooks/).
+
+Add a webhook to your GitHub project on the Settings → Webhooks page of your project:
+
+ * Set a Payload URL like
+
+       http://release-cockpit.apps.ci.centos.org/tools/cockpituous-release
```
martinpitt: The host name will need some adjustment, preferrably `webhook-cockpit.apps...`.

tasks/Dockerfile
Gundersanne: Oh, no idea it existed, will change that of course.

tasks/cockpit-tasks-webhook.yaml
Gundersanne: true

Gundersanne:
@pitti I undeployed the release bot containing the webhook, and deployed a cockpit/tasks with the webhook in it. Haven't updated the release container yet though.

PR #228, opened 2018-11-28
learn: Render a cluster graph for all points

stefwalter:
This allows us to visually see what's going on with the different clusters


stefwalter:
Example graph:

clusters

PR #230, opened 2018-11-30
tasks: Handle pull request events in webhook

Gundersanne:
Add AMQP aware tests-scan:

 - [x] https://github.com/cockpit-project/cockpituous/pull/241
 - [x] try out new webhook pod

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ all:
 	@echo "       make release-install" >&2
 
 TAG := $(shell date --iso-8601)
-TASK_SECRETS := /var/lib/cockpit-tasks/secrets
+TASK_SECRETS := /var/lib/cockpit-secrets/tasks
+WEBHOOK_SECRETS := /var/lib/cockpit-secrets/webhook
 TASK_CACHE := /var/cache/cockpit-tasks
 DOCKER := $(shell which podman docker 2>/dev/null)
 
@@ -73,6 +74,7 @@ tasks-shell:
 		--privileged --uts=host \
 		--volume=$(CURDIR)/tasks:/usr/local/bin \
 		--volume=$(TASK_SECRETS):/secrets:ro \
+		--volume=$(WEBHOOK_SECRETS):/run/webhook/secrets:ro \
 		--volume=$(TASK_CACHE):/cache:rw \
 		--entrypoint=/bin/bash \
         docker.io/cockpit/tasks -i
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -47,7 +47,6 @@ RUN dnf -y update && \
         python-lxml \
         python-scikit-learn \
         python3 \
-        python3-pika \
         python3-scikit-learn \
         rpm-build \
         rpmdevtools \
@@ -61,7 +60,8 @@ RUN dnf -y update && \
     curl -s -o /tmp/cockpit.spec https://raw.githubusercontent.com/cockpit-project/cockpit/master/tools/cockpit.spec && \
     sed -i 's/%{npm-version:.*}/0/' /tmp/cockpit.spec && \
     dnf -y builddep /tmp/cockpit.spec && \
-    dnf clean all
+    dnf clean all && \
+    pip3 install pika # python3-pika RPM doesn't support SNI extension for TLS, which is needed for SSL auth with RabbitMQ on OpenShift
 
 COPY cockpit-tasks install-service webhook github_handler.py /usr/local/bin/
 
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -63,6 +63,11 @@ And now you can start the service:
     $ sudo systemctl start cockpit-tasks
     $ sudo systemctl enable cockpit-tasks
 
+To generate the certificates needed for cross-cluster amqp auth follow this
+guide:
+
+    https://www.rabbitmq.com/ssl.html
+
 ## Troubleshooting
 
 Some helpful commands:
@@ -142,10 +147,6 @@ command:
 
 ## Let GitHub webhook trigger actions
 
-The intention is that the webhook pod automatically starts a release runner
-whenever a new release tag gets pushed to a project. This can be done with a
-[GitHub webhook](https://developer.github.com/webhooks/).
-
 Add a webhook to your GitHub project on the Settings → Webhooks page of your project:
 
  * Set a Payload URL like
@@ -161,4 +162,24 @@ Add a webhook to your GitHub project on the Settings → Webhooks page of your p
  * Change the Content Type to `application/json`.
 
  * Select "Let me select individual events" and let the hook run on "Branch or
-   tag creation", and nothing else.
+   tag creation", "Pull requests", and "Statuses".
+
+See [GitHub development documentation](https://developer.github.com/webhooks/)
+for more information.
+
+### Automated releases
+
+Whenever a new release tag gets pushed to a project a "create" event is received
+by the webhook which automatically starts a release runner.
+
+### Automated testing
+
+When a pull request event is received, the webhook will scan the pull request
+for tests that need to be run. These will be put in a distributed queue to later
+be consumed by other task bots. If "[no-test]" is in the title of the pull
+request, it won't be scanned.
+
+When a status event is received where the description ends with
+"(direct trigger)", the webhook will scan the pull request with the revision
+equal to the sha1 associated with the status for a specific test to run. The
+test will be put in a distributed queue to later be consumed by other task bots.
--- a/tasks/cockpit-tasks-webhook.yaml
+++ b/tasks/cockpit-tasks-webhook.yaml
@@ -17,6 +17,19 @@ items:
           infra: webhook
       spec:
         containers:
+          - name: amqp
+            image: rabbitmq:3-management
+            ports:
+            - containerPort: 5671
+              protocol: TCP
+              name: amqp-secure
+            volumeMounts:
+            - name: webhook-secrets
+              mountPath: /run/secrets/webhook
+              readOnly: true
+            - name: amqp-config
+              mountPath: /etc/rabbitmq
+              readOnly: true
           - name: webhook
             image: cockpit/tasks
             ports:
@@ -31,6 +44,9 @@ items:
         - name: webhook-secrets
           secret:
             secretName: webhook-secrets
+        - name: amqp-config
+          configMap:
+            name: amqp-config
         serviceAccountName: create-job
 
 - kind: Service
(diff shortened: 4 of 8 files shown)


tasks/webhook
```diff
@@ -62,12 +70,40 @@ def setup():
         shutil.copyfile(src, dest)
     os.umask(old_umask)
 
+def ensure_cockpit_checkout():
+    if not os.path.isdir(HOME_DIR + '/cockpit'):
```
martinpitt: should also use `COCKPIT_CHECKOUT`

tasks/webhook
```diff
@@ -62,12 +70,40 @@ def setup():
...
     def handle_event(self, event, request):
-        if event != 'create':
-            return (501, 'unsupported event ' + event)
+        if event == 'create':
+            return self.handle_create_event(event, request)
+        elif event == 'pull_request':
+            return self.handle_pull_request_event(event, request)
+        return (501, 'unsupported event ' + event)
 
+    @classmethod
```
martinpitt: for consistency, make this an instance method as well.

tasks/webhook
```diff
@@ -62,12 +71,62 @@ def setup():
...
+            return None
+
+        number = 0
+        try:
+            number = int(request['number'])
+        except (ValueError, TypeError):
+            return None
+
+        ensure_cockpit_checkout()
+        env = {}
```
martinpitt: It's a bit surprising that `tests-scan` works with a completely empty environment. Shouldn't it at least maintain `$PATH`? Initializing to `os.environ.copy()` seems harmless enough.

tasks/webhook
```diff
@@ -62,12 +71,62 @@ def setup():
...
+        if 'GITHUB_BASE' in os.environ:
+            env['GITHUB_BASE'] = os.environ.get('GITHUB_BASE')
+        subprocess.check_call(['bots/tests-scan', '--pull-number', str(number), '--amqp', 'amqp.cockpit.svc.cluster.local', '--queue', 'tasks'], cwd=COCKPIT_CHECKOUT, env=env)
+        return None
+
+    def handle_status_event(self, event, request):
+        if request['state'] != 'pending':
+            return None
+        # Only allow manually triggered tests
+        if not request.get('description', '').endswith('(direct trigger)'):
```
martinpitt: This, as well as the other magic value (`[no-test]`) deserve a comment at the top or in the README.

Also, `rstrip()` for robustness?

tasks/webhook
```diff
@@ -105,7 +164,12 @@ class ReleaseHandler(github_handler.GithubHandler):
 
 logging.basicConfig(level=logging.DEBUG)  # INFO
 
+# ensure we have a passwd entry for random UIDs
+# https://docs.openshift.com/container-platform/3.7/creating_images/guidelines.html
+subprocess.call(PASSWD_ENTRY_SCRIPT, shell=True)
```
martinpitt: Let's do `check_call()`. Sorry if I didn't do that in the existing place.

tasks/webhook
```diff
@@ -62,12 +71,62 @@ def setup():
         shutil.copyfile(src, dest)
     os.umask(old_umask)
 
+def ensure_cockpit_checkout():
+    if not os.path.isdir(COCKPIT_CHECKOUT):
+        url = 'https://github.com/cockpit-project/cockpit/'
+        subprocess.check_call(['git', 'clone', url, COCKPIT_CHECKOUT])
+    subprocess.check_call(['git', '-C', COCKPIT_CHECKOUT, 'fetch', 'origin'])
```
martinpitt: Wouldn't *this* be a good place for a temporary hack, i. e. checking out your repo/branch instead? Then we can watch the queue get filled and experiment with it, without having to touch the existing runners at all. If you are already past that stage, then ignore this of course.

tasks/cockpit-amqp.yaml
```diff
@@ -0,0 +1,33 @@
+---
+apiVersion: v1
+kind: List
+items:
+- kind: Pod
```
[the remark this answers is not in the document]
Gundersanne: Hm, the only downside would be when we rebuild the tasks container, the queue would have to build again, which for now is a manual affair still (going into the pod and calling tests-scan).

Though I think we can put the latter in webhook's  `setup()` and then it wouldn't make a difference.

tasks/cockpit-tasks-webhook.yaml
```diff
@@ -24,10 +24,16 @@ items:
...
+            - name: secrets
+              mountPath: "/secrets"
+              readOnly: true
             - name: webhook-secrets
               mountPath: /run/secrets/webhook
               readOnly: true
         volumes:
+        - name: secrets
+          secret:
+            secretName: cockpit-tasks-secrets
```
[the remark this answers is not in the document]
Gundersanne: Ack, will move the github token out.

tasks/webhook
```diff
@@ -62,12 +71,62 @@ def setup():
...
+        if request['pull_request']['title'].endswith('[no-test]'):
+            return None
+        if request['action'] == 'closed':
+            return None
+
+        number = 0
+        try:
+            number = int(request['number'])
+        except (ValueError, TypeError):
+            return None
```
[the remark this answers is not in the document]
Gundersanne: I'll remove the parsing. Returning None means success, returning None here wasn't correct.

tasks/webhook
Gundersanne: will do that

tasks/webhook
Gundersanne: oh this is just for the publishing, the consumers need to have the hack branch too, which is why it needs to be edited in 2 places. Once in the webhook and once in cockpit-tasks

tasks/webhook
Gundersanne: actually I think I could just leave that out, since check_call with env=None would just inherit the current environment, no?

tasks/webhook
```diff
@@ -62,12 +71,62 @@ def setup():
...
+        elif event == 'pull_request':
+            return self.handle_pull_request_event(event, request)
+        elif event == 'status':
+            return self.handle_status_event(event, request)
+        return (501, 'unsupported event ' + event)
+
+    def handle_pull_request_event(self, event, request):
+        if request['pull_request']['title'].endswith('[no-test]'):
+            return None
+        if request['action'] == 'closed':
```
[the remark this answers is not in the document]
martinpitt: Thanks. With so many action types that should not trigger a test (reviews, labels, etc.), shouldn't that be a whitelist for `opened` and `edited`?

tasks/cockpit-amqp.yaml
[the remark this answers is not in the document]
martinpitt: Right, that was the idea. Then it would always self-initialize. We want to do this every few weeks anyway to make sure it keeps working.

tasks/webhook
martinpitt: Oh, you are right -- `GITHUB_BASE` would just get inherited.

tasks/webhook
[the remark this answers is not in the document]
Gundersanne: And `synchornized` (for force pushes). I guess by removing/adding labels etc you could duplicate items in the queue. Whitelist might be cleaner.

martinpitt:
I split out the webhook secrets mounting into the tasks container into PR #241, as that is urgent and required right now. So this will need to be rebased afterwards.

amqp/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM rabbitmq:3-management
+
+COPY enabled_plugins /etc/rabbitmq/
+COPY rabbitmq.conf /etc/rabbitmq/
+COPY definitions.json /etc/rabbitmq
```
martinpitt: Trivial thing, but can you please merge these as `COPY enabled_plugins rabbitmq.conf definitions.json /etc/rabbitmq/` to avoid two unnecessary extra docker layers?

Makefile
```diff
@@ -103,3 +103,18 @@ learn-container:
 
 learn-push:
 	base/push-container docker.io/cockpit/learn
+
+amqp-shell:
+	$(DOCKER) run -ti --rm \
+		--publish=5671:5671 \
+		--volume=$(TASK_SECRETS):/run/webhook/secrets:ro \
```
martinpitt: That looks wrong? should be `WEBHOOK_SECRETS`?

amqp/definitions.json
```diff
@@ -0,0 +1,50 @@
...
+            "hashing_algorithm":"rabbit_password_hashing_sha256",
+            "tags": ""
+        },
+        {
+            "name": "worky",
+            "password_hash":"pB0/JnjGQ9tS8/w9gU7Fx8yRsquH54tnX3bufGZ0KQqho5Wz",
+            "hashing_algorithm":"rabbit_password_hashing_sha256",
+            "tags": ""
+        }
+    ],
```
martinpitt: As discussed on IRC, let's please just keep one user. Unfortunately JSON can't have comments, but please at least mention in the commit  message that the password is never being used (due to SSL auth), and thus a public static password is ok.

amqp/Dockerfile
martinpitt: More generally, do we really need our own container just for this? Can't we use the official container and add in /etc/rabbitmq/ as a volume? This would also make it easier to change the configuration later on.

amqp/enabled_plugins
```diff
@@ -0,0 +1 @@
+[rabbitmq_auth_mechanism_ssl,rabbitmq_management].
```
martinpitt: The trailing dot looks weird. Intended/necessary, or a typo?

tasks/Dockerfile
```diff
@@ -65,6 +64,8 @@ RUN dnf -y update && \
 
 COPY cockpit-tasks install-service webhook github_handler.py /usr/local/bin/
 
+RUN pip3 install pika
```
martinpitt: Please add a comment why the python3-pika RPM doesn't work. Also, please move this into the previous `RUN` block to avoid even more docker image layers. It semantically fits well into the "package install" layer.

tasks/README.md
```diff
@@ -161,4 +162,24 @@ Add a webhook to your GitHub project on the Settings → Webhooks page of your p
...
  * Select "Let me select individual events" and let the hook run on "Branch or
-   tag creation", and nothing else.
+   tag creation", "Pull requests", and "Statuses".
+
+See [GitHub development documentation](https://developer.github.com/webhooks/)
+for more information.
+
+### Automated releases
+
+Whenever a new release tag gets pushed to a project a creation event is received
```
martinpitt: It's "create", no? Let's please match the official event names to avoid confusion.

tasks/cockpit-tasks-webhook.yaml
```diff
@@ -17,6 +17,19 @@ items:
...
         containers:
+          - name: amqp
+            image: cockpit/amqp
+            ports:
+            - containerPort: 5671
+              protocol: TCP
+              name: amqp
+            - containerPort: 5672
+              protocol: TCP
+              name: amqp
```
martinpitt: Why does this need two ports? With the same name even?

tasks/webhook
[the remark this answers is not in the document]
martinpitt: This is still relevant, AFAICS.

amqp/enabled_plugins
Gundersanne: necessary (https://www.rabbitmq.com/plugins.html#ways-to-enable-plugins); I agree it looks weird

tasks/cockpit-tasks-webhook.yaml
Gundersanne: this is incorrect, and I plan on only exposing 5671

amqp/Dockerfile
Gundersanne: hmm I created our own Dockerfile because at one point i wanted to change the entrypoint, but scrapped that later on. I'll try it with a simple volume.

Makefile
Gundersanne: yea I was a bit confused as that variable isn't defined in the Makefile.  think tasks-shell should mount those secrets too, but wasn't sure what the actual path should be in `/var/lib/cockpit-tasks/?`

martinpitt: FTR, it's `/var/lib/cockpit-secrets/{release,tasks,webhook}/` now.

amqp/Dockerfile
martinpitt: This might be what we are looking for: https://kubernetes.io/docs/concepts/storage/volumes/#configmap

However, I'm fine with merging the PR with the current cockpit/amqp container (and the other fixes), and keeping this as a future optimization.

tasks/webhook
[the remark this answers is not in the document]
Gundersanne: I think this patch says
```
        title = request['pull_request']['title']
        number = int(request['number'])
        action = request['action']
```

that is correct, no?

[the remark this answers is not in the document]
martinpitt: Sorry, I mis-read the "returning None is not correct", but it is. All good!

Makefile
```diff
@@ -6,7 +6,8 @@ all:
 	@echo "       make release-install" >&2
 
 TAG := $(shell date --iso-8601)
-TASK_SECRETS := /var/lib/cockpit-tasks/secrets
+TASK_SECRETS := /var/lib/cockpit-tasks/tasks
```
martinpitt: It's /var/lib/cockpit-secrets/tasks, not *-tasks/tasks

tasks/cockpit-tasks-webhook.yaml
```diff
@@ -57,3 +73,79 @@ items:
...
+  spec:
+    to:
+      kind: Service
+      name: amqp
+    port:
+      targetPort: amqp-secure
+    tls:
+      termination: passthrough
+
+- kind: ConfigMap
```
martinpitt: Thanks for figuring this out. This is beautiful!

tasks/cockpit-tasks-webhook.yaml
```diff
@@ -57,3 +73,79 @@ items:
...
+  apiVersion: v1
+  metadata:
+    name: amqp-config
+  data:
+    definitions.json: |
```
martinpitt: YAML does have comments, so this would be a good place to mention that the password cannot actually be used, thus exposing it doesn't matter.

PR #231, opened 2018-12-05
sink: Improve error for state collisions

martinpitt:
Right now it just says

    Expected [expected data] but not present

Also show what the actual state is, so that we can debug what's going
wrong with collision detection.

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -109,7 +109,8 @@ def contains(expect, result, debug=True):
                     found += 1
             if found == 0:
                 if debug:
-                    sys.stderr.write("Expected " + json.dumps(exp) + " but not present\n")
+                    sys.stderr.write("Expected " + json.dumps(exp) + " but not present in " +
+                                     json.dumps(result) + "\n")
                 return False
             elif found > 1:
                 if debug:


martinpitt:
This is for finding out what's wrong in https://github.com/cockpit-project/cockpit/pull/10760, https://github.com/cockpit-project/cockpit/pull/10758, and https://github.com/cockpit-project/cockpit/pull/10759, where all/many tests get stuck on

    Expected {"target_url": "http://fedorapeople.org/groups/cockpit/logs/pull-10759-20181205-080101-fcb49d1f-verify-rhel-8-0/log.html", "description": "Testing in progress [bos-tasks-k38b6]", "context": "verify/rhel-8-0"} but not present
State not as expected. Possible collision. Aborting.

For [this case](https://fedorapeople.org/groups/cockpit/logs/pull-10759-20181205-080101-fcb49d1f-verify-rhel-8-0/log.html), the [API says exactly the same values](https://api.github.com/repos/cockpit-project/cockpit/statuses/fcb49d1f8083928c95682881e79b19faccec7054), so it's totally not clear where the collision comes from.

This change should tell us more, I rolled it out to fedorapeople.org.

PR #233, opened 2019-01-11
Use podman when available instead of docker

stefwalter:


stefwalter:
Tested by building the learn container and pushing the learn container multiple times.

PR #234, opened 2019-01-31
learn: Use PCA instead of MDS

emmil:
By treating each row of the distance matrix as a multidimensional
vector, PCA can be used for drawing cluster content.
This is a significant speed up compared to MDS


emmil:
Self NACK.

The MDS produces nicer graphs.

PR #235, opened 2019-02-04
release, tasks: Fix secrets building

martinpitt:
Remove the github-webhook-token special-cases. The master secrets are
now properly organized on cockpit-9 in /var/lib/cockpit-secrets/, into
tasks/, webhook/, and release/ subdirectories.


martinpitt:
I re-deployed all secrets on both clouds, and restarted the webhook and tasks RCs everywhere.

PR #237, opened 2019-02-06
secrets cleanups

martinpitt:


martinpitt:
I rebuilt base/tasks/images containers with these and redeployed. Tests seem happy again.

PR #238, opened 2019-02-06
tasks: Delete job object before creating it

martinpitt:
This is useful to be able to restart a failed release. Otherwise that
will fail because the release job already exists.


martinpitt:
I ran the `oc delete job` command manually after the 187 release, and the job object was correctly deleted. I rebuilt and re-deployed a new cockpit/tasks webhook ccontainer with this change. Now we should do another release to validate that it works.

PR #240, opened 2019-02-11
tasks: Install libappstream-glib

martinpitt:
We need this for validating AppStream metadata, see
https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/#_app_data_validate_usage


martinpitt:
I built and rolled out the tasks container with this, testing in https://github.com/weldr/welder-web/pull/534

martinpitt:
https://github.com/weldr/welder-web/pull/534 tests succeed, so this works.

PR #241, opened 2019-02-13
tasks: Mount webhook secrets into tasks container

martinpitt:
The GitHub token has moved to the webhook secrets a while ago, and
the copy got dropped from the tasks secrets.


martinpitt:
Last night's redeployment caused all of our tasks pods to not have the GitHub token. This fixes it, I rolled this out.

PR #242, opened 2019-02-18
tasks: Sleep after main cycle

Gundersanne:
@martinpitt  I just took your patch, and added the `work=` on line 116

### DIFF
--- a/tasks/cockpit-tasks-webhook.yaml
+++ b/tasks/cockpit-tasks-webhook.yaml
@@ -149,4 +149,3 @@ items:
         auth_mechanisms.1 = EXTERNAL
         ssl_cert_login_from = common_name
         management.load_definitions = /etc/rabbitmq/definitions.json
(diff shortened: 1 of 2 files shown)


tasks/cockpit-tasks-webhook.yaml
```diff
@@ -149,4 +149,4 @@ items:
         auth_mechanisms.1 = EXTERNAL
         ssl_cert_login_from = common_name
         management.load_definitions = /etc/rabbitmq/definitions.json
-        heartbeat = 360
+        heartbeat = 720
```
martinpitt: It seems a bit weird to change it on this end -- don't the heartbeats from the clients come through? If they do, but are sent too rarely, then it would be better to increase their frequency on the client side. If they don't, we should disable them completely (value 0).

But, this won't break anything, so let's try that.

PR #244, opened 2019-02-28
tasks: Only pull latest cockpit in webhook when new PR is merged

Gundersanne:
- [x] deploy this
- [ ] test this

### DIFF
--- a/tasks/webhook
+++ b/tasks/webhook
@@ -100,17 +100,21 @@ class ReleaseHandler(github_handler.GithubHandler):
         title = request['pull_request']['title']
         number = int(request['number'])
         action = request['action']
+        merged = request['pull_request'].get('merged', False)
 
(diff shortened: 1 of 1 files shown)


Gundersanne:
If this looks decent I'll restart the webhook with the rebuilt container

tasks/webhook
```diff
@@ -100,17 +100,21 @@ class ReleaseHandler(github_handler.GithubHandler):
         title = request['pull_request']['title']
         number = int(request['number'])
         action = request['action']
+        merged = request['pull_request']['merged']
```
martinpitt: Not sure if that's guaranteed to exist? The common approach is `.get('merged', False)`, but maybe like that it's better initially, so that it doesn't always silently default to False even when we expect it to exist.

PR #245, opened 2019-03-07
release: Don't put tabs into .spec files

martinpitt:
This causes rpmlint warnings:

    cockpit-composer.src:4: W: mixed-use-of-spaces-and-tabs (spaces: line 4, tab: line 2)

Keep the alignment for a 17-character first column, as we use in cockpit
and welder-web.

PR #246, opened 2019-03-07
release: Fix container build

martinpitt:
Recently, the container build started to fail with

    Problem: problem with installed package whois-mkpasswd-5.4.1-1.fc28.x86_64
    - package whois-mkpasswd-5.4.1-1.fc28.x86_64 conflicts with expect provided by expect-5.45.4-1.fc28.x86_64

We don't need mkpasswd, so remove it first.


martinpitt:
I ran into that trying to rebuild the container for PR #245. With this hack it succeeds again: https://hub.docker.com/r/cockpit/release/tags/

martinpitt:
This has done today's cockpit-composer and cockpit-podman releases, so proved to work.

PR #247, opened 2019-03-08
webhook: Update checkouts when merging no-test PRs

martinpitt:
Do the "[no-test]" check later, to avoid short-circuiting the git
update.


martinpitt:
I rebuilt the container, uploaded to dockerhub, and deployed new webook.

PR #248, opened 2019-03-11
Drop cockpit/base container

martinpitt:
This makes it easier to upgrade to newer Fedora releases independently from the other containers, makes containers smaller, and easier to rebuild and maintain.


martinpitt:
All containers are rebuilt and deployed. images and learn work fine, tasks has started to process PRs (will check again for results later). release hasn't been field-tested yet; but contingency plan is to just revert the "latest" tag to the previous version if  stuff fails.

martinpitt:
Unblocking, this works well enough.

PR #249, opened 2019-03-12
release: sink configuration fixes

martinpitt:

### DIFF
--- a/tasks/cockpit-tasks-webhook.yaml
+++ b/tasks/cockpit-tasks-webhook.yaml
@@ -40,6 +40,9 @@ items:
             - name: webhook-secrets
               mountPath: /run/secrets/webhook
               readOnly: true
+            env:
+            - name: RELEASE_SINK
+              value: sink
         volumes:
         - name: webhook-secrets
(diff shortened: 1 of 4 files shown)


martinpitt:
This is rolled out now.

tasks/cockpit-tasks-webhook.yaml
```diff
@@ -40,6 +40,10 @@ items:
             - name: webhook-secrets
               mountPath: /run/secrets/webhook
               readOnly: true
+            env:
+            - name: RELEASE_SINK
+              value: sink
+          - name: AMQP_SERVER
```
Gundersanne: This is an error I think.

Plus the webhook doesn't recognize `AMQP_SERVER` at this point, though that would probably be a good improvement.

martinpitt: Whoops, thanks for spotting! Fixed. (copy&paste error indeed).

I didn't dare to re-deploy the webhook yet, as I just pulled out CI from a huge swamp :/

PR #250, opened 2019-03-12
tasks: Revert to Fedora 28

martinpitt:
There are too many cockpit test failures with Fedora 29 (e. g. qemu
monitor API), these need to be sorted out first.


martinpitt:
@Gundersanne: I'm currently rebuilding and re-deploying, setting blocked until this is verified to work. Sorry for the mess-up last time! Can you please already review this?

martinpitt:
Yep, that works much better.

PR #251, opened 2019-03-12
tasks: Drop cockpit checkout and npm install

martinpitt:
The `npm install` was meant as pre-caching in the past, but it is moot
now: cockpit-tasks runs `git clean` as pretty much the first thing it
does. The webhook container entirely ignores this and does its own 
checkout. So let's save these > 400 MB from the docker image.

Also drop the checkout -- cockpit-tasks will do that, it doesn't take
long, and it's never going to be really useful as-is, as it quickly gets
out of date.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -63,8 +63,6 @@ RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user && \
     ln -snf /secrets/rhel-password /home/user/.rhel/pass && \
     ln -snf /secrets/zanata.ini /home/user/.config/zanata.ini && \
     ln -snf /run/secrets/webhook/.config--github-token /home/user/.config/github-token && \
-    git clone https://github.com/cockpit-project/cockpit /build/cockpit && \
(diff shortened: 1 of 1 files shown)


martinpitt:
As it's been my hobby of the day to meddle with our containers, let's round it up with this. :-)
```
docker.io/cockpit/tasks     2019-03-12          b40fc0b33f51        11 minutes ago      2.47 GB
docker.io/cockpit/tasks     <none>              e2a49b14bae4        10 hours ago        2.78 GB
```
So quite a noticeable and cheap size drop. Setting blocked until that got fully field tested.

martinpitt:
Deployed, and this works fine. The webhook acts normally, and tasks have successfully ran some tests, for example:
 - https://209.132.184.41:8493/logs/pull-11179-20190312-203737-9cf5e64f-verify-rhel-8-0/log.html
 - https://209.132.184.41:8493/logs/pull-11179-20190312-203737-9cf5e64f-verify-centos-7/log.html

@Gundersanne , can I ask you to review again, please?

PR #252, opened 2019-03-13
images: Serve files under /logs/ as text/plain

mvollmer:
Sometimes a cockpit/images container is used as a log sink as well,
and then we want files like "log", "status", and
"TestStorage-blah.FAIL.log" to be served as text/plain.


martinpitt:
Rolled out to CentosCI cloud: https://images-cockpit.apps.ci.centos.org/  Works fine.

PR #253, opened 2019-03-13
Makefile: Fix breakage if both podman and docker are installed

martinpitt:


martinpitt:
Without this:
```
❱❱❱ sudo  make images-container
/usr/bin/podman /usr/bin/docker build -t docker.io/cockpit/images:2019-03-13 images
Error: unknown shorthand flag: 't' in -t
```

PR #254, opened 2019-03-14
tasks: Add wget

martinpitt:
It was previously pulled in through a dependency. Cockpit's
atomic.bootstrap uses that, and replacing it with curl is a bit
inconvenient.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -35,6 +35,7 @@ RUN dnf -y update && \
         sudo \
         tar \
         virt-install \
+        wget \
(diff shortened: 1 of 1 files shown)


martinpitt:
I rolled this out now. The [fedora-atomic  refresh](https://github.com/cockpit-project/cockpit/issues/11388) works again, and it also successfully [ran a cockpit test](https://209.132.184.41:8493/logs/pull-11396-20190314-084646-de4d9ed4-verify-fedora-29/log.html).

PR #255, opened 2019-03-25
release: Add option to sit on failure

martinpitt:
If a release fails, the release container stops and makes it hard to
debug the failure or finish the release manually. For that, introduce a
new `-s` (sit) option that will sleep indefinitely instead.

As with a sink the actual release-runner runs in a subprocess, we can't
just pass a variable to the parent; so let the child process remove a
temporary flag file on success.

Enable this option in the webhook.

### DIFF
--- a/release/release-runner
+++ b/release/release-runner
@@ -24,6 +24,7 @@
 # -z         RELEASE_CHECK=1        Check access and configuration
 # -l         RELEASE_SINK=1         Host to sink logs to
 # -t         RELEASE_TAG=0.X        Tag to release
+# -s                                Keep running in infinite sleep when release fails
 #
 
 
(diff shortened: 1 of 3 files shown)


martinpitt:
I locally tested this with a fake cockpit-ostree release and verified that it sits properly with the option, and works normally without.

I rebuilt and pushed the cockpit/release and cockpit/tasks containers and re-deployed the webhook container. I'll now do a welder-web release to give it a full exercise.

martinpitt:
The [welder-web 0.2.0 release](https://fedorapeople.org/groups/cockpit/logs/release-welder-web-0.2.0/log) was done successfully with the new containers.

PR #256, opened 2019-03-25
webhook: Increase receive buffer size

martinpitt:
GitHub's POST requests are pretty big (~ 30 kB) due to the large JSON
blob. Processing a new pull request takes ~ 5 seconds due to tests-scan,
so we are currently prone to lose requests to "Connection refused".

Increase the default buffer size to fit more pending requests.

### DIFF
--- a/tasks/webhook
+++ b/tasks/webhook
@@ -2,6 +2,7 @@
 
 import os
 import logging
+import socket
 import subprocess
 import shutil
 import http.server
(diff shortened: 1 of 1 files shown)


martinpitt:
It's not yet clear whether this really is the reason for our ignored test requests and frequent service timeouts, but it's worth a shot. This still needs the container be rebuilt and rolled out.

martinpitt:
I want to add logging of the current buffer size, for better debugging.

martinpitt:
This doesn't actually seem to help, we are still getting force-pushes without test requests.

PR #257, opened 2019-04-01
webhook: Check payload for no-test label

martinpitt:
Tests-scan also checks the labels, but apparently /issues/#/labels is
not synchronous right at PR creation. This caused interactively created
PRs with the label to still get all test requests.


martinpitt:
This container rebuild apparently pulled in a new python-pip version, which changes API. tests-scan now crashes with
```
File "bots/tests-scan", line 183, in distributed_queue
 server_hostname=host),
 TypeError: __init__() got an unexpected keyword argument 'ssl_version' 
```

martinpitt:
I validated in https://github.com/cockpit-project/starter-kit/pull/158 that this works fine.

PR #258, opened 2019-04-04
tasks: Check for tasks in cockpit-podman

martinpitt:
Welcome to the bots family!


martinpitt:
I build/uploaded/deployed the new container. I'll watch the bots for a bit and unblock/land when everything is right.

martinpitt:
We had a few successful test runs with these.

PR #259, opened 2019-04-09
webhook: Use tests-scan --pull-data option

martinpitt:
GitHub webhooks suffer from a race condition: When getting a PR event,
calling the API about that very PR often delivers outdated results.  But
the webhook event already gets the entire `pull_request` object as part
of its payload, so let's just use that instead of querying it from the
API again.

https://github.com/cockpit-project/cockpit/pull/11556 introduced a new
`--pull-data` option to tests-scan, so pass the data to it.

This also finally obsoletes checking for the no-test label, so revert
commit f7918f6af.

Fixes https://github.com/cockpit-project/cockpit/issues/11554

 - [x] requires https://github.com/cockpit-project/cockpit/pull/11556
 - [x] test: https://github.com/cockpit-project/cockpit/issues/11558

### DIFF
--- a/tasks/webhook
+++ b/tasks/webhook
@@ -6,6 +6,7 @@ import subprocess
 import shutil
 import http.server
 import pipes
+import json
 
 import github_handler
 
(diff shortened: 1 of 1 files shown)


martinpitt:
I built/pushed the container and rolled this out into the current webhook pod. debian-stable refresh should be triggered soon.

martinpitt:
Tests were triggered in https://github.com/cockpit-project/cockpit/pull/11558 and the image refresh went alright.

PR #264, opened 2019-05-15
tasks: Update manual docker mode

martinpitt:
- Pass the webhook secrets as a volume
- Pass our CentOS-CI AMQP server
- Switch to our current sink
- Drop obsolete --privileged and --uts options, our containers run fine
  unprivileged
- Set default TEST_JOBS to 8, which is right for our beefy machines
- Use a systemd template unit, and run three instances on every host

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -53,15 +53,14 @@ Make sure docker and atomic are installed and running:
 
 You may want to customize things like the operating system to test or number of jobs:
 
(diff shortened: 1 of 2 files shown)


martinpitt:
After our complete OpenShift breakage last night, I now rolled this out to our machines, and so far the containers seem to run in general. This isn't a mode which we want to use long-term, but it's useful as emergency measure to keep stuff running.

PR #271, opened 2019-05-16
tasks: Shorten initial sleep

martinpitt:
It's a bit unnerving having to wait for up to five minutes when
deploying new containers, and it's also not really necessary at that
time.


martinpitt:
I took the liberty of just landing this without a container rebuild. It's not urgent, and will just be picked up at the next regular rebuild, when we actually need it.

PR #272, opened 2019-05-17
tasks: Increase systemd unit timeout

martinpitt:
The `docker pull` operation usually takes more than one minute if it
actually has to pull the container. Don't let the service fail because
of that.


martinpitt:
I rolled this out onto our current infra, and it works.

PR #273, opened 2019-05-20
Fix systemd tasks containers

martinpitt:


martinpitt:
This is rolled out onto bos, and should finally fix our testing woes.

martinpitt:
One more slight fix, I reversed the %i and %H for the hostname, as %H expands to the FQDN and docker cuts it off after the first period. @Gundersanne , can you please review again?

PR #274, opened 2019-05-22
Drop unnecessary volumes

martinpitt:
docker creates anonymous, but persistent volumes for all `VOLUME` paths
that don't get an explicit one with `--volume`. These slowly pile up and
lead to `ENOSPC` errors on the hosts.

This mostly affects tasks' /tmp, but it also makes little sense to
create external anonymous volumes for the secrets when they don't get
specified (for manual usage).

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -16,7 +16,7 @@ RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user && \
 COPY install-service /usr/local/bin/
 COPY nginx.conf /etc/nginx/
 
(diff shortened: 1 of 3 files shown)


martinpitt:
I rebuilt and deployed new cockpit/images. Both the [public](https://images-cockpit.apps.ci.centos.org/) and the RH-internal ones work fine.

I also built/pushed/deployed the tasks container, cleaned up all leftover volumes. I'll watch some tests before landing.

martinpitt:
Several tests succeeded with this, unblocking.

PR #275, opened 2019-05-23
tasks: Increase root fs size

martinpitt:
Commit ab56ed3a08 dropped the automatic volume for /tmp. This caused
weldr/lorax tests to run out of space, as they need lots of temporary
space. Configure the container to have 50 GB root fs space instead of
the default 10 GB.

This is simpler than re-introducing a /tmp volume and having to clean it
up manually.


martinpitt:
This is rolled out, and seems to work at first sight. But let's wait until the previously failing [lorax test](https://github.com/weldr/lorax/pull/730) succeeds to double-check.

martinpitt:
Unblocking, this works fine.

PR #276, opened 2019-06-03
tasks: Add secrets symlink for lorax-test-env

martinpitt:
This has the cloud credentials for running lorax'es cloud image tests.


martinpitt:
I added and rolled out the secret everywhere, built and pushed the container, and deployed it. This needs to survive a few tests before calling it good.

martinpitt:
Several tests succeeded here, and the first cloud tests are passing in https://github.com/weldr/lorax/pull/745

PR #281, opened 2019-06-26
tasks: Change default release sink to cluster-local sink

martinpitt:
After commit 6fea47f28 we now have a local sink, which is more robust
than a remote one. Similarly to CentOS CI tasks (commit 702389db),
switch the default release sink to the local service.

### DIFF
--- a/tasks/cockpit-tasks-webhook.yaml
+++ b/tasks/cockpit-tasks-webhook.yaml
@@ -42,7 +42,7 @@ items:
               readOnly: true
             env:
             - name: RELEASE_SINK
-              value: sink
+              value: sink-local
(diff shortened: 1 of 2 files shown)


martinpitt:
I deployed that to our current webhook, but let's wait on a release to confirm it's working properly.

martinpitt:
Fails, the release secrets don't know about "sink-local", need to add this first.

martinpitt:
Updated the secret, let's try with another release.

martinpitt:
Worked :tada:  https://logs-https-cockpit.apps.ci.centos.org/logs/release-cockpit-podman-4/log

PR #282, opened 2019-06-26
Move issues-scan to queue

Gundersanne:


tasks/webhook
```diff
@@ -171,6 +173,19 @@ class ReleaseHandler(github_handler.GithubHandler):
         publish_to_queue('webhook', json.dumps(body))
         return None
 
+    def handle_issues_event(self, event, request):
+        action = request['action']
+        if event == 'issues' and action not in ['opened', 'synchronize', 'edited']:
```
martinpitt: What is "synchronize" for an issue?

Also, isn't the `event == 'issues'` redundant here?

tasks/webhook
```diff
@@ -171,6 +173,19 @@ class ReleaseHandler(github_handler.GithubHandler):
...
+            logging.info("action not in ['opened', 'synchronize'], skipping issue")
```
martinpitt: not the same set; how about making this more useful for debugging and say something like `"action %s unknown, skipping event" % action`?

tasks/webhook
```diff
@@ -146,7 +146,7 @@ class ReleaseHandler(github_handler.GithubHandler):
             return None
 
         body = {
-            'event': 'pull_request',
+            'event': event,
```
martinpitt: Good call, and then the whole "build body and publish it to webhook" can be factorized.

tasks/webhook
Gundersanne: wel li want to catch someone unchecking the checkbox; i should probably just see what event that is and catch that :)

Gundersanne: just realized then if we want to listen to checkboxes, we'd also need to listen to PRs, but a different action.

Gundersanne: just verified, checkboxes on issues are 'edited'.  On PRs we'll need to listen to the same, but pass it to issue-scan I guess ?

martinpitt: Right, whenever the description of a PR changes we need to pipe it through issue-scan.

Gundersanne:
reordered and squashed!

Gundersanne:
@marusak This is the current deployment. Seems to be working fine. As you can see the checkouts are now done in XDG_CACHE_HOME

PR #278, opened 2019-06-18
webhook: Pubish payload to queue directly

Gundersanne:
- [x] https://github.com/cockpit-project/cockpit/pull/12091

PR #283, opened 2019-06-26
webhook: Update documentation concerning the webhook queue

Gundersanne:
Related to #278

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -173,13 +173,16 @@ by the webhook which automatically starts a release runner.
 
 ### Automated testing
 
-When a pull request event is received, the webhook will scan the pull request
-for tests that need to be run. These will be put in a distributed queue to later
-be consumed by other task bots. If "[no-test]" is in the title of the pull
-request, it won't be scanned (this is implemented in cockpit's
-bots/tests-scan).
+When an pull request event or a status event is received, the webhook will
+publish the event to the webhook AMQP queue. These are then consumed and
+interpreted from the webhook queue by task bots. Task bots will publish tasks to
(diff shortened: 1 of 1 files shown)


tasks/README.md
```diff
@@ -173,13 +173,19 @@ by the webhook which automatically starts a release runner.
 
 ### Automated testing
 
-When a pull request event is received, the webhook will scan the pull request
-for tests that need to be run. These will be put in a distributed queue to later
-be consumed by other task bots. If "[no-test]" is in the title of the pull
+When a pull request event is received, the webhook will publish the event to a
+distributed queue. These events are then consumed and interpreted by task bots
```
martinpitt: As that part will be the same for any webhook event, I think these two sentences should be factored out into a first paragraph by themselves. "a distributed queue" should be more specific, like "to the webhook AMQP queue".

tasks/README.md
```diff
@@ -173,13 +173,19 @@ by the webhook which automatically starts a release runner.
...
+who will publish tasks to the relevant queues, to be consumed again by task bots
+at a later point in time. If "[no-test]" is in the title of the pull
 request, it won't be scanned (this is implemented in cockpit's
 bots/tests-scan).
```
martinpitt: Let's drop the [no-test] part here. With the recent changes, cockpituous is now completely oblivious of this.

tasks/README.md
Gundersanne: ack!

tasks/README.md
Gundersanne: ack

PR #285, opened 2019-08-02
learn: expose cockpit-learn over ssl

Gundersanne:


learn/Dockerfile
```diff
@@ -3,12 +3,24 @@ LABEL maintainer='cockpit-devel@lists.fedorahosted.org'
...
+
+RUN mkdir -m 775 -p /cache/train \
+ && mkdir -m 775 -p /home/user \
+ && mkdir -m 775 -p /cache/client_body_temp \
+ && ln -sf /dev/stdout /var/log/nginx/access.log \
+ && ln -sf /dev/stderr /var/log/nginx/error.log \
+ && rm -rf /usr/share/nginx/html \
+ && ln -snf /cache/images /usr/share/nginx/html \
+ && groupadd -g 1111 -r user \
+ && useradd -r -g user -u 1111 user \
```
martinpitt: Can we avoid that for new containers? These won't work for CentOS CI deployments anyway, instead the volumes just need to be world-writable.

learn/nginx.conf
```diff
@@ -0,0 +1,66 @@
+user user user;
+worker_processes 1;
+error_log /var/log/nginx/error.log;
```
martinpitt: Could this just say /dev/stderr or similar directly, instead of jumping through the symlink hoop?

learn/nginx.conf
```diff
@@ -0,0 +1,66 @@
...
+events {
+    worker_connections 1024;
+}
+
+http {
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
```
martinpitt: Same here.

learn/nginx.conf
```diff
@@ -0,0 +1,66 @@
...
+worker_processes 1;
+error_log /var/log/nginx/error.log;
+pid /dev/null;
+daemon off;
+
+# Load dynamic modules. See /usr/share/nginx/README.dynamic.
+include /usr/share/nginx/modules/*.conf;
+
+env TEST_DATA=/cache;
+env HOME=/home/user;
```
martinpitt: Does it actually need a legit home dir? Would /tmp/ do, to avoid having to set up a user and permissions?

learn/nginx.conf
Gundersanne: this can say `stderr`

learn/nginx.conf
Gundersanne: sadly the http module only accepts files, so we need the symlink here

learn/nginx.conf
Gundersanne: just removed, works fine without. Copied over a bit too much :)

learn/Dockerfile
Gundersanne: seems to work fine without, running stuff as the nginx user and making the dirs world-writeable

learn/cockpit-learn.yaml
```diff
@@ -46,3 +47,14 @@ items:
...
       targetPort: 8080
       protocol: TCP
+- kind: Route
+  apiVersion: v1
+  metadata:
+    name: cockpit-learn
+  spec:
+    selector:
+      infra: cockpit-learn
+    path: "/predict"
```
martinpitt: Is that intentional? prediction and learning are two opposite functions of this, having it in the same pod feels weird? Or do we run learning (log scanning and json creation) and prediction (REST API to map a log to a "flake?") on the same container?

Gundersanne:
@martinpitt  while this does result in a service that's actually online for both bos and centos now, it still doesn't work properly. The amount of data cockpit now generates every 120 days result in this being too slow (predict requests take up several seconds and even time out). So to get this into a useful state would take more work.

And with marius's retry each test max 3 times in one single run, the need for accurately predicting flakes has gone down by a lot.

PR #286, opened 2019-08-08
tasks: Support custom NPM registry

martinpitt:
For the Red Hat internal tasks machines we want to use an internal
mirror to lessen the strain on npmjs.com.

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -33,6 +33,12 @@ if [ $loop != "yes" ]; then
     exec "$@"
 fi
 
+# set up custom NPM registry
+if [ -n "$NPM_REGISTRY" ]; then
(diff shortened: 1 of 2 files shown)


martinpitt:
This is rolled out now. On bos I set the internal NPM registry, and it works. The most obvious proof is [this log](https://209.132.184.41:8493/logs/pull-12519-20190808-121515-3de363b8-cockpit-project-cockpit-rhel-8-1/log.html) where the internal NPM registry has some hiccup (that's bad, but let's see how often it happens). Still waiting for some successful test runs, after that I'll unblock.

martinpitt:
I reset the actual NPM_CONFIG setting on pittishift, as the internal registry has no valid SSL cert. But changing this doesn't need a container rebuild.

PR #287, opened 2019-08-13
tasks: Support extra CA for custom npm registry

martinpitt:
Using a non-default registry may need a custom SSL certificate. If
present in the secrets, set it through `$NODE_EXTRA_CA_CERTS`.

https://nodejs.org/api/cli.html#cli_node_extra_ca_certs_file


martinpitt:
I built/pushed the container and rolled it out on BOS with re-enabling the internal registry. Blocked until we get some successful test runs.

martinpitt:
There are some successful test runs, like https://209.132.184.41:8493/logs/pull-12554-20190813-064141-d5293456-cockpit-project-cockpit-debian-testing/log.html .

PR #290, opened 2019-08-29
tasks: Adapt to standalone bots project

Gundersanne:
TODO after:
- [ ] remove run_queue from .tasks cockpit-project/cockpit#12873
- [ ] remove bots/ from cockpit-project/cockpit#12872

My only concern is that this kinda messes with the whole chances system. Maybe po-refreshes and other tasks would get drowned out too much.

Currently `https://github.com/Gundersanne/cockpituous/tree/simulate-cockpit-bots` is running on some of our CI to simulate `bots/` being a separate project. This seems to be running fine.

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -102,8 +102,8 @@ function run_one_task() {
 
     update_repo "$1" "$2"
 
-    if [ ! -d "$dir/bots" ]; then
-        cp -drl "$XDG_CACHE_HOME"/cockpit-project/bots/bots "$dir/"
+    if [ ! -d "$dir/bots" ] && [ ! -L "$dir/bots" ]; then
+        cp -drl "$XDG_CACHE_HOME"/cockpit-project/bots "$dir/"
     fi
 
     # Get a task and execute it, 12 hours max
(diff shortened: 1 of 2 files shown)


tasks/cockpit-tasks
```diff
@@ -113,14 +113,29 @@ function run_one_task() {
...
 sleep $(shuf -i 0-120 -n 1)
 
 # Perform a couple of tasks or waits, then restart
 for i in $(seq 1 30); do
     work_done=
 
-    update_repo  "$XDG_CACHE_HOME"/cockpit-project/bots             https://github.com/cockpit-project/cockpit
+    update_repo  "$XDG_CACHE_HOME"/cockpit-project/bots             https://github.com/cockpit-project/bots
+
+    if [ -n "$AMQP_SERVER" ]; then
```
Gundersanne: Would it be better to have an 80% change to run something from the queue; otherwise run one of the tasks?

tasks/cockpit-tasks
```diff
@@ -113,14 +113,30 @@ function run_one_task() {
     fi
 }
 
+function run_from_queue() {
+    cd "$XDG_CACHE_HOME"/cockpit-project/bots
+    # Consuming directly from queue, commands have bots/ prepended
+    [ -e bots ] || ln -s . bots
+    if /usr/bin/timeout 12h /bin/sh -c "set -ex; ./run-queue --amqp \"$AMQP_SERVER\""; then
```
martinpitt: You can run `/bin/sh -exc ./run-queue ...` to save one level of quoting and make this look nicer.

tasks/cockpit-tasks
```diff
@@ -113,14 +113,30 @@ function run_one_task() {
...
 sleep $(shuf -i 0-120 -n 1)
 
 # Perform a couple of tasks or waits, then restart
 for i in $(seq 1 30); do
     work_done=
 
-    update_repo  "$XDG_CACHE_HOME"/cockpit-project/bots             https://github.com/cockpit-project/cockpit
+    update_repo  "$XDG_CACHE_HOME"/cockpit-project/bots             https://github.com/cockpit-project/bots
+
+    # 80% to try to consume from queue so other tasks don't get starved
```
martinpitt: <bikeshed> 20% for tasks is probably waaaaaay to much. We need to run them about once a day, and this loop iterates ~ 100 times every day if it could. In fact, it would be sufficient to call the `run_one_task` exactly once at startup, as the pod restarts after 30 iterations the loop anyway. This makes the script easier to understand, more predictable, and avoids the shuffling.

But if you prefer, this can happen in a separate PR (even before landing this one).

tasks/cockpit-tasks
```diff
@@ -113,19 +113,41 @@ function run_one_task() {
...
 
-    update_repo  "$XDG_CACHE_HOME"/cockpit-project/bots             https://github.com/cockpit-project/cockpit
-    run_one_task "$XDG_CACHE_HOME"/cockpit-project/cockpit          https://github.com/cockpit-project/cockpit
-    run_one_task "$XDG_CACHE_HOME"/cockpit-project/starter-kit      https://github.com/cockpit-project/starter-kit
-    run_one_task "$XDG_CACHE_HOME"/weldr/cockpit-composer           https://github.com/weldr/cockpit-composer
-    run_one_task "$XDG_CACHE_HOME"/cockpit-project/cockpit-podman   https://github.com/cockpit-project/cockpit-podman
+# check for tasks
+for P in $PROJECTS; do
+    work_done=
+    update_repo  "$XDG_CACHE_HOME"/cockpit-project/bots https://github.com/cockpit-project/bots
```
martinpitt: I don't think we need to update the bots repo for every task, as these tasks themselves can't modify bots master.

tasks/cockpit-tasks
```diff
@@ -113,19 +113,41 @@ function run_one_task() {
...
+    run_one_task "$XDG_CACHE_HOME"/"$P"                 https://github.com/"$P"
+    # Nothing to do, or failure
+    [ -n "$work_done" ] || slumber
```
martinpitt: Also, we don't need to sleep between these IMHO -- we didn't so far either, and it didn't hurt.

tasks/cockpit-tasks
```diff
@@ -113,19 +113,41 @@ function run_one_task() {
     fi
 }
 
+function run_from_queue() {
+    cd "$XDG_CACHE_HOME"/cockpit-project/bots
+    # Consuming directly from queue, commands generated by run-queue,
+    # tests-scan, and issue-scan have bots/ prepended
+    [ -e bots ] || ln -s . bots
+    if /usr/bin/timeout 12h /bin/sh -exc "./run-queue --amqp \"$AMQP_SERVER\""; then
```
martinpitt: What does the extra `sh -exc` actually give you, compared to just directly running run-queue in timeout?

Gundersanne: oh good point, I read the previous comment wrong.

martinpitt: That's fine, I wrote it wrongly in the first place :-)

Gundersanne:
As the symlink is now permanent, I removed it ln command.

PR #294, opened 2019-10-08
tasks: Stop copying bots into bots

martinpitt:
Only copy our bots checkout into a project directory if it the target
isn't the bots project  itself.

Also drop the redundant explicit `run_one_task()` call for bots, the
loop below will do that as the first thing.


martinpitt:
I rolled this out this morning, and tests are happy again.

PR #295, opened 2019-10-18
tasks: Allow npm registry and #instances to be set from environment

martinpitt:
This makes it easier to integrate this into Ansible playbooks.


martinpitt:
I tested this on the newly installed ci-srv-01 host by slurping in the install-service from this branch instead of from master. Worked fine.

PR #296, opened 2019-10-21
tasks: Add `firefox` so we can run tests on it

marusak:


KKoukiou:
> Wow, are you already that far along? Nice! Setting to blocked until the container rebuild is rolled out and survived a few test runs.

Whoa - does this mean that we can start dropping the duplicated tests from selenium-firefox?

martinpitt:
@KKoukiou : well, that's the first step only. We also need to make our tests run against Edge with CDP, and then pick a few verify tests that we want to run in selenium and fix them to work against an existing testbed. E. g. https://projects.engineering.redhat.com/browse/COCKPIT-472

martinpitt:
We got several successful runs with the new container. Landing.

PR #298, opened 2019-10-30
tasks: Add new zanata client

marusak:
`zanata-client` does not exist in Fedora 31. Also it requires a lot of
java deps.
`python3-zanata-client` provides `zanata` cli tool.
For now lets have both until all projects migrate to this new zanata
tool.

I'll build and deploy this, then open PR in cockpit which will use this new tool and if that refresh works, then we can land this.


marusak:
Bunch of tests were successful on centosci and on bos as well. Also https://github.com/cockpit-project/cockpit/pull/13067 worked as well.

PR #300, opened 2019-11-04
tasks: Lock down to older version of Firefox

marusak:
Install Firefox 69, since 70 has broken cdp, see https://bugzilla.mozilla.org/show_bug.cgi?id=1593226

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -1,12 +1,13 @@
 FROM fedora:30
 LABEL maintainer='cockpit-devel@lists.fedorahosted.org'
 
+# HACK: Install Firefox 69, since 70 has broken cdp, see https://bugzilla.mozilla.org/show_bug.cgi?id=1593226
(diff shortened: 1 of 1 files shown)


marusak:
@martinpitt do you agree with this approach? Fixed Firefox won't be here for at least another month (they won't fix it in F70, but only F71 earliest) and not sure how fast it gets fixed.

martinpitt:
I agree to the goal of pinning down firefox to 69 for the time being. I just doubt that the Fedora package indexes will keep older versions around for a longer time. I. e. you can pin down to the version at fedora 30 release (these composes are fixed), but not in anything in -updates. Fine for me to land this now, but for a rebuild we may need to move to hardcoding a koji URL of the latest 69 build.

martinpitt:
Some successful test runs in https://github.com/cockpit-project/cockpit/pull/13085 and in https://github.com/cockpit-project/bots/pull/204

PR #301, opened 2019-11-04
Adjust tasks and images for podman

martinpitt:
This will allow us to run our bots in AWS EC2 instances with RHEL 8.


martinpitt:
I tested both in actual EC2 instances. I tore them down again, as these are meant to be fallback solutions.

PR #302, opened 2019-11-18
tasks: PSI task bots

Gundersanne:
We've been using this config for quite a while now, even though it was scaled down.

### DIFF
--- a/tasks/cockpit-tasks-psi.yaml
+++ b/tasks/cockpit-tasks-psi.yaml
@@ -1,0 +1,65 @@
+---
+apiVersion: v1
+kind: List
+items:
+- apiVersion: v1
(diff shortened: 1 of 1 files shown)


tasks/cockpit-tasks-psi.yaml
```diff
@@ -0,0 +1,65 @@
...
+          infra: cockpit-tasks
+      spec:
+        containers:
+        - name: cockpit-tasks
+          image: docker.io/cockpit/tasks
+          env:
+          - name: TEST_JOBS
+            value: '8'
+          - name: TEST_PUBLISH
+            value: sink-infracloud
```
martinpitt: AFAIR this is just "sink" now, right?

PR #304, opened 2019-11-25
Fix sink on Fedora 31

martinpitt:
This is currently a hidden mine: As soon as we'll rebuild the cockpit/images container, this will break the sink.

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 from __future__ import print_function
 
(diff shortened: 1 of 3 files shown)


martinpitt:
This still doesn't work right, hang on.

martinpitt:
It doesn't work because tests-invoke calls the sink through an explicit "python", which is again Python 3. Argh!

martinpitt:
So this is necessary, but not sufficient, so again ready for review.

martinpitt:
The bots side of that fix is https://github.com/cockpit-project/bots/pull/275

PR #305, opened 2019-11-25
Clean up test invocation, pyflakes/pycodestyle fixes

martinpitt:
The example directory mostly seems to come from some semaphore tutorial.

Instead, move these bits to `make check`. Semaphore config already got
changed to invoke that instead.

### DIFF
--- a/example/semaphore-prepare
+++ b/example/semaphore-prepare
@@ -1,2 +1,0 @@
-docker pull fedora:25
-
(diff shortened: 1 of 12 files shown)


example/semaphore-prepare
```diff
@@ -1,2 +0,0 @@
-docker pull fedora:25
-
```
marusak: Why you removed the example directory? README still mentions example directory.

martinpitt: Good point, thanks! I removed it from README, too. As to the "why", it was rather bogus -- most semaphore scripts weren't ever been used, and this wasn't an example for something cockpituous-y at all.

PR #306, opened 2019-11-25
Port sink to Python 3

martinpitt:
- [x] This is based on PR #305, needs to land first
 - [x] Roll this out to fedora-infracloud sink, initial test
 - [x] Switch e2e back to fedora-infracloud, let a few tests run
 - [x] Fix semaphore test hang

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -1,6 +1,4 @@
-#!/usr/bin/env python2
-
-from __future__ import print_function
+#!/usr/bin/env python3
 
 import argparse
 import errno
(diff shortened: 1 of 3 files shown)


martinpitt:
Rolled out to fedora-infracloud sink, and did initial smoke test:
```
$ printf '{"status": "mocking"}\nsome logging\n' | ssh sink-infracloud ./sink foo
some logging
https://209.132.184.41:8493/logs/foo/
```

This and the generated log+status at https://209.132.184.41:8493/logs/foo/ look as expected.

I will now switch e2e back to sink-infracloud, as I still don't trust the fedorapeople.org sink enough to last for more than a few days (it blocked us off regularly in the past).

martinpitt:
At least [this](https://209.132.184.41:8493/logs/pull-13185-20191125-114747-4e904367-cockpit-project-cockpit--fedora-30-firefox/log.html) and [this](https://209.132.184.41:8493/logs/pull-13165-20191125-114848-8b98ba9f-cockpit-project-cockpit--fedora-30/log.html) test run ran against our fedorainfracloud sink (check the IP), and obviously that succeeded.

martinpitt:
I tracked down the semaphore hang, json.loads() does not get along with binary data in earlier Python 3.5. Fixed now, and rebased. I also rolled out the updated sink to sink-infracloud.

sink/sink
```diff
@@ -592,37 +576,39 @@ def main():
     # Initialize status reporters
     status = Status(config, identifier)
 
-    log = open("log", "w")
-    sys.stdout = Tee(log, sys.stdout)
-    sys.stderr = Tee(log, sys.stderr)
+    log = open("log", "wb")
+    # treat streams as binary stream, to avoid problems with non-ASCII data
+    global stdout, stderr
```
marusak: ech :/
Why you did not leave it as it was before? That you set up `sys.stderr` and `sys.stdout` and use them everywhere? Does not this work with python3?

martinpitt: This leads to all kinds of strange errors, as libraries and other internal code also uses sys.std*. Redefining them to be byte streams instead of a TextWrapper (string streams) wreaks havoc there. Therefore I thought it would be cleaner to have new variables, to point out that these are very much *not* the standard IO streams.

martinpitt: @marusak : Does the above explanation make sense, or should I expand further?

PR #308, opened 2019-12-11
webhook: Scan all projects on startup

marusak:


marusak:
I opened https://github.com/cockpit-project/starter-kit/pull/256 while webhook was down. When I started the webhook back in, those tests started to run. So it was picked upon starting.

martinpitt:
I rolled out the new container to PSI and e2e.

martinpitt:
This has successfully run several tests in #13186 at least, plus some bots updates.

PR #310, opened 2020-01-08
tasks: Use git credentials store

marusak:
So we are able to do `git push` directly.
See https://manpages.ubuntu.com/manpages/precise/en/man1/git-credential-store.1.html


martinpitt:
@marusak: The [container refresh two days ago](https://hub.docker.com/layers/cockpit/tasks/latest/images/sha256-2a2f09d1ee8dbded60319f96d413d380b837b6822b0792913c3d697a184996f0), does it include this PR? If so, it has run enough tests now, and we can land this.

PR #313, opened 2020-01-30
tasks: Run .tasks for cockpit-ostree

martinpitt:


martinpitt:
This by itself won't do anything yet, as cockpit-ostree doesn't have a .tasks file. I'll enable that later on, e. g. npm-update [does not quite work yet](https://github.com/cockpit-project/cockpit-ostree/pull/37).

martinpitt:
Pushed and rolled out.

martinpitt:
Got a few successes in https://github.com/cockpit-project/cockpit/pull/13456 , unblocking

PR #314, opened 2020-02-14
tasks: Allow install-service caller to override sink

martinpitt:


martinpitt:
This is similar to how we already handle a non-default NPM registry. I just created a new [public log sink in AWS](https://54.89.13.31/logs/) and want to test this before making this the new "sink" ssh host alias.

PR #316, opened 2020-03-09
Enable running tests in skobyda/cockpit-certificates

marusak:


martinpitt:
We can do that at some point to enable .tasks files, i. e. the cron-like stuff like npm-update or po-refresh. But for tests we don't need it yet, that just requires a webhook.

PR #317, opened 2020-03-10
tasks: Limit systemd task containers to 16 GiB RAM

martinpitt:
This prevents broken processes (such as firefox on some of our CI
machines) to block other tests and take unnecessarily long to die.


martinpitt:
I rolled this out onto our e2e machines and so far it works fine.

martinpitt:
This is the same memory limit that we apply in our k8s resources.

PR #318, opened 2020-03-19
images: Properly serve compressed log files

martinpitt:
Without that, they got served as application/gzip and thus could only be
downloaded. Like that they can be watched inline in the browser
conveniently.


martinpitt:
I rolled out that new container, you can test it on https://logs.cockpit-project.org/logs/pull-13753-20200318-211707-f19f90d1-debian-testing/log.html . Compressed .log.gz journal links should work fine.

PR #320, opened 2020-03-29
sink: Exit with code 77 on collision

martinpitt:
That's a more distinctive error code than 1, which happens for any
random crash. We want to be able to detect collisions in bots'
publish-wrapper. See https://github.com/cockpit-project/bots/issues/697
for details.

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -24,6 +24,7 @@ from configparser import ConfigParser
 from io import StringIO
 
 MAX_BUFF_SIZE = sys.maxsize
+EXIT_COLLISION = 77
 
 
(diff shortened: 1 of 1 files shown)


martinpitt:
I rolled this out to our AWS sink. The next emails we get about tracebacks should contain the 77 exit code.

martinpitt:
@marusak: no convention, really; in tests, "77" is a convention for "skip" (we use that as well), and it's kinda-sorta like "skip this test run". But really there isn't any deeper meaning behind this than "sufficiently different from 1" :-)

PR #324, opened 2020-04-20
tasks: Use a different Firefox 69 build

martinpitt:
The previous one does not exist any more.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -7,7 +7,7 @@ RUN dnf -y update && \
         'dnf-command(builddep)' \
         american-fuzzy-lop \
         chromium-headless \
-        https://kojipkgs.fedoraproject.org//packages/firefox/69.0.3/2.fc30/x86_64/firefox-69.0.3-2.fc30.x86_64.rpm \
+        https://kojipkgs.fedoraproject.org//packages/firefox/69.0.3/2.module_f30+6823+393aa9e5/x86_64/firefox-69.0.3-2.module_f30+6823+393aa9e5.x86_64.rpm \
         curl \
         expect \
         gcc \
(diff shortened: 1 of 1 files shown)


martinpitt:
Container rebuilt, pushed, rolled out. Need to wait for a few successful results.

martinpitt:
Meh, that apparently [did not work](https://logs.cockpit-project.org/logs/pull-771-20200420-090245-c523f840-fedora-31-firefox-cockpit-project-cockpit/log.html#2-2)

martinpitt:
The fc32 build doesn't work on F30, fails with an unresolved symbol. So I added an extra symlink into $PATH. Rebuilt, rolled out, tasks restarted.

martinpitt:
[cockpit tests](https://logs.cockpit-project.org/logs/pull-771-20200420-094909-c523f840-fedora-31-firefox-cockpit-project-cockpit/log.html) now work, and [composer/firefox tests](https://logs.cockpit-project.org/logs/pull-771-20200420-094927-c523f840-fedora-31-firefox-osbuild-cockpit-composer/log.html) look generally okay as well.

marusak:
> composer/firefox tests look generally okay as well.

but that is selenium test, so it does not use this firefox browser.

martinpitt:
Darn -- the new container subtly changed something in locale configuration -- e. g. [this](https://logs.cockpit-project.org/logs/pull-13927-20200420-102833-880df459-debian-stable/log.html) or [this](https://logs.cockpit-project.org/logs/pull-13921-20200420-103756-9ef83fe0-debian-stable/log.html) log. The debian images are known to not have any non-C locales. It's not obvious to me why that happens, but I think we can just fix it in cockpit's default expected messages.

PR #326, opened 2020-05-13
tasks: Use /tmp tmpfs if host has enough memory

martinpitt:
On hosts with a lot of parallel containers/tests, the main bottleneck is
I/O contention. At least our e2e hosts we have tons of unused RAM, but
slow disks, so let's make use of the former. This mainly benefits the VM
overlays, which live in /tmp/run/.

If the host has more RAM available than what (n+1) docker containers
need, mount a 10GiB tmpfs into /tmp. A typical container needs ~ 2 GB,
but do limit it in case there is a leak somewhere.

----

I rolled this out into e2e, and will monitor some tests to see how it is going.


martinpitt:
I added some more cleanups and also applied the tmpfs thing to CentOS CI bots.

martinpitt:
This seems to work rather well -- almost lal failures I've looked at today are "honest" bugs, not timeouts. So good to review/land.

PR #329, opened 2020-06-04
tasks: Update to Firefox nightly builds, prepare for $TEST_DATA elimination

martinpitt:
This will finally get us off the old version 69.
https://bugzilla.mozilla.org/show_bug.cgi?id=1593226 was fixed, so
versions ≥ 77 work again for testing Cockpit.

Unfortunately neither upstream beta nor Fedora downstream builds contain
the remote debugger any more, thus we are forced to use nightlies now.

 - [x] Adjust cockpit tests: https://github.com/cockpit-project/cockpit/pull/14192
 - [x] Fix cockpit-composer for new firefox: https://github.com/osbuild/cockpit-composer/pull/988
 - [x] Fix remaining composer failure: https://github.com/osbuild/cockpit-composer/pull/990
 - [x] Fix most of cockpit for new firefox: https://github.com/cockpit-project/cockpit/pull/14199

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -1,14 +1,13 @@
 FROM fedora:32
 LABEL maintainer='cockpit-devel@lists.fedorahosted.org'
 
-# HACK: Install Firefox 69, since 70 has broken cdp, see https://bugzilla.mozilla.org/show_bug.cgi?id=1593226
 RUN dnf -y update && \
     dnf -y install \
         'dnf-command(builddep)' \
         american-fuzzy-lop \
         chromium-headless \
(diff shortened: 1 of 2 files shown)


martinpitt:
Firefox needs another package (libXt) installed, fixed. I also added the minimal necessary changes to support https://github.com/cockpit-project/bots/pull/943. I'd land this and that PR first, and will then send a follow-up which cleans up the tasks container.

martinpitt:
https://github.com/cockpit-project/bots/pull/943 has some initial results, and they don't look good for firefox: [composer](https://logs.cockpit-project.org/logs/pull-943-20200604-113520-281416b3-fedora-31-firefox-osbuild-cockpit-composer/log.html) is completely broken, but that's presumably because it checks out a stable cockpit test/common. I'll send a PR to check out the fixed one.

In [cockpit](https://logs.cockpit-project.org/logs/pull-943-20200604-113520-281416b3-fedora-31-firefox-cockpit-project-cockpit/log.html) tests, it looks like eval_js() is broken somehow.

PR #330, opened 2020-06-05
tasks: Various cleanups

martinpitt:
- [x] Validate that tests and image refresh works with new container: https://github.com/cockpit-project/bots/pull/952

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -84,4 +84,4 @@ CMD ["/usr/local/bin/cockpit-tasks", "--publish", "$TEST_PUBLISH", "--verbose"]
 LABEL INSTALL /usr/bin/docker run -ti --rm --privileged --volume=/:/host:rw --user=root IMAGE /bin/bash -c \"/usr/sbin/chroot /host /bin/sh -s < /usr/local/bin/install-service\"
 
 # Run a simple interactive instance of the tests container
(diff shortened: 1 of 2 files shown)


tasks/Dockerfile
```diff
@@ -68,26 +64,24 @@ RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user && \
...
-    chmod -R ugo+w /cache /build /secrets /cache /home/user && \
+    chmod -R ugo+w /cache /secrets /cache /home/user && \
     ln -s /app/bin/firefox /usr/bin/firefox
 
 # Prevent us from screwing around with KVM settings in the container
 RUN touch /etc/modprobe.d/kvm-amd.conf && touch /etc/modprobe.d/kvm-intel.conf
 
 ENV LANG=C.UTF-8 \
     LIBGUESTFS_BACKEND=direct \
-    XDG_CACHE_HOME=/build \
```
martinpitt, line 79: Hmm, this works, but it's a bit odd -- now all the builds literally happen in ~user/.cache/cockpit-project/.. , which is hard to see. I'll think about  how to move them to ~user cleanly.

martinpitt: Fixed in another commit.

martinpitt:
Tests look fine, and the task containers now look natural -- you ssh in, you are in your home dir, and the checkouts are right there.

PR #333, opened 2020-06-15
release: Quiesce make

martinpitt:
Don't run `make` with `V=1`. There are no surprises in the build log
during a release, and no unit test runs either.

This reduces the release log by ~ 90%.

### DIFF
--- a/release/release-source
+++ b/release/release-source
@@ -37,7 +37,7 @@ VERBOSE=${RELEASE_VERBOSE:-0}
 SOURCE=${RELEASE_SOURCE:-}
 TAG=${RELEASE_TAG:-}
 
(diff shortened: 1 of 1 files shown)


martinpitt:
I am just running a cockpit release (which is where this finally annoyed me enough). But let's first do a new one before landing this. I'm building/uploading the release container now. This also updates it from Fedora 30 to 31, so let's be extra careful.

martinpitt:
Release 222 was done  successfully with this, and the [log](https://logs-https-cockpit.apps.ci.centos.org/logs/release-cockpit-222/log) shrank from 11 MB to 170 kB. Yay!

PR #335, opened 2020-07-29
Batches and more jobs

marusak:
Firstly lets see if 5 jobs won't make things too brittle.

### DIFF
--- a/tasks/cockpit-tasks-centosci.yaml
+++ b/tasks/cockpit-tasks-centosci.yaml
@@ -17,7 +17,7 @@ spec:
         image: docker.io/cockpit/tasks
         env:
         - name: TEST_JOBS
-          value: '4'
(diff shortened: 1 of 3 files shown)


tasks/cockpit-tasks-centosci.yaml
```diff
@@ -18,6 +18,8 @@ spec:
         env:
         - name: TEST_JOBS
           value: '5'
+        - name: TEST_BATCHES
+          value: '2'
```
martinpitt, line 22: See discussion in the cockpit PR -- I'd rather provide a sane default which ends up as "2" with TEST_JOBS==4 or 5.

marusak: Agreed, dropped the second commit.

PR #336, opened 2020-08-05
tasks: Document how to integrate a new project into CI

martinpitt:
Merely setting up the webhook is not sufficient. You also need to add
cockpituous as a collaborator and set up the bots test map.

### DIFF
--- a/tasks/README.md
+++ b/tasks/README.md
@@ -162,6 +162,17 @@ Add a webhook to your GitHub project on the Settings → Webhooks page of your p
 See [GitHub development documentation](https://developer.github.com/webhooks/)
 for more information.
 
+### Set up bot as collaborator
+
+ * On your project's "Settings → Manage Access" page, invite the [cockpituous user](https://github.com/cockpituous) as a collaborator.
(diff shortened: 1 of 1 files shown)


martinpitt:
I  just went through this exercise on https://github.com/martinpitt/performance-graphs , and these steps worked: https://github.com/martinpitt/performance-graphs/pull/2

tasks/README.md
```diff
@@ -162,6 +162,17 @@ Add a webhook to your GitHub project on the Settings → Webhooks page of your p
 See [GitHub development documentation](https://developer.github.com/webhooks/)
 for more information.
 
+### Set up bot as collaborator
+
+ * On your project's "Settings → Manage Access" page, invite the [cockpituous user](https://github.com/cockpituous) as a collaborator.
+ * Once you see the "Pending invite", in the list, the clipboard symbol copies the corresponding `/invitations` URL to the clipboard. Send that to Martin Pitt (`pitti` in `#cockpit` on FreeNode IRC) to log into GitHub as cockpituous user and accept the invite.
```
Gundersanne, line 168: The way the last sentence here is phrased makes it sound like the reader will log into GH as cockpituous. Wonder if something like 

`Send that to Martin Pitt (...), who will accept the invite as cockpituous.`

marusak: Maybe this would be worth to make the bus factor a bit bigger :)

martinpitt: Good points you both. I added Marius, and fixed the sentence structure. I think Lars can do that as well, but he's on PTO so can't confirm right now.

PR #337, opened 2020-08-10
sink: Exit 77 on collision

Gundersanne:
- [x] deploy this first and see how it runs

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -116,6 +116,12 @@ def contains(expect, result, debug=True):
         return True
 
 
+def github_usr1_handler(sig, frame):
+    stderr.write(b"State not as expected. Possible collision. Aborting.\n")
+    stderr.flush()
+    sys.exit(EXIT_COLLISION)
+
+
 class GitHub(object):
     def __init__(self, config):
         self.token = None
@@ -130,6 +136,7 @@ class GitHub(object):
                 raise
         self.results = {}
         self.child = 0
+        signal.signal(signal.SIGUSR1, github_usr1_handler)
 
     def __del__(self):
         self.cleanup()
@@ -206,10 +213,9 @@ class GitHub(object):
                 ret = 0
                 try:
                     if not self.watches(token, watches):
-                        os.kill(os.getppid(), signal.SIGTERM)
-                        stderr.write(b"State not as expected. Possible collision. Aborting.\n")
-                        stderr.flush()
-                        ret = EXIT_COLLISION
+                        # signal to the parent process to exit with code 77
+                        os.kill(os.getppid(), signal.SIGUSR1)
+                        ret = 1
                 except:
                     traceback.print_exc()
                 os._exit(ret)
@@ -339,7 +345,7 @@ class Status(object):
         for reporter in self.reporters:
             try:
                 reporter.push(data)
-            except:
+            except Exception:
                 traceback.print_exc()
                 log.write(traceback.format_exc().encode('UTF-8'))
         self.data = data
--- a/sink/test-sink
+++ b/sink/test-sink
@@ -155,7 +155,7 @@ class TestGitHub(unittest.TestCase, ):
         # Should terminate by itself due to /expect
         ret = sink.wait()
         sink.stdin.close()
-        self.assertEqual(ret, -signal.SIGTERM)
+        self.assertEqual(ret, 77)
 
 
 if __name__ == '__main__':


allisonkarlitskaya:
The test is failing...

sink/sink
```diff
@@ -116,6 +116,10 @@ def contains(expect, result, debug=True):
         return True
 
 
+def github_usr1_handler(sig, frame):
+    raise RuntimeError("Sink main thread received sigusr1 signal")
```
allisonkarlitskaya, line 120: Wording nag: you're dealing with processes, not threads  (please also change this in the commit message).

Also: why not create a custom exception type for this, or just sys.exit() on the spot?  sys.exit() is, itself, equivalent to just throwing a `SystemExit` exception.

Specifically, I'm worried that your `except` below might catch other `RuntimeError`s than the ones thrown from here.

sink/sink
```diff
@@ -202,14 +206,14 @@ class GitHub(object):
         watches = (github.get("watches") or []) + (http.get("watches") or [])
         if watches:
             self.child = os.fork()
-            if self.child == 0:
+            if self.child > 0:
+                signal.signal(signal.SIGUSR1, github_usr1_handler)
```
allisonkarlitskaya, line 210: There's a race here.  The child might raise the signal before the parent has its handler setup.  Either use some form of synchronisation to avoid that, or set the handler up before the `fork()`.

In fact, it makes sense to set up the handler exactly once, at the start, and not at each `fork()`.

sink/test-sink
```diff
@@ -155,7 +155,7 @@ class TestGitHub(unittest.TestCase, ):
         # Should terminate by itself due to /expect
         ret = sink.wait()
         sink.stdin.close()
-        self.assertEqual(ret, -signal.SIGTERM)
+        self.assertEqual(ret, 1)
```
allisonkarlitskaya, line 158: ...where do we test that we get `77`?

sink/sink
Gundersanne: if I throw an exception straight from the handler it exits really weirdly. I could create a custom exception for this though.

sink/sink
Gundersanne: I'll set the handler up before the fork :)

sink/test-sink
Gundersanne: we don't, the tests here don't test collisions and i'm not sure how

sink/sink
allisonkarlitskaya: you *are* throwing an exception straight from the handler...

Gundersanne: to clarify this slightly: there's blanket try/excepts around the place which catch the `SystemExit` thrown from the handler, and either we have to catch `SystemExit` again and call `sys.exit` again from the `except`, or we have the custom exception.

sink/sink
```diff
@@ -116,6 +116,14 @@ def contains(expect, result, debug=True):
         return True
 
 
+class GithubPushError(Exception):
+    pass
+
+
+def github_usr1_handler(sig, frame):
+    raise GithubPushError("Sink main thread received sigusr1 signal")
```
allisonkarlitskaya, line 124: you still have "thread" here.

allisonkarlitskaya, line 124: you still have "thread" here and in the commit message.

sink/sink
```diff
@@ -228,10 +235,13 @@ class Irc(object):
         self.pipe = None
 
     def __del__(self):
-        if self.pipe is not None:
-            self.pipe.close()
-        if self.pid:
-            os.waitpid(self.pid, 0)
+        try:
```
allisonkarlitskaya, line 238: This seems really suspicious!!

sink/sink
```diff
@@ -339,6 +349,10 @@ class Status(object):
         for reporter in self.reporters:
             try:
                 reporter.push(data)
+            except GithubPushError:
```
allisonkarlitskaya, line 352: I don't like this at all.  This exception could now be fired at any time at any point from within execution of the program.  You therefore need to be able to catch it from any context that the program might find itself in from the instant the signal handler has been installed.

Probably that should even include the call to signal.signal() itself, since it's even possible that the exception could be thrown before that call returns.

This is part of why I was trying to push you toward calling sys.exit() from the handler directly...

sink/sink
```diff
@@ -339,6 +349,10 @@ class Status(object):
...
+                stderr.write(b"State not as expected. Possible collision. Aborting.\n")
+                stderr.flush()
+                sys.exit(EXIT_COLLISION)
             except:
```
allisonkarlitskaya, line 356: You can change this to `except Exception:`.  `SystemExit` derives from `BaseException`, but not `Exception`, and is the only class in the standard library for which this is true.  That will allow the exit to pass through, but will still catch all of the "real" exceptions.

sink/sink
Gundersanne: yea this was a mistake

sink/sink
```diff
@@ -206,10 +213,8 @@ class GitHub(object):
                 ret = 0
                 try:
                     if not self.watches(token, watches):
-                        os.kill(os.getppid(), signal.SIGTERM)
-                        stderr.write(b"State not as expected. Possible collision. Aborting.\n")
-                        stderr.flush()
-                        ret = EXIT_COLLISION
+                        os.kill(os.getppid(), signal.SIGUSR1)
```
allisonkarlitskaya, line 216: A comment here, perhaps.  "Signal the parent process to exit with 77" or so?

marusak:
Thanks @Gundersanne, can you please deploy it and report back if and how it performs? Keeping blocked until then.

Gundersanne:
@marusak @allisonkarlitskaya so it's running on https://github.com/cockpit-project/bots/pull/1141, it's hard to force an actual collision, and the collision won't look differently until that same bots PR is merged. Just wanted to see if it didn't crash :p

PR #338, opened 2020-08-18
release-bodhi: Use `--autotime` option

marusak:
Otherwise updates are not directly pushed to stable after 7 days.

This used to work but is not for at least the last month. No idea though when or why it changed, maybe it is a bug-report candidate?
But in any case `--autotime` cannot hurt.


marusak:
Ok, I'll build the release container and tomorrow verify it worked. Setting to `blocked` in the meantime.

marusak:
Worked in https://bodhi.fedoraproject.org/updates/FEDORA-2020-2b6de47fe5, now autotime is enabled

PR #339, opened 2020-09-15
Specify docker.io registry on container image pulls

martinpitt:
This avoids circling through all available registries, and prevents
accidentally using a wrong image.

Also drop the duplicated non-locked docker pull in
tasks/install-service.


martinpitt:
I deployed this change on e2e, and it works fine.

marusak (approved):
Thanks!

PR #340, opened 2020-09-30
tasks: Bump container RAM

martinpitt:
With the current combination of chromium, `$TEST_JOBS`, and cockpit
tests, the tests ran into the oom-killer a lot with 16 GiB, causing a
lot of "weird" and unspecific test failures. Bump to 24 GiB.

### DIFF
--- a/tasks/cockpit-tasks-centosci.yaml
+++ b/tasks/cockpit-tasks-centosci.yaml
@@ -41,7 +41,7 @@ spec:
           mountPath: "/tmp"
         resources:
           limits:
-            memory: 16G
+            memory: 24G
       volumes:
       - name: secrets
         secret:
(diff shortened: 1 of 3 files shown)


martinpitt:
I already deployed this on e2e yesterday, and since then we didn't have new OOMs.

    ansible -i inventory -m shell -a 'dmesg | grep oom-killer | wc -l' e2e

I just deployed it on CentosCI as well. Although that's rather wonky right now due to https://bugs.centos.org/view.php?id=17762

Gundersanne (approved):
lgtm, we might need to run less of them now though :/

martinpitt:
@Gundersanne: e2e machines have enough RAM. Not sure about CentOS CI, but we can leave that to kubernetes -- it should know the nodes' RAM, and knows the container's RAM requirement.

PR #341, opened 2020-10-05
WIP Aws-lambda

Gundersanne:
Just exploring what the (re)structure should look like, draft PR to track this. Things to figure out:

- [ ] make sure the OS container still works
- [ ] figure out how to store the secrets (amqp things, github token)
- [ ] do we still need /tmp/home ? (regardless aws lambdas have a /tmp scratch space)
- [ ] this PR focuses mostly on the webhook itself, we'd still need to run an amqp instance somewhere (aws also has AmazonMQ which implements amqp as well, so this could work for the aws lambda)
- [ ] packaging for aws (zip file)


martinpitt:
This will be unblocked by moving releases to GitHub actions, PR #343 and PR #344. I.e. after all that landed, we don't need the "create release job" thing any more which is k8s/OpenShift specific.

PR #343, opened 2020-10-10
release container updates

martinpitt:
These make the container easier to use with local podman, in GitHub actions, and GitLab pipelines.

It also updates the container to Fedora 32 (implicitly, as we use `fedora:latest`).

### DIFF
--- a/release/Dockerfile
+++ b/release/Dockerfile
@@ -3,7 +3,6 @@ LABEL maintainer='cockpit-devel@lists.fedorahosted.org'
 
 # whois-mkpasswd conflicts with expect and we don't need it
 RUN dnf -y update && \
-    dnf -y remove whois-mkpasswd && \
     dnf -y install \
 bind-utils \
 bodhi-client \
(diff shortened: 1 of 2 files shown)


martinpitt:
Built and pushed to https://hub.docker.com/r/cockpit/release/tags . Let's block this until we have a successful release of cockpit next Wednesday. @Gundersanne, if it fails, let's just revert back to the previous tag.

I am currently working on moving the entire release process to GitHub actions, and these cleanups fell out of various hacks that I had to make to the action workflow to work with the current container.

Gundersanne:
Tangentially related: I'm sure the release container won't be affected, but the latest cockpit-tasks container broke the verify tests due to a chromium update. Currently trying to fix the test in question. But as the release doesn't need chromium, it's probably going to be fine.

marusak (approved):
Seems fine, thanks!
I understand why you added separate RUN in the last commit, but it creates additional layer. Not a nack though, just saying.

martinpitt:
@marusak: It was deliberate, and that layer is very thin. I made up for it by dropping the `ADD` one. If you prefer, I can re-merge it, but I think it's a bit nicer to develop that way, and it doesn't really hurt (unlike two dnf layers)

marusak:
>  It was deliberate, and that layer is very thin. I made up for it by dropping the ADD one. If you prefer, I can re-merge it, but I think it's a bit nicer to develop that way, and it doesn't really hurt (unlike two dnf layers)

sure, I was just saying :) I am fine with it, no need to merge it.

martinpitt:
Today's cockpit and cockpit-podman releases were successful, so landing.

martinpitt:
Argh, forgot Closes #, pushed.

PR #344, opened 2020-10-14
tasks: Drop release from webhook

martinpitt:
Releases are now done through GitHub actions for each project.

-----

After that, the webhook by and large just feeds the JSON payloads into an AMQP queue, so it's very generic. We could replace it with an AWS lambda thing or something else which is "not our code". The only cockpit specific thing that it does now is to "cold-boot" the queues, i.e. call {tests,issue}-scan at the beginning. If our webhook replacement is going to be basically "always there", we could dispense of it.

But a better idea woud be to try and teach this to the tasks containers: they could check if AMQP queues don't exist yet, and if not, initialize them and call -scan at the beginning. @Gundersanne , WDYT?

 - [x] Ensure that today's cockpit/cockpit-podman releases work
 - [x] Land PR #343

### DIFF
--- a/tasks/webhook
+++ b/tasks/webhook
@@ -23,35 +23,6 @@ SINK = os.getenv('RELEASE_SINK', 'sink-local')
 BOTS_CHECKOUT = HOME_DIR + '/bots'
 sys.path.append(BOTS_CHECKOUT)
 
-# Kubernetes Job template for actually running a release
-JOB = '''---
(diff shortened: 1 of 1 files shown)


martinpitt:
Unblocked. @Gundersanne , I'll leave it to you whether we want to land this, I lost track a bit.

Gundersanne (approved):

PR #345, opened 2020-10-16
release, tasks: Install sassc

martinpitt:
The node-sass NPM package contains/builds ELF binaries, which is
problematic for some distributions. Let's move from node-sass to
sassc, which is packaged in all major distributions.

Install sassc into the tasks container so that packages can start using
it.

### DIFF
--- a/release/Dockerfile
+++ b/release/Dockerfile
@@ -28,6 +28,7 @@ npm \
 psmisc \
 rpm-build \
 rsync \
+sassc \
 tar \
 which \
 dnf-utils \
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -36,6 +36,7 @@ RUN dnf -y update && \
         rpm-build \
         rpmdevtools \
         rsync \
+        sassc \
         sed \
         tar \
         virt-install \


martinpitt:
This recently came up on the [mailing list](https://lists.fedorahosted.org/archives/list/cockpit-devel@lists.fedorahosted.org/thread/UJKOSUUKYBIY6QZ6IL5P455E63XYIAW5/), from @lnussel . After that lands, I'll try to move starter-kit to node-sass, then cockpit and the others. I first want to experiment a bit whether there is any difference in generated css.

martinpitt:
We need it on the release container as well.

I built, pushed, and deployed both containers. Will report some successful tasks container results here.

marusak (approved):
Lets check that chrome and firefox both work. (there are currently two broken firefox tests that I'll look at soon - https://github.com/cockpit-project/cockpit/issues/14720#issuecomment-707701919)

martinpitt:
All tests on https://github.com/cockpit-project/cockpit/pull/14760 and many tests on https://github.com/cockpit-project/cockpit/pull/14740 run against the new container. Failures seem fairly comparable to the previous container, e.g. [firefox](https://logs.cockpit-project.org/logs/pull-14740-20201016-192256-bb6ef843-fedora-32-firefox/log.html)

martinpitt:
https://github.com/cockpit-project/cockpit/pull/14761 was also comparably successful to the previous container.

PR #346, opened 2020-10-21
tasks: Fix make-checkout's --reference-if-able

Gundersanne:
This was removed in #329 but I'm not sure if it was intentional. If it was then let's drop the `--reference-if-able` in make-checkout instead.

- [ ] deploy new tasks

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -41,6 +41,7 @@ fi
 
 # let's just do our work in the current directory
 WORKDIR="$PWD"
+XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
 
 # Set up github user and token
 git config --global credential.helper store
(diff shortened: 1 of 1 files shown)


martinpitt:
I'm not sure how #329 relates to this, but I did not intentionally break the cache dir. Thanks for fixing this!

martinpitt:
@Gundersanne: I'm not sure if you rebuilt the container with this branch, or someone else -- but the 2020-10-23 one was [broken](https://github.com/cockpit-project/starter-kit/pull/386) -- it was missing sassc. I rebuilt 2020-10-28 from current cockpituous master to bring it back. Possibly you need to rebase your branch before rebuilding the container. So I'm afraid this needs to be re-rolled out.

Gundersanne:
@martinpitt i'm not even sure if this makes sense anymore. Because the way it cached the projects for make-checkout was through update_repo. So unless this reintroduces a list which calls update_repo on each of the items this is kinda pointless.

martinpitt:
As discussed on IRC, let's just ditch that then. With `.tasks` support gone, there is much less polling going on, and instead let's just make this as simple as possible.

PR #348, opened 2020-10-23
Ocp4

Gundersanne:


sink/sink-centosci.yaml
```diff
@@ -117,6 +117,6 @@ items:
   data:
     sink: |
         [Sink]
-        Url: https://logs-https-cockpit.apps.ci.centos.org/logs/%(identifier)s/
+        Url: https://logs-https-cockpit.apps.ocp.ci.centos.org/logs/%(identifier)s/
```
martinpitt, line 120: This needs to be logs-https-**frontdoor**.apps.ocp..

martinpitt (changes requested):
Thanks for figuring out the /dev/kvm trick! Please add a Closes # as well. It looks like this is mostly deployed already? I see a bunch of tasks containers with /dev/kvm, and e.g. [this test](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-14391-20201023-173829-efc52d71-debian-testing/log.html) ran successfully.

martinpitt:
Should we try changing the webhook of e.g. starter-kit to OCP4 and do a full test run? Or did you do that already?

martinpitt:
At least cockpit-podman already uses the new webhook, and thus generates wrong log links. See https://github.com/cockpit-project/cockpit-podman/pull/577 for an example.

martinpitt (approved):
Thanks! This is still marked as draft, but :+1: anyway.

martinpitt (changes requested):
Ah, still missing Closes #

martinpitt:
We really need to land this, otherwise the webhook keeps crashing on OCP4 due to the hardcoded wrong AMQP server in tasks/webhook. I disabled the repeatedly crashing containers on OCP4, and will rebuild/re-deploy the tasks container now.

martinpitt (approved):
I rebuilt/pushed the container, and rolled it out. Unfortunately OCP4 seems to hang right now, but at least the test in https://github.com/cockpit-project/bots/pull/1351 started before it got aborted due to the OCP outage. That container had been in production for a while and worked (until we rebuilt it), so let's land this now. Thank you!

PR #349, opened 2020-10-26
tasks: Drop .tasks file support

martinpitt:
All projects moved away from .tasks to GitHub workflows.

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -69,51 +69,12 @@ function update_repo() {
     fi
 }
 
-function task() {
-    dir=$1
-
-    (
-        # If the tasks file is executable, run it
-        if [ -x "$dir"/.tasks ]; then
-            cd "$dir/" && ./.tasks
-
-        # If it exists then cat it
-        elif [ -e "$dir/.tasks" ]; then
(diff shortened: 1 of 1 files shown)


martinpitt:
I did *not* build a container with this. I think @Gundersanne wants to land #346 and possibly #341 first, and this is not at all urgent. I just wanted to stash it so that we don't forget to clean it up.

martinpitt:
Unblocking. There has been no movement on the other two PRs, and removing this code may help the others to become easier.

martinpitt:
Oops, putting back blocked -- we should first build/roll out a container before landing this, but I'd like to get a review/ack first.

Gundersanne (approved):
This should work I think

martinpitt:
Just rebased to current master to make container build more convenient/robust.

I pushed and rolled out the container, let's give it some time to prove that it works.

martinpitt:
https://logs.cockpit-project.org/logs/pull-14854-20201103-102218-c0f44c18-fedora-32-firefox/log.html and https://logs.cockpit-project.org/logs/pull-14854-20201103-102205-c0f44c18-fedora-coreos/log.html ran with the new container, and the webhook container works fine.

PR #350, opened 2020-11-02
Fix "latest" tagging of images

martinpitt:
Add the missing registry name, to be consistent with all the other tags
and making the push to dockerhub actually work.


martinpitt:
I'm honestly not sure how that worked before (possibly by "docker.io/" being implied?"), but it did not work on cockpit-7 this morning, and it just looks wrong. The whole point of that tagging is to declare to which registry the container should go.

marusak (approved):
> I'm honestly not sure how that worked before (possibly by "docker.io/" being implied?")

I think so. I don't remember that I had to tag images also with `docker.io` prefix back in the day the way I have to do today with podman.

PR #351, opened 2020-11-05
Add infrastructure Ansible playbooks

martinpitt:
These were previously kept in a Red Hat internal repository, but nothing
in them is private, and they may be useful for other people/teams as
well.

Secrets are kept in a separate Red Hat internal repository.


martinpitt:
I think the most "pesonal" thing here is ansible/roles/users/tasks/ssh-keys . These are public keys, though. If you prefer, I'm happy to move them into the secrets git, and instead we add something to README to document where to locally copy it from.

marusak:
> I think the most "pesonal" thing here is ansible/roles/users/tasks/ssh-keys . These are public keys, though. If you prefer, I'm happy to move them into the secrets git, and instead we add something to README to document where to locally copy it from.

Yeah, I find it a bit weird that they are in git repo (yes yes, I am fully aware they are no secret, but still). Not a nack from my side though. Looked over this and seems fine, I want to diff between this and that internal git repo to see, what was still left behind.

martinpitt:
> what was still left behind.

Nothing really; the public ssh keys would be the only thing. I'm happy to treat them differently though, ideas welcome!

Gundersanne:
I think i'm ok with them being there. The secrets repo would be better but if that's the only thing in there then let's not bother.

martinpitt:
@Gundersanne , @marusak : I soon need to make some changes to the Ansible scripts. Should I rather make them in the private git and re-sync this PR, if you want more time to deliberate this? Or should I stack them up here?

marusak (approved):
Oh sorry for blocking this. This is surely fine - and you and Sanne out-voted me :)

PR #353, opened 2020-12-02
release: Move to GitHub registry, add build workflow

martinpitt:
We only use the container in GitHub workflows now, so it makes sense to
keep the image in the nearby registry. This also avoids running into
pull rate limits on docker.io.

Add a manually triggered "build-release" workflow that will allow us to
conveniently build and push a new image from a branch.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -1,0 +1,20 @@
+name: build-release
+on:
+  # this is meant to be run on an approved PR branch for convenience
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Log into container registry
+        run: podman login -u cockpituous -p ${{ secrets.COCKPITUOUS_GHCR_TOKEN }} ghcr.io
+
+      - name: Build release container
+        run: make release-container
+
+      - name: Push container to registry
+        run: make release-push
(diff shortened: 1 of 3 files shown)


martinpitt:
I [ran this on my fork](https://github.com/martinpitt/cockpituous/actions/runs/395458561) (with temporarily copied secret), it succeeded, and [published the image](https://github.com/orgs/cockpit-project/packages/container/package/release). I can locally pull and run it:
```
❱❱❱ podman run -it --rm ghcr.io/cockpit-project/release /nonexisting
Trying to pull ghcr.io/cockpit-project/release...
[...]
release-runner: No release tag set
```

martinpitt:
Once that lands, I'll switch cockpit and the other projects to use that container in their workflows.

martinpitt:
I just realized that we need this even more urgently for the tasks container -- our e2e bots (and possibly our centosci ones) won't be able to pull it any more from docker.io. So once this lands, I'll do that. (I have to do this on my fork's master branch to run workflows)

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,20 @@
+name: build-release
+on:
+  # this is meant to be run on an approved PR branch for convenience
+  workflow_dispatch:
```
marusak, line 4: That branch would have to be a cockpit one, so not a standard PR.

marusak (approved):
Looks good to me, thanks!

Just to make sure, we can run this only when we create PR from branch in cockpitouos, right? Or from our own, if we set up `COCKPITUOUS_GHCR_TOKEN` secrete in our own forks.

martinpitt:
@marusak: Correct, you have to push branches to the origin repository for running that. Installing the secret into your fork works too, but is more involved and prone to forgetting about it. (I removed the secret from my fork again).

PR #356, opened 2020-12-18
tasks: Add packages and configuration for interactive development

martinpitt:
With these, developers can run a cockpit/tasks container pretty much
anywhere to do Cockpit development. This works well on our e2e machines
or for toolbox. Further packages can always be installed with
`docker exec -it -u root`, but these provide the most important tools.

Fix ~user home directory permissions, so that tmux and ssh work.

Configure kerberos to use a file ccache, as nothing else (reliably)
works in a container.

A comfortable and recommended workflow is to start the container in the
background and (re)attach to it through byobu. Add a `DEV` run label
with a comment to show how this works. Do *not* pass in the production
cockpit secrets, developers ought to use their own token in such a
situation.

-----

During the week when I had no laptop, I had used https://github.com/martinpitt/fedora-dev on an e2e machine. This is 95% what cockpit/tasks is, so let's unify this. I know @allisonkarlitskaya has also used that during her laptop breakage, and in general it seems useful for us to occasionally do some development/debugging on these super-fast machines.

The main thing that is missing from my fedora-dev container is the ssh config and the [setup-cockpit](https://github.com/martinpitt/fedora-dev/blob/master/setup-cockpit) script, but that's easy enough to pull in through  curl -- in retrospect, this should not have been in the container in the first place.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -80,4 +80,4 @@ CMD ["/usr/local/bin/cockpit-tasks", "--publish", "$TEST_PUBLISH", "--verbose"]
 LABEL INSTALL /usr/bin/docker run -ti --rm --privileged --volume=/:/host:rw --user=root IMAGE /bin/bash -c \"/usr/sbin/chroot /host /bin/sh -s < /usr/local/bin/install-service\"
 
 # Run a simple interactive instance of the tests container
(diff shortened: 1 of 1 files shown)


Gundersanne (approved):
Guess we also better wait for an ack from @allisonkarlitskaya since I don't use toolbox.

My only complaint is that it doesn't have emacs installed :)

martinpitt:
It does not quite work with toolbox yet anyway, this is mostly about unifying the bots container with the opportunistic "need to debug something on a powerful remote machine" case where I used my fedora-dev. (Not sure what Lis used).

martinpitt:
I fixed the tasks container for using with toolbox now. As I want to use our fancy new "rebuild tasks container" workflow, I'll close this PR and replace it with one from origin instead of my fork.

PR #357, opened 2020-12-25
tasks: Add packages and configuration for interactive development, fix for toolbox

martinpitt:
With these, developers can run a cockpit/tasks container pretty much
anywhere to do Cockpit development. This works well on our e2e machines
or for toolbox. Further packages can always be installed with
`docker exec -it -u root`, but these provide the most important tools.

Fix ~user home directory permissions, so that tmux and ssh work.

Configure kerberos to use a file ccache, as nothing else (reliably)
works in a container.

A comfortable and recommended workflow is to start the container in the
background and (re)attach to it through byobu. Add a `DEV` run label
with a comment to show how this works. Do *not* pass in the production
cockpit secrets, developers ought to use their own token in such a
situation.

-----

This also fixes the tasks container to Just Work™ with toolbox:

        toolbox create --image quay.io/cockpit/tasks -c cockpit
        toolbox enter cockpit

During the week when I had no laptop, I had used https://github.com/martinpitt/fedora-dev on an e2e machine. This is 95% what cockpit/tasks is, so let's unify this. I know @allisonkarlitskaya has also used that during her laptop breakage, and in general it seems useful for us to occasionally do some development/debugging on these super-fast machines.

The main thing that is missing from my fedora-dev container is the ssh config and the [setup-cockpit](https://github.com/martinpitt/fedora-dev/blob/master/setup-cockpit) script, but that's easy enough to pull in through  curl -- in retrospect, this should not have been in the container in the first place.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -80,4 +80,4 @@ CMD ["/usr/local/bin/cockpit-tasks", "--publish", "$TEST_PUBLISH", "--verbose"]
 LABEL INSTALL /usr/bin/docker run -ti --rm --privileged --volume=/:/host:rw --user=root IMAGE /bin/bash -c \"/usr/sbin/chroot /host /bin/sh -s < /usr/local/bin/install-service\"
 
 # Run a simple interactive instance of the tests container
-LABEL RUN /usr/bin/docker run -ti --rm --privileged --uts=host --volume=/var/lib/cockpit-secrets/tasks:/secrets:ro --volume=/var/cache/cockpit-tasks/images:/cache/images:rw IMAGE /bin/bash -i
+LABEL RUN /usr/bin/docker run -ti --rm --volume=/var/lib/cockpit-secrets/tasks:/secrets:ro --volume=/var/cache/cockpit-tasks/images:/cache/images:rw IMAGE /bin/bash -i
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -9,7 +9,6 @@ RUN dnf -y update && \
         curl \
         dbus-glib \
         expect \
-        gcc \
         gcc-c++ \
         git \
         gnupg \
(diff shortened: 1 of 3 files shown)


martinpitt:
Once more with feeling^Wdocumentation.

martinpitt:
I did a full round of building, image-prepare and testing cockpit in that toolbox, and it's working well. I'll move from my self-built image to that now, and see how it works long-term. Of course it can still be wrapped into some customization scripts, with extra packages that developers want (*cough* emacs, hello @Gundersanne!), but that will be a lot quicker than starting from scratch.

I [rebuilt](https://github.com/cockpit-project/cockpituous/actions/runs/443956315) the [official container](https://quay.io/repository/cockpit/tasks?tab=tags), and rolled it out to our CI. But this failed, it seems it really does need `$HOME`. I'll need to move that to cockpit-tasks.

martinpitt:
Ugh, that also didn't work -- there is a rather strong tension between "real" docker/podman where uid == 1111 and permissions must be correct, and the random UID on OpenShift which then could not access ~/.ssh/ any more. I reverted the stricter permissions of ~/.ssh, that's something which can easily enough be done in the "set up container for cockpit" script.

martinpitt:
This works well enough now. I refreshed the container again, deployed it, and there is [one test on CentOS CI](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15072-20201225-130807-844d042c-fedora-32/log.html) and [one on e2e](https://logs.cockpit-project.org/logs/pull-15072-20201225-130807-844d042c-fedora-33/log.html).

I.e. for CI and toolbox this is perfect now; for an occasional "I need a dev container on an e2e machine" it's good enough also, and just needs a tiny setup script. But I'll submit that separately.

marusak (approved):
The last commit message says `Move user home directory from /user/home to /build.` yet you moved it to `/work`. Just saying, no reason to repush :) 

Seems all well to me, thanks!

martinpitt:
@marusak: Argh, now I have to fix that, and it's also confusing. As the tests take virtually no time, IMHO it's worth having that be correct. No code changes, can you please re-approve?

martinpitt:
Oh, we don't do the "PR updates reset approvals" thing on cockpituous. So landing, but I'll enable this now for this repo.

PR #359, opened 2021-01-19
tasks: Various fixes and improvements; add script to run AMQP+tasks locally

martinpitt:
I would like to make our tasks container self-testable, and run integration tests on tasks container refreshes and changes to the logic eventually. This is as far as I got today. It's not yet self-testing, but at least one can now run it locally in a very easy way. Please test it and let me know how it works for you!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,25 @@
+name: tests
+on: [pull_request, workflow_dispatch]
+jobs:
+  tasks:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          # need this to get origin/master for git diff
+          fetch-depth: 0
+
+      - name: Rebase to current master
+        run: git rebase origin/master
+
+      - name: Build and test tasks container if it changed
+        run: |
+          changes=$(git diff --name-only origin/master..HEAD -- tasks/)
+          if [ -n "${changes}" ]; then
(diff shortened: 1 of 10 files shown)


martinpitt:
I [refreshed the container](https://github.com/cockpit-project/cockpituous/actions/runs/496109775) and deployed it everywhere. Will link to successful test runs.

martinpitt:
[this test](https://logs.cockpit-project.org/logs/pull-15154-20210119-140037-ddb634c7-ubuntu-stable/log.html) ran with the new container, and looks reasonable.

martinpitt:
I'm experimenting with running this new script in a GitHub workflow. This does not change the container content, and thus does not invalidate the previous cockpit integration test results.

martinpitt:
Cool, [works](https://github.com/cockpit-project/cockpituous/actions/runs/496400942)! Pushing again with more sensible naming. I want to also integrate `make check` into this (replacing semaphore) to have all test definitions inside of the archive. But that's for a follow-up.

martinpitt:
I just couldn't resist and added running unit tests here, but surprisingly [it fails on GitHub](https://github.com/cockpit-project/cockpituous/pull/359/checks?check_run_id=1729194055). Our semaphore still uses a rather ancient Ubuntu 16.04. Investigating..

Gundersanne (approved):
Lovely! Tried it locally and honestly this is great. Can't wait for an easy way to produce workloads/post stuff to the queue :)

PR #362, opened 2021-01-20
ansible: Get secrets from git, fix image containers after secrets update

martinpitt:
The tasks/webhook secrets are now in the internal CI secrets git
repository. Change sync-secrets.yml to get them from git, stop
considering cockpit-9 as the "holy master copy", and treat all machines
equally.

----

I ran this on my machine and rolled it out everywhere.

### DIFF
--- a/ansible/inventory/e2e
+++ b/ansible/inventory/e2e
@@ -1,7 +1,3 @@
-# master copy of secrets
-[master]
-cockpit-9.e2e.bos.redhat.com ansible_user=root
-
 [e2e]
 cockpit-7.e2e.bos.redhat.com ansible_user=root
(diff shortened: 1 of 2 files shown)


martinpitt:
Hang on, this fails. We need to turn the ca.pem → ../ca.pem symlinks into file copies, due to the way volumes work.

martinpitt:
This works fine now at last. So much implicit state on our e2e machines! I'm glad that this is all explicit now.

martinpitt:
I completely fixed the deployment now. After yesterday's, image uploads were broken. I tried to fix it in PR #363, but was completely misled about the failure. The second commit here is the real fix, I take the liberty to just add it here as it affects the same script.

Gundersanne (approved):
lgtm :)

PR #363, opened 2021-01-20
images: Give webhook secrets to images container for GitHub token

martinpitt:
We split off the webhook secrets (with the GitHub token) from the tasks
secrets a while ago, and image's Dockerfile links
~user/.config/github-token to /run/secrets/webhook/ (just like the tasks
container).

The images container needs the token for authenticating image uploads.

This does the same as commit 75fc8a3a17226 did for the tasks container.

### DIFF
--- a/images/install-service
+++ b/images/install-service
@@ -3,7 +3,7 @@
 set -eufx
 
 CACHE=/var/cache/cockpit-tasks
-SECRETS=/var/lib/cockpit-secrets/tasks
+SECRETS=/var/lib/cockpit-secrets
 
 if RUNC=$(which podman 2>/dev/null); then
     UNIT_DEPS=''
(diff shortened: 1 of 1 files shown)


martinpitt:
This previously failed [like this](https://logs.cockpit-project.org/logs/image-refresh-1562-20210120-112323/log). I deployed that on cockpit-11 and the AWS image server, and validated that I can do

    ./image-upload --store https://cockpit-11.e2e.bos.redhat.com:8493/ cirros
    ./image-upload --store https://10.29.163.169:8493/ cirros

from a tasks bot again. I can't really explain how it was *not* broken before -- this must have been lingering for over a year already. I just cleaned up the secrets this morning, but that doesn't explain how /run/secrets/webhook/ came into existence in the previous container. So I'm really stunned right now..

marusak (approved):
Seems ok, thanks!

martinpitt:
Hmm -- our CentOS CI container [does not get the webhook secrets either](https://github.com/cockpit-project/cockpituous/blob/master/images/cockpit-images.yaml). I had a thinko here -- that ~/.config file is for the benefit of the sink only, the image server uses /secrets/htaccess.

So the reason was that I updated the secrets this morning, but forgot to restart the image container -- so the old volume went away and nginx did not have a htaccess file any more. A simple container restart was enough.

D'oh! :hankey:

PR #364, opened 2021-01-21
.github: Self-testing tasks workflow

Gundersanne:


.github/workflows/tests.yml
```diff
@@ -27,6 +27,19 @@ jobs:
       - name: Rebase to current master
         run: git rebase origin/master
 
+      - name: Prepare
+        run: |
+          # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
+          export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
+          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
```
martinpitt, line 34: For the record: To use that, you can't run this from your fork. Push your branch to origin, and work from there. We also have to do that when we want to use the manual workflows to rebuild containers. It's totally legit.

Gundersanne:
Reopening from a branch of the repo itself so the workflow has access to secrets.

PR #365, opened 2021-01-22
.github: Self-testing tasks workflow

Gundersanne:


.github/workflows/tests.yml
```diff
@@ -27,13 +27,74 @@ jobs:
       - name: Rebase to current master
         run: git rebase origin/master
 
+      - name: Prepare
+        run: |
+          # # override GitHub's bind mount from host, we don't want anything from there and it interferes with ssh
+          # export HOME=$(getent passwd $(id -u) | cut -f6 -d:)
```
martinpitt, line 33: Is that just some copy&paste leftover? We do use that in cockpit's release workflow, but it seems to work without now?

.github/workflows/tests.yml
```diff
@@ -27,13 +27,74 @@ jobs:
...
+
+          # Put the /secrets in place which we'll mount to the tasks-container
+          mkdir -p ~/secrets
+          echo '${{ secrets.COCKPITUOUS_TOKEN }}' > ~/.config/github-token
+          echo '${{ secrets.SSH_KNOWN_HOSTS }}' > ~/secrets/known_hosts
+          echo '${{ secrets.SINK_SSH_CONFIG }}' > ~/secrets/ssh-config
+          echo '${{ secrets.SINK_ID_RSA }}' > ~/secrets/id_rsa
+          echo '${{ secrets.SINK_ID_RSA_PUB }}' > ~/secrets/id_rsa.pub
```
martinpitt, line 41: Let's hope we don't submit a PR which will make a mess out of our sinks :grin: However, ok for now I think -- I want to include the sink into the local deployment anyway. For the test run we then wouldn't get the `make check` on our public production logs, but that's kind of the point, and they are not interesting anyway. We can instead validate the generated test log from the local sink in the workflow.

But that's just thinking aloud. Good enough for now.

.github/workflows/tests.yml
```diff
@@ -27,13 +27,74 @@ jobs:
...
+          sudo chmod 400 ~/secrets/*
+          # change permissions to user in cockpit-tasks
+          # TODO sudo needed?
```
martinpitt, line 44: Alternatively, make the secrets world readable. The container shouldn't write/change them anyway, and maybe that's a good way to ensure that it won't try to?

.github/workflows/tests.yml
```diff
@@ -27,13 +27,74 @@ jobs:
...
+          sudo chown 1111:1111 ~/secrets/*
+
+          # install dependencies needed to use bots stuff in the host container
+          sudo pip3 install pika
```
martinpitt, line 48: "container" == "GitHub workflow host VM"?

I'd rather we let the webhook container handle the AMQP bits, and only inject HTTP requests to the webhook container. Or perhaps, to work with this version, it can do something like `podman exec -i cockpituous-tasks sh -exc 'cd cockpit-project/bots; ./tests-scan ...`?

.github/workflows/tests.yml
```diff
@@ -27,13 +27,74 @@ jobs:
...
-              # This does not do much yet, just iterate to the AMQP queue 30 times and exit
-              # TODO: Inject a workload that does not need /dev/kvm
-              sudo tasks/run-local.sh
+              # TODO uncomment, make it faster now
+              # sudo make tasks-container
+
+              # Generate the secrets in tree and copy the working github token
+              (cd tasks/credentials; ./generate-ca.sh)
+              (cd tasks/credentials/webhook; ./generate.sh)
+              chmod -R go+rX tasks/credentials/webhook
```
martinpitt, line 65: Wait, what does that do? run-local.sh already generates secrets, why does it need to happen again here? This feels too complex for the workflow file.

.github/workflows/tests.yml
```diff
@@ -27,13 +27,74 @@ jobs:
...
+              cp ~/.config/github-token tasks/credentials/webhook/.config--github-token
+
+              # Start the amqp/tasks pod in the background and supply working task secrets
+              sudo tasks/run-local.sh -d -s ~/secrets
+
+              # clone bots
+              # TODO: figure out how to split this into different steps if possible
+              git clone https://github.com/cockpit-project/bots
+              git -C ./bots fetch origin pull/1567/head
+              git -C ./bots checkout FETCH_HEAD
```
martinpitt, line 75: The tasks container already clones the bots -- I really think it's cleaner and also more useful (as it validates much more actual code) to do this inside the cockpituous-tasks container.

tasks/credentials/webhook/serial
```diff
@@ -0,0 +1 @@
+01
```
martinpitt, line 1: Please don't commit that.

tasks/run-local.sh
```diff
@@ -3,10 +3,36 @@
 # You can run against a tasks container tag different than latest by setting "$TASKS_TAG"
 set -eu
 
+DAEMON=
+TASK_SECRETS=
+
+while getopts "dhs:" opt; do
+    case $opt in
+        h)
+            echo -d runs the cockpit-tasks container as daemon
```
martinpitt, line 12: Fine for me to make that the default mode, it's probably better for local development as well - then you can work with the thing without it getting ripped away under your feet after a minute.

martinpitt (changes requested):
Thanks, this is starting to look promising! I know, still draft, but I had a first rough look (couldn't resist)

.github/workflows/tests.yml
Gundersanne: Can't, ssh-stuff has file permission requirements.

.github/workflows/tests.yml
Gundersanne: yea that's probably better :)

.github/workflows/tests.yml
Gundersanne: Yea I wasn't sure how to solve this. `run-local` won't generate them if it thinks they're already in the tree, and I need them in a predictable location, so I can copy in the webhook token before running the tasks container.

.github/workflows/tests.yml
Gundersanne: yea this was just because I needed to check out the specific bots PR so it would recognize the test it triggers in the testmap.

tasks/credentials/webhook/serial
Gundersanne: woops, sorry sneaked in :)

tasks/run-local.sh
Gundersanne: oh it still gets ripped away, as the tasks container exits after 30 `run-queues`

.github/workflows/tests.yml
martinpitt: Ah! In fact I [have a commit](https://github.com/martinpitt/cockpituous/commit/ccd5639aa7a40bc12df85f6b44aa5bf36541e7e3) to put them into a static path. I just didn't propose it yet as I didn't want to interfere with your work too much. But maybe that one should go in then? (Feel free to just cherry-pick it)

martinpitt:
@Gundersanne : I now have a local sink/image server deployment in my [local-sink branch](https://github.com/martinpitt/cockpituous/commits/local-sink). Now I seriously hit a point where we overlap -- I now need to send it some test request to exercise and log. Maybe this PR would even get easier with that state, as this simplifies the secrets handling.

I think we should stop using the external public sink for this -- for testing changes it seems better to use a local deployment? This will also allow us to automatically *verify* the published test log.

.github/workflows/tests.yml
```diff
@@ -1,5 +1,5 @@
 name: tests
-on: [pull_request, workflow_dispatch]
+on: [pull_request]
```
martinpitt, line 2: Was there a  particular reason to disable it? It seemed useful to be able to trigger it from one's own fork during the development of a new thing, without having to spam a PR with a dozen pushes

.github/workflows/tests.yml
martinpitt: Let's drop this again in a future iteration, I already have a "local sink" branch.

.github/workflows/tests.yml
```diff
@@ -27,13 +27,49 @@ jobs:
...
+              PRN=$(echo "$GITHUB_REF" | cut -f3 -d '/')
+              # wait until bots checkout
+              sudo podman exec -i cockpituous-tasks sh -exc "
+              until [ -f cockpit-project/bots/tests-trigger ]; do
+                  sleep 5;
+              done;
+              cd cockpit-project/bots;
+              ./tests-trigger -f --repo cockpit-project/cockpituous $PRN unit-tests;
+              until (./tests-scan --repo cockpit-project/cockpituous -p $PRN -dv 2>/dev/null | grep -q 'unit-tests'); do
+                 ./tests-scan --repo cockpit-project/cockpituous -p $PRN -dv;
```
martinpitt, line 63: This seems unnecessary? The `until` condition already calls the exact same command. Also, let's introduce a timeout. Maybe something like
```sh
for retry in `seq 10`; do
    OUT=$(tests-scan..)
    [ "${OUT%unit-tests*}" = "$OUT" ] || break
    [... message, sleep.. ]
done
```

Also, nitpick: Not necessary to have the `;` line endings, as these are already in separate lines.

.github/workflows/tests.yml
```diff
@@ -27,13 +27,49 @@ jobs:
...
+                  echo waiting until the status is visible;
+                  sleep 10;
+              done;
+              ./tests-scan -p $PRN --amqp 'localhost:5671' --repo cockpit-project/cockpituous;
+              ./inspect-queue;"
```
martinpitt, line 68: Can we rely on that being *that* synchronous? Then again it's not doing any assertions about it, so that's just for debugging? Should it loop until the test is queued? That may be useful for directly seeing *why* a test is failing (instead of idly looping on an empty queue later)

tasks/run-local.sh
```diff
@@ -3,12 +3,47 @@
...
+DAEMON=
+TASK_SECRETS=
+TOKEN=
+
+while getopts "dhs:t:" opt; do
+    case $opt in
+        h)
+            echo -d runs the cockpit-tasks container as daemon
+            echo -s supply the tasks-secret directory
+            echo -t supply a token which will be copied into the webhook secrets
```
martinpitt, line 15: Better quote these -- `echo` by itself has command line arguments too :)

tasks/run-local.sh
```diff
@@ -46,8 +81,12 @@ else
     chmod -R go+rX "$SECRETS"/webhook
 fi
 
+if [ -n "$TOKEN" ]; then
+    cp -fv "$TOKEN" "$SECRETS"/webhook/.config--github-token
+fi
+
 # start podman and run RabbitMQ in the background
-podman run -d --name cockpituous-rabbitmq --pod=new:cockpituous -v "$RABBITMQ_CONFIG":/etc/rabbitmq:ro -v "$SECRETS"/webhook:/run/secrets/webhook:ro docker.io/rabbitmq:3-management
+podman run -d --name cockpituous-rabbitmq --pod=new:cockpituous -v "$RABBITMQ_CONFIG":/etc/rabbitmq:ro -v "$SECRETS"/webhook:/run/secrets/webhook:ro -p 5671:5671 docker.io/rabbitmq:3-management
```
martinpitt, line 89: Do we still need to expose the queue in that way? It now calls the scans inside the container, and for outside testing I think we really want to end JSON payloads to the webhook http.

martinpitt (changes requested):

.github/workflows/tests.yml
Gundersanne: It's just for debugging, the queuing is synchronous, there's of course a small chance the tasks container will grab it before inspect-queue can grab/release it :)

.github/workflows/tests.yml
Gundersanne: mostly for debugging, i'll remove it. agreed on the timeout

.github/workflows/tests.yml
Gundersanne: Right now it takes the pull number from `GITHUB_REF`, as in this really only works right now with PRs against main.

.github/workflows/tests.yml
```diff
@@ -27,13 +27,51 @@ jobs:
...
+              # wait until bots checkout
+              sudo podman exec -i cockpituous-tasks sh -exc "
+              for retry in `seq 5`; do
+                  [ -f cockpit-project/bots/tests-trigger ] && break
+                  sleep 5;
+              done;
+              cd cockpit-project/bots;
+              ./tests-trigger -f --repo cockpit-project/cockpituous $PRN unit-tests;
+              for retry in `seq 10`; do
+                  OUT=$(./tests-scan --repo cockpit-project/cockpituous -p $PRN -dv)
```
martinpitt, line 64: I suppose the `$` here needs to be quoted.

martinpitt: Pushed a fixup. ... which I just realized is broken because of the $PRN.

martinpitt:
That didn't work. Resetting back to your last version, sorry for meddling.

martinpitt (approved):
Awesome, thank you! This is a nice milestone. I'll then look how to build my "run sink" stuff on top of this.

PR #367, opened 2021-01-26
Fix rebasing in tests workflow

martinpitt:
For an actual rebase to work, git insists on having an user name.

-----

Fixes [this failure](https://github.com/cockpit-project/cockpituous/runs/1769486155?check_suite_focus=true) when branches are actually out of sync with master. I deliberately made this to be an outdated branch, so that we can let this self-test.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,10 @@ jobs:
           fetch-depth: 0
 
       - name: Rebase to current master
-        run: git rebase origin/master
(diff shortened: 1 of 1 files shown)


martinpitt:
[worked](https://github.com/cockpit-project/cockpituous/pull/367/checks?check_run_id=1769534609#step:3:6)

marusak (approved):
Interesting, thanks for taking a look at it!

PR #368, opened 2021-01-27
tasks: Deploy image/sink in run-local.sh

martinpitt:
- Create ssh config/key tasks secrets
 - Expose port 8080 (as "image server") on the pod
 - Start "cockpituous-images" container in the pod, configured like
   sink/sink-centosci.yaml
 - Tell tasks container to use it via `$TEST_PUBLISH`
 - Do some coarse assertions about the sink's generated test logs

 - Revert using the production sink secrets in the tests workflow.
   This will mean the GitHub status URL is http://localhost:8080/...
   which is a bit odd; but that address actually works if you run
   run-local.sh manually locally, and for the workflow run it's not
   really important to have the unit-tests log on the official sink.
   It's much more important to exercise a changed sink interaction.

 - [x] Depends on #370
 - [x] Depends on #371
 - [x] fix ssh failure when running containers in workflow: PR #372 
 - [x] validate that image HTTP server works
 - [x] Update README documentation

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,7 @@ jobs:
 
   tasks:
     runs-on: ubuntu-20.04
+    timeout-minutes: 30
     steps:
       - name: Clone repository
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Gundersanne (approved):
neat!

PR #369, opened 2021-01-27
Clean up certificate SANs

martinpitt:
Remove the triplication of subjectAltName: Make
tasks/credentials/openssl.cnf's server_ca_extensions the single source
of truth, drop the unnecessary SAN from the AMQP client certificate, and
use it in the image certificate script as well.

This also avoids using /etc/pki/tls/openssl.cnf, which does not exist in
Debian/Ubuntu (i.e. on GitHub workflows), and we don't want to rely on
the system config anyway.

### DIFF
--- a/images/generate-image-certs.sh
+++ b/images/generate-image-certs.sh
@@ -1,7 +1,15 @@
 #!/bin/bash -ex
 # Generate image server certificate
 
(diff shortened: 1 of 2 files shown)


martinpitt:
Uh, our new workflow does not work for PRs from forks. Will repost from origin branch.

PR #370, opened 2021-01-27
Clean up certificate SANs

martinpitt:
Remove the triplication of subjectAltName: Make
tasks/credentials/openssl.cnf's server_ca_extensions the single source
of truth, drop the unnecessary SAN from the AMQP client certificate, and
use it in the image certificate script as well.

This also avoids using /etc/pki/tls/openssl.cnf, which does not exist in
Debian/Ubuntu (i.e. on GitHub workflows), and we don't want to rely on
the system config anyway.

### DIFF
--- a/images/generate-image-certs.sh
+++ b/images/generate-image-certs.sh
@@ -1,7 +1,16 @@
-#!/bin/bash -ex
+#!/bin/sh
 # Generate image server certificate
+set -eux
 
-SAN="DNS:*.apps.ci.centos.org,DNS:*.apps.ocp.ci.centos.org,DNS:*.e2e.bos.redhat.com,DNS:*.cockpit-project.org,DNS:cockpit-tests"
-openssl req -new -newkey rsa:2048 -nodes -keyout server.key -out server.csr -subj '/O=Cockpit/OU=Cockpituous/CN=cockpit-tests' -extensions SAN -reqexts SAN -config <(cat /etc/pki/tls/openssl.cnf; printf "\n[SAN]\nsubjectAltName=$SAN")
-openssl x509 -req -days 365000 -in server.csr -CA ../ca.pem -CAkey ../ca.key -set_serial $(date +%s) -extensions SAN -extfile <(printf "\n[SAN]\nsubjectAltName=$SAN") -out server.pem
+ROOTDIR="$(realpath -m "$0"/../../)"
+OPENSSL_CNF="$ROOTDIR/tasks/credentials/openssl.cnf"
+
+openssl req -new -newkey rsa:2048 -nodes -keyout server.key -out server.csr \
+    -subj '/O=Cockpit/OU=Cockpituous/CN=cockpit-tests' \
+    -extensions server_ca_extensions -reqexts server_ca_extensions \
+    -config "$OPENSSL_CNF"
+openssl x509 -req -days 365000 -in server.csr -out server.pem \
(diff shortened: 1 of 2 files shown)


images/generate-image-certs.sh
```diff
@@ -1,7 +1,15 @@
 #!/bin/bash -ex
 # Generate image server certificate
 
-SAN="DNS:*.apps.ci.centos.org,DNS:*.apps.ocp.ci.centos.org,DNS:*.e2e.bos.redhat.com,DNS:*.cockpit-project.org,DNS:cockpit-tests"
-openssl req -new -newkey rsa:2048 -nodes -keyout server.key -out server.csr -subj '/O=Cockpit/OU=Cockpituous/CN=cockpit-tests' -extensions SAN -reqexts SAN -config <(cat /etc/pki/tls/openssl.cnf; printf "\n[SAN]\nsubjectAltName=$SAN")
-openssl x509 -req -days 365000 -in server.csr -CA ../ca.pem -CAkey ../ca.key -set_serial $(date +%s) -extensions SAN -extfile <(printf "\n[SAN]\nsubjectAltName=$SAN") -out server.pem
+ROOTDIR=$(dirname $(dirname $(realpath $0)))
```
allisonkarlitskaya, line 4: I usually write something like

```
top_srcdir="$(realpath -m "$0"/../../..)"
```

in any case, please mind the quoting (here, and on all uses of `$ROOTDIR` and `$OPENSSL_CNF` under).

allisonkarlitskaya (changes requested):
In general, very much in favour of the "DRY" improvements.

images/generate-image-certs.sh
martinpitt: Good point, thanks! Done.

images/generate-image-certs.sh
```diff
@@ -1,7 +1,16 @@
-#!/bin/bash -ex
+#!/bin/sh
 # Generate image server certificate
+set -eux
 
-SAN="DNS:*.apps.ci.centos.org,DNS:*.apps.ocp.ci.centos.org,DNS:*.e2e.bos.redhat.com,DNS:*.cockpit-project.org,DNS:cockpit-tests"
-openssl req -new -newkey rsa:2048 -nodes -keyout server.key -out server.csr -subj '/O=Cockpit/OU=Cockpituous/CN=cockpit-tests' -extensions SAN -reqexts SAN -config <(cat /etc/pki/tls/openssl.cnf; printf "\n[SAN]\nsubjectAltName=$SAN")
-openssl x509 -req -days 365000 -in server.csr -CA ../ca.pem -CAkey ../ca.key -set_serial $(date +%s) -extensions SAN -extfile <(printf "\n[SAN]\nsubjectAltName=$SAN") -out server.pem
+ROOTDIR=$(realpath -m "$0"/../../)
```
allisonkarlitskaya, line 5: quotes here too plz

```suggestion
ROOTDIR="$(realpath -m "$0"/../../)"
```

allisonkarlitskaya (changes requested):

images/generate-image-certs.sh
allisonkarlitskaya: After some reading, it seems that this isn't strictly required, but for consistency and clarity of purpose, I'd include them anyway.

martinpitt: OK, done.

allisonkarlitskaya (approved):
thx <3

PR #371, opened 2021-01-27
self-test code shuffle

martinpitt:
This makes it easier to read the logs, as the lengthy container build is
then not part of the test log any more.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,30 +60,7 @@ jobs:
       - name: Test tasks container if it changed
         if: steps.tasks_changed.outputs.changed
         run: |
-          # Start the amqp/tasks pod in the background and supply working task secrets
-          sudo tasks/run-local.sh -d -s ~/secrets -t ~/.config/github-token
-
-          # trigger a test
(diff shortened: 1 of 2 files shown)


martinpitt:
With just the first commit, [this run](https://github.com/cockpit-project/cockpituous/pull/371/checks?check_run_id=1775331325) correctly skipped all the container build/test steps. I'll now work on a second commit which moves the test logic out of the workflow.

martinpitt:
I added a second commit with some code reorganization, which now does change tasks/, and runs the full glory.

.github/workflows/tests.yml
```diff
@@ -30,7 +30,16 @@ jobs:
...
           git rebase origin/master
 
-      - name: Prepare
+      - name: Check if tasks container changed
+        id: tasks_changed
+        run: |
+          changes=$(git diff --name-only origin/master..HEAD -- tasks/)
+          # print for debugging
+          echo "$changes"
+          [ -z "$changes" ] || echo "::set-output name=changed::true"
```
Gundersanne, line 39: NICE! I was wondering if this was possible :D

Gundersanne (approved):
lgtm !

PR #372, opened 2021-01-28
images: Fix container to run as user, test it in workflow, move to quay.io

martinpitt:
Old container fails like this:
```
❱❱❱ podman run -it --rm -p 8080:8080 -u user -v ./local-data/secrets/tasks:/secrets:ro docker.io/cockpit/images 
nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (13: Permission denied)
2021/01/28 10:11:10 [warn] 1#0: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:1
2021/01/28 10:11:10 [warn] 1#0: could not build optimal types_hash, you should increase either types_hash_max_size: 2048 or types_hash_bucket_size: 64; ignoring types_hash_bucket_size
2021/01/28 10:11:10 [emerg] 1#0: mkdir() "/var/lib/nginx/tmp/client_body" failed (13: Permission denied)
```
This is also what makes the workflow fail in PR #368. In GitHub's version of podman using `--tmpfs ..:mode=777` is broken. But moreover, let's not require this hack in the first place!

With the rebuilt container, this works nicely locally, and you can do http://localhost:8080/ to test nginx.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -1,0 +1,21 @@
+name: build-images
+on:
+  # this is meant to be run on an approved PR branch for convenience
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      # use podman as root, so that it uses overlayfs; user with vfs takes too long and uses too much space
+      - name: Log into container registry
+        run: sudo podman login -u ${{ secrets.QUAY_BOTUSER }} -p ${{ secrets.QUAY_TOKEN }} quay.io
+
+      - name: Build images container
+        run: sudo make images-container
+
+      - name: Push container to registry
+        run: sudo make images-push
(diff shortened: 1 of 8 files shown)


martinpitt:
I created the repo on quay: https://quay.io/repository/cockpit/images?tab=info

Unfortunately I can't yet run the workflow while it's in a PR, but it's identical to the tasks workflow other than the naming.

martinpitt:
Force-pushing to retry failed unit-tests (github.com outage).

I deleted/redeployed sink/sink-centosci.yaml and images/cockpit-images.yaml . https://images-frontdoor.apps.ocp.ci.centos.org/ is happy, and I can enter a centosci tasks bot and do "ssh sink-local-4".

martinpitt:
I re-deployed the images container on cockpit-11 as well. There was a bug in the script wrt. the port (gosh, it must have been ages since we deployed that last). Added another comit to fix that, now http and https both work again as expected.

Gundersanne (approved):

PR #373, opened 2021-02-01
run-local.sh: Test image-{upload,download} and attachments

martinpitt:
Create a GitHub token and corresponding htpasswd; the latter is
statically computed, as we don't want to introduce a `htpasswd` program
dependency, and reimplementing `$apr1` is too cumbersome. For that
reason, delay setting up the real GitHub token until the time when the
test touches the PR.

Test that image uploading and downloading works. This needs to happen in
a copy of the primary bots checkout, as the cockpit-tasks main loop
constantly resets this to origin/master; but we need to modify the bot
checkout's ca.pem to the locally generated one.

### DIFF
--- a/.cockpit-ci/run
+++ b/.cockpit-ci/run
@@ -2,4 +2,7 @@
 
 set -eux
 
+if [ -n "$TEST_ATTACHMENTS" ]; then
+    echo "heisenberg compensator at 99.8% efficiency" > "$TEST_ATTACHMENTS"/bogus.log
+fi
 make check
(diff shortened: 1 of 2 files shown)


martinpitt:
The "copy bots checkout" is a bit meh. I sent https://github.com/cockpit-project/bots/pull/1615 to make it possible to use a different ca.pem without racing against cockpit-tasks.

martinpitt:
Testing attachments was simple, I added that here right away. That's the end of my todo list for local deployment for now.

martinpitt:
https://github.com/cockpit-project/bots/pull/1615 landed, I want to update this to use the standard checkout and just set the env variable now.

Gundersanne (approved):

PR #374, opened 2021-02-02
Drop learn container

martinpitt:
We have not used it in more than two years, it has been broken for a
while, got removed from bots [1] and it's coming back any time soon.

[1] https://github.com/cockpit-project/bots/pull/1614


marusak (approved):
Is it already spring? (since doing big spring cleanups :) )

martinpitt:
marusak: Heh, I mostly noticed because my "git grep bots" produced a giant pile of noise from that jsonl file :grin:

PR #375, opened 2021-02-02
tasks: Introduce environment variables for bots repo/branch

martinpitt:
This will allow us to regression-test a bots PR with run-local.sh.

PR #376, opened 2021-02-03
tasks: Replace run-local.sh tests-trigger with testmap injection

martinpitt:
In <https://github.com/cockpit-project/bots/pull/1623> the bots testmaps
start to support injecting an additional test with
`COCKPIT_TESTMAP_INJECT=branch/context`. This allows run-local.sh to use
the same event/tests-scan workflow as production, without the `_manual`
testmap hack and calling tests-trigger.

The latter approach would not work at all for writing similar
integration self-tests for bots, as doing tests-trigger races with the
real bots (the bots repo has an issues webhook trigger, unlike
cockpituous).

 - [x] Builds on top of PR #375 
 - [x] Builds on top of PR #377
 - [x] Requires PR #378 
 - [x] Requires bots testmap injection: https://github.com/cockpit-project/bots/pull/1623
 - [x] Drop `COCKPIT_BOTS_{REPO,BRANCH}` hack from workflow again

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -129,6 +129,7 @@ podman run -d -it --name cockpituous-tasks --pod=cockpituous \
     -e COCKPIT_CA_PEM=/run/secrets/webhook/ca.pem \
     -e COCKPIT_BOTS_REPO=${COCKPIT_BOTS_REPO:-} \
     -e COCKPIT_BOTS_BRANCH=${COCKPIT_BOTS_BRANCH:-} \
+    -e COCKPIT_TESTMAP_INJECT=master/unit-tests \
     -e AMQP_SERVER=localhost:5671 \
     -e TEST_PUBLISH=sink-local \
     quay.io/cockpit/tasks:${TASKS_TAG:-latest}
@@ -169,7 +170,7 @@ if [ -n "$PR" ]; then
 
     podman exec -i cockpituous-tasks sh -exc "
     cd bots;
-    ./tests-trigger -f --repo $PR_REPO $PR unit-tests;
+    ./tests-scan -p $PR --amqp 'localhost:5671' --repo $PR_REPO;
     for retry in \$(seq 10); do
         ./tests-scan --repo $PR_REPO -vd;
         OUT=\$(./tests-scan --repo $PR_REPO -p $PR -dv);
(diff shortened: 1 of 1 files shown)


martinpitt:
All dependencies landed, so this is now relatively urgent to unbreak tests for cockpituous again.

tasks/run-local.sh
```diff
@@ -169,15 +170,14 @@ if [ -n "$PR" ]; then
 
     podman exec -i cockpituous-tasks sh -exc "
     cd bots;
-    ./tests-trigger -f --repo $PR_REPO $PR unit-tests;
+    ./tests-scan -p $PR --amqp 'localhost:5671' --repo $PR_REPO;
```
marusak, line 173: Does it make sense in this order? To do tests-scan, then do loop waiting for status to appear (the one, we already hope to have scanned) and then check if the original scan inserted it into queue?
Or am I reading this wrongly?

Gundersanne: I'm confused as to where the test is triggered now? Does `COCKPIT_TESTMAP_INJECT` mean you can run tests-scan and it'll always return the value ? If so it kinda feels like using less of the bots in the process and introducing a hack :/

martinpitt: The previous tests-trigger was only a crutch because we had the `unit-tests` context in `_manual`, so that a direct tests-scan wouldn't see it. Now I think it makes sense -- we call tests-scan (with the injected context from the environment), which will trigger "unit-tests" on github *and* queue it. The retry loop below is read-only (`-d`), it just waits until tests-scan finished its thing and the status is visible on github. We could probably also drop it and just leave it to the "waiting for unit-tests run to finish..." loop, but I suppose Sanne's intention was to make it clearer in case of a failure *why* it failed -- was it not queued or never ran from the queue?

Or I entirely misunderstood your question.

martinpitt: @Gundersanne: See https://github.com/cockpit-project/bots/pull/1623 and my initial attempt at https://github.com/cockpit-project/bots/pull/1622 ; I just can't see how to make this work with tests-trigger and manual for bots (and that approach is totally not what our infra is doing). We need this little bit of a dynamic testmap for this purpose. Of course I'd be  happy to have a better approach.

Gundersanne: I think i understand it now, +1

Gundersanne (approved):

PR #378, opened 2021-02-03
tasks: Fix race condition in run-local.sh

martinpitt:
Wait for the images server to be fully up before calling image-upload.
In GitHub workflow it sometimes starts too late (perhaps because of
container image downloads happening in the background?)

Also add "cockpituous-images" to the list of valid DNS names covered by
the TSL certificate, so that we don't need to jump through hoops to use
curl directly.


martinpitt:
Ugh, this just has cost me 1.5 hours of "WTF??" staring at my relatively simple code changes in #376. As the certificate build change triggers a container rebuild, I'd appreciate landing this quickly (so that other PRs can test quicker).

marusak (approved):
good catch!

PR #379, opened 2021-02-08
Ansible deployment updates

martinpitt:
I ran all affected playbooks and validated that they worked on our target machines. Among other things, this also finally replaced the docker.io/cockpit/images container with the quay one on our AWS sink and log servers.


martinpitt:
Oops, the "tasks" can't work from a fork. I'll re-push this to origin.

PR #380, opened 2021-02-08
Ansible deployment updates

martinpitt:
I ran all affected playbooks and validated that they worked on our target machines. Among other things, this also finally replaced the docker.io/cockpit/images container with the quay one on our AWS sink and log servers.

PR #381, opened 2021-02-08
ansible: Add SSH gateway user/key to our sink

martinpitt:
This can be used to temporarily expose ports on workflow/pipeline
machines by doing port forwarding to our sink.

Create a new unprivileged user and use a separate key for this,  to get
proper privilege separation to the sink.

### DIFF
--- a/ansible/aws/README.md
+++ b/ansible/aws/README.md
@@ -103,6 +103,7 @@ Public log sink/server setup
        ansible-playbook -i inventory aws/setup-host.yml
        ansible-playbook -i inventory maintenance/sync-secrets.yml
        ansible-playbook -i inventory cockpituous/sink.yml
+       ansible-playbook -i inventory aws/sink-ssh-gateway.yml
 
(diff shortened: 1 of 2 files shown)


martinpitt:
This does not actually depend on #380, so can be reviewed separately.

martinpitt:
I committed the new SSH keypair to the ci-secrets repo, rolled it out, and ran this playbook. It is deployed now. I also updated the sink's security groups in EC2 to allow all incoming ports -- that's fine, there is no port on that machine that we don't want to expose.

I can now run this:

    ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o CheckHostIP=no -R ':9999:localhost:9090' -i ~/redhat/ci-secrets/workflows/SINK_SSHRELAY_PRIVATE sshgw@logs.cockpit-project.org

and then look at my own laptop's cockpit on https://logs.cockpit-project.org:9999/

KKoukiou (approved):

PR #383, opened 2021-02-08
ansible: Fix sink url pattern

martinpitt:
We shouldn't use `ansible_host` there, as that will be something clunky
like `ec2-<IP>.*.amazonaws.com` and not match the TLS certificate's
subjectAlternateNames.

Configure the logs.cockpit-project.org domain name instead. We had this
for  a long time, until the re-deployment earlier this morning changed
it.

### DIFF
--- a/ansible/cockpituous/sink.yml
+++ b/ansible/cockpituous/sink.yml
@@ -74,8 +74,9 @@
     copy:
       dest: ~sink/.config/sink
       mode: 0644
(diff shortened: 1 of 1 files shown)


martinpitt:
Noticed in https://github.com/cockpit-project/cockpit/pull/15274 which started to get ugly log URLs. This was another case of us not properly re-deploying everything when we moved to the dynamic inventory.

marusak (approved):
Thanks!

PR #385, opened 2021-02-10
tasks: Fix and simplify bots checkout

martinpitt:
Commit 8b27da1eb44ff had a bug where the *first* cockpit-tasks round
always checked out the default branch of bots, instead of
`$COCKPIT_BOTS_BRANCH`. This led to bots PR failures when running
against an outdated master branch from a clone (as they usually are).
Fix that by using the `clone -b` option.

The bots checkout is now only for cockpit-tasks's benefit -- the
individual tasks and tests all check out their own bots. Thus we can
simplify `update_bots()` and only differ between "new checkout for the
first time" and "pull to get the latest version".

-----

Seen in [this bots PR](https://github.com/cockpit-project/bots/pull/1660) in [this workflow](https://github.com/cockpit-project/bots/pull/1660/checks?check_run_id=1871579295)

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -35,7 +35,7 @@ function update_bots() {
     if [ -d "$BOTS_DIR" ]; then
         git -C "$BOTS_DIR" pull --rebase
     else
-        git clone -b "$COCKPIT_BOTS_BRANCH" "$COCKPIT_BOTS_REPO" "$BOTS_DIR"
+        git clone --quiet -b "$COCKPIT_BOTS_BRANCH" "$COCKPIT_BOTS_REPO" "$BOTS_DIR"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Nice, [green](https://github.com/cockpit-project/cockpituous/pull/385/checks?check_run_id=1872952004)! That's exactly the kind of change that is self-validated now. But, `git pull` complains a lot about
```
warning: Pulling without specifying how to reconcile divergent branches is
discouraged. You can squelch this message by running one of the following
commands sometime before your next pull:
```

while I am at it, l also want to fix the super-verbose git clone, which uses hundreds of log lines like

   remote: Counting objects:  47% (24/51)

martinpitt:
https://github.com/cockpit-project/bots/pull/1660 now got green, so the race is fixed.

Log is much cleaner now. I rebuilt the container and deployed it. Waiting for some successful test runs now.

martinpitt:
All the tests in https://github.com/cockpit-project/cockpit/pull/15302 ran against the new container. Unblocking.

Gundersanne (approved):
How nice to have this self-testing container :D

PR #386, opened 2021-02-16
tasks: Call inspect-queue on local AMQP server in run-local.sh

martinpitt:
https://github.com/cockpit-project/bots/pull/1687 changes the default
AMQP server to the real official one on CentOS CI. Specify the local one
explicitly to avoid relying on the internal default.

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -178,7 +178,7 @@ if [ -n "$PR" ]; then
         echo waiting until the status is visible;
         sleep 10;
     done;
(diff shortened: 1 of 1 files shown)


Gundersanne (approved):

marusak:
Isn't ./inspect-queue reading AMQP env. variable?

martinpitt:
> Isn't ./inspect-queue reading AMQP env. variable?

No, that's just something that the kubernetes resource sets, and tasks/webhook reads.

PR #388, opened 2021-02-18
images: Minimize container

martinpitt:
Use a multi-stage build [1], pick some "small container targetted" basic
packages, and remove some larger packages that we really don't need in
the container. This shrinks the container from 313 to 137 MB.

[1] https://docs.docker.com/develop/develop-images/multistage-build/

----

This was mostly a learning opportunity for me, I just saw @asamalik 's and @fatherlinux 's excellent [building smaller container images](https://devconfcz2021.sched.com/event/gmSS/building-smaller-container-images) devconf.cz talk. But then why not, now I earned some twitter bragging rights. :grin: 

I tested this locally, but our CI will test it in full.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,11 +1,23 @@
-FROM fedora:latest
+FROM fedora:latest AS builder
 LABEL maintainer='cockpit-devel@lists.fedorahosted.org'
 
-RUN dnf -y update && \
-    dnf -y install nginx openssh-server /usr/bin/python && \
-    dnf clean all && \
-    mkdir -p /home/user
+# avoid unnecessary stuff in the container; systemd is a protected package, so apply some extra force
+RUN source /etc/os-release && \
+    dnf install -y --releasever=$VERSION_ID --setopt=install_weak_deps=False --installroot=/build \
+        coreutils-single \
+        glibc-minimal-langpack \
+        nginx \
+        openssh-server \
+        tar \
+        /usr/bin/python3 && \
+    rpm --root=/build --verbose --erase --nodeps systemd && \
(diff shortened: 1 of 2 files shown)


martinpitt:
Ugh, what is that now -- totally unrelated:
```
Error: error streaming container content for copy up into volume 8c71c95cfd6484af8af086d9467dc6d8bf3d60070b24b774a9f36c65c8543231: copier: get: globs [/rabbitmq] matched nothing (0 filtered out): no such file or directory
```

martinpitt:
OK, seems I first need to debug that weird podman regression...

martinpitt:
I reported the podman failure at https://github.com/containers/podman/issues/9432 , found a workaround, and added it here.

martinpitt:
The podman failure is known upstream and just got fixed. The current [failure](https://github.com/cockpit-project/cockpituous/pull/388/checks?check_run_id=1933216541) is now entirely my  fault: `No such file or directory: 'tar'`.  Yay integration tests!

martinpitt:
Nice, works! Just pushing again to clean up the commit message for the podman workaround.

marusak (approved):
Cool stuff indeed!
I am not super convinced it actually is necessary for this image to try to slim it down for the cost of making the dockerfile much harder to read.
I am not against nor in favor.

martinpitt:
Right, this was mostly a fun little learning project for me. It's not really relevant for production in this case, I'm also happy to just close this.

martinpitt:
I'd like to gain some experience with this, and also land the podman regression workaround. So if it's all the same to you, I'll land this. If it causes any problems, I'll revert.

PR #390, opened 2021-02-21
ansible: Add playbook to reset docker on e2e machines

martinpitt:
Sometimes docker gets itself into a bad state where containers are bad,
undeletable, or starting/stopping them throws weird storage errors. This
playbook does a "factory reset". This removes all containers, images,
and volumes.


martinpitt:
I just used that to repair some accrued damage on our e2e machines. Don't ask too much about the sleeps.. It was not totally obvious how to do that right, so I'd like to keep this as a recipe.

martinpitt:
I keep forgetting that we should submit PRs from origin now, so that the tasks test can work.

PR #391, opened 2021-02-21
ansible: Add playbook to reset docker on e2e machines

martinpitt:
Sometimes docker gets itself into a bad state where containers are bad,
undeletable, or starting/stopping them throws weird storage errors. This
playbook does a "factory reset". This removes all containers, images,
and volumes.

----

I just used that to repair some accrued damage on our e2e machines. Don't ask too much about the sleeps.. It was not totally obvious how to do that right, so I'd like to keep this as a recipe.


marusak (approved):
I mean, this is ugly. Going around deleting `/var/lib/docker` and also sleeping between commands... But! If it is just what it takes and it is what would one do manually anyway, then surely having simple task for it won't hurt us.

PR #392, opened 2021-02-24
ansible: Update reset-docker playbook

martinpitt:
The sleeps introduced in commit 83d8c4d9f81e7c were not necessary after
all, they were just a leftover when the commands were done in the wrong
order.

In addition, stop and restart the work containers.


martinpitt:
The bots were behaving strangely again, so I just did that exercise again.

Gundersanne (approved):

PR #398, opened 2021-03-09
tasks: Increase /tmp to 14 GB

martinpitt:
Current cockpit tests often need a bit more than 10 GB as especially the
nondestructive VMs pile up a lot of cruft in their overlays.

----

I deployed this, and our tasks containers now have a 14 GB /tmp.

This happend particularly badly in https://github.com/cockpit-project/cockpit/pull/15484 , I'll now retry this PR to validate that it is enough.

### DIFF
--- a/tasks/cockpit-tasks-centosci.yaml
+++ b/tasks/cockpit-tasks-centosci.yaml
@@ -63,4 +63,4 @@ spec:
       - name: tmp
         emptyDir:
           medium: Memory
-          sizeLimit: 10G
+          sizeLimit: 14G
--- a/tasks/install-service
+++ b/tasks/install-service
@@ -7,6 +7,8 @@ CACHE=/var/cache/cockpit-tasks
 INSTANCES=${INSTANCES:-3}
 TMPFS_GB=14
 
+systemctl stop 'cockpit-tasks@*.service'
+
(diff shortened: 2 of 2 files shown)


martinpitt:
https://github.com/cockpit-project/cockpit/pull/15484 succeeded, so this is good.

tasks/cockpit-tasks-centosci.yaml
```diff
@@ -63,4 +63,4 @@ spec:
       - name: tmp
         emptyDir:
           medium: Memory
-          sizeLimit: 10G
+          sizeLimit: 14G
```
marusak, line 66: Why do "nondestructive VMs pile up a lot of cruft in their overlays"? I guess there are stuff like journal and pcp that run the whole time, but I don't think that it should pile up a few GBs in the hour it runs.

martinpitt: It's not in the file system. The overlays work on the  block layer, so if you create a 10 MB file, delete it, and do this a couple of times, file systems tend to allocate different blocks each. This reflects in the overlay, which over time will fill the entire virtual disk size. I think there's some techniques to compress them regularly, but I don't want to spend much time on that right now honestly.

martinpitt:
@marusak: It's not a short time, these @nd VMs run for a few hours and run hundreds of tests.

PR #399, opened 2021-03-10
ansible: Restart all containers in sync-secrets.yml

martinpitt:
Updating the secrets directory yanks it away from existing running
containers and thus breaks them. Commit c78c681725b13 already fixed for
the images containers, but it applies to tasks as well.


martinpitt:
I just rolled out some new secrets for @allisonkarlitskaya , we noticed /secrets being empty in running containers. I ran the fixed  playbook, containers restarted properly everywhere, and secrets seem okay.

martinpitt:
argh, need to push this to origin for CI

PR #400, opened 2021-03-10
ansible: Restart all containers in sync-secrets.yml

martinpitt:
Updating the secrets directory yanks it away from existing running
containers and thus breaks them. Commit c78c681725b13 already fixed for
the images containers, but it applies to tasks as well.

### DIFF
--- a/ansible/maintenance/sync-secrets.yml
+++ b/ansible/maintenance/sync-secrets.yml
@@ -36,8 +36,8 @@
         cmd: "{{ oc_command }} create -f -"
         stdin: "{{ build_secrets.stdout }}"
 
-    - name: Restart image pods to pick up changed certificates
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Not my area, but I definitely confirm that this is needed (after observing what happens without it)

One note: is there any chance that someone will create adhoc containers in OpenShift that we shouldn't kill?  Think similar to how we sometimes work in the cockpit cluster...

martinpitt:
@allisonkarlitskaya : On e2e the restarts are restricted to the systemd-controlled containers. On OCP they *would* kill manually started containers indeed. Honestly I'm not too worried about them, as (1) we run this playbook maybe twice a year, and (2) you are not supposed to have ad-hoc containers on the infra for nontrivial amounts of time (in fact we often forget to kill them).

We could limit them with enumerating all labels, but as we saw this is also error prone. But :man_shrugging: I'm ok either way.

allisonkarlitskaya:
It sounds fine to me, thanks :)

PR #402, opened 2021-03-11
run-local.sh: fix SELinux label issues

allisonkarlitskaya:
run-local.sh makes liberal use of containers with bind-mounted files and
directories from the host.  This is a recipe for disaster with SELinux enabled,
leading to messages about waiting for for RabbitMQ to come up.

Let's sprinkle -v :,z around everywhere to relabel the files in question.
Pretty much all of them are exclusively intended to be bind-mounted into
containers anyway.

There remain two issues to running this on a stock Fedora 34 Silverblue system:
openssl needs to be installed, and the python3 'yaml' module as well.  In both
of those cases, though, the error message is clear and the solution is obvious.


martinpitt (approved):
Cheers!

PR #404, opened 2021-03-15
Debug test failure

martinpitt:


martinpitt:
This seems to happen because the tasks container now iterates through the "30 iterations" loop *very* quickly -- it does not seem to block on AMQP for the usual few seconds any more. Thus it rips the tasks container away from the main test prematurely. I suppose for this interactive testing it would be good to have better control over the iteration -- do it indefinitely and tell it when to stop (which will then also speed up the test).

PR #406, opened 2021-03-19
tasks: Set up S3 secrets for OpenShift, stop iteration when receiving SIGTERM

martinpitt:
I tested this locally with a hacked up secrets dir and a local container. The integration tests model the situation quite closely.

The "test local deployment" step now takes only 2:15 minutes instead of the previous 7 minutes. :tada:

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -21,6 +21,16 @@ if [ -r /secrets/npm-registry.crt ]; then
     export NODE_EXTRA_CA_CERTS=/secrets/npm-registry.crt
 fi
 
+# set up S3 keys for OpenShift secrets volume
+if [ ! -d /secrets/s3-keys ]; then
+    # then our container symlink will point into the void, replace it with a directory and set up all files that we can find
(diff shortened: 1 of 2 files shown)


tasks/cockpit-tasks
```diff
@@ -21,6 +21,16 @@ if [ -r /secrets/npm-registry.crt ]; then
...
 
+# set up S3 keys for OpenShift secrets volume
+if [ ! -d /secrets/s3-keys ]; then
+    # then our container symlink will point into the void, replace it with a directory and set up all files that we can find
+    rm ~/.config/cockpit-dev/s3-keys
+    mkdir ~/.config/cockpit-dev/s3-keys
+    find /secrets/ -name 's3-keys--*' | while read f; do
+        ln -s "$f" ~/.config/cockpit-dev/s3-keys/"${f#*--}"
+    done
+fi
```
allisonkarlitskaya, line 32: it ain't pretty, but it'll work.

allisonkarlitskaya (approved):

PR #407, opened 2021-03-19
tasks: Fix OpenShift S3 keys setup

martinpitt:
Fix commit cce26c464: The files in /secrets from the OpenShift secrets
volume are themselves links to a hidden ..data/ subdirectory. This
caused `find` to see them twice, and ran into "ln: file exists" errors.

Use a glob instead, and protect against zero matches with `test -e`.

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -26,7 +26,8 @@ if [ ! -d /secrets/s3-keys ]; then
     # then our container symlink will point into the void, replace it with a directory and set up all files that we can find
     rm ~/.config/cockpit-dev/s3-keys
     mkdir ~/.config/cockpit-dev/s3-keys
-    find /secrets/ -name 's3-keys--*' | while read f; do
+    for f in /secrets/s3-keys--*; do
+        [ -e "$f" ] || continue # non-matching glob
         ln -s "$f" ~/.config/cockpit-dev/s3-keys/"${f#*--}"
     done
 fi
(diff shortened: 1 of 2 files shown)


tasks/cockpit-tasks
```diff
@@ -26,7 +26,8 @@ if [ ! -d /secrets/s3-keys ]; then
     # then our container symlink will point into the void, replace it with a directory and set up all files that we can find
     rm ~/.config/cockpit-dev/s3-keys
     mkdir ~/.config/cockpit-dev/s3-keys
-    find /secrets/ -name 's3-keys--*' | while read f; do
+    for f in /secrets/s3-keys--*; do
```
allisonkarlitskaya, line 29: This is a lot nicer thanks

tasks/cockpit-tasks
```diff
@@ -26,7 +26,8 @@ if [ ! -d /secrets/s3-keys ]; then
...
+        [ -e "$f" ] || continue # non-matching glob
```
allisonkarlitskaya, line 30: continue?  with what?  :rofl:

allisonkarlitskaya (approved):

tasks/cockpit-tasks
martinpitt: In theory there could also be a single broken symlink in the secrets; feels a bit more robust. This is just my go-to pattern for dealing with POSIX glob behaviour..

martinpitt:
[built/pushed](https://github.com/cockpit-project/cockpituous/actions/runs/667629538), rolled out to CentOS CI, and I confirmed that in its tasks container I can now run `python3 -m lib.s3 ls https://cockpit-images.eu-central-1.linodeobjects.com/`

Now waiting for a successful test result.

PR #410, opened 2021-04-07
tasks: Add git-lfs

mvollmer:
So that we can check out from repositories that use git lfs.

Note: git lfs also uses hooks to perform its magic, but they seem to
be only needed when actually making changes to files tracked by lfs
and pushing them.  But simply cloning a reporsitory that uses lfs
works fine as long as the git-lfs RPM is installed and configured
globally (which happens during installation).

To setup up the hooks, we would need to run "git lfs install" in the
local repository after cloning.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -16,6 +16,7 @@ RUN dnf -y update && \
         fpaste \
         gcc-c++ \
         git \
+        git-lfs \
         gnupg \
         intltool \
         jq \


martinpitt:
:+1 on the general approach, git-lfs rocks. Next steps: run the [build-tasks](https://github.com/cockpit-project/cockpituous/actions/workflows/build-tasks.yml) workflow to build and push the container, prod @marusak or me to roll it out to ci, wait for some successful test runs, and then land.

martinpitt (approved):

martinpitt:
[tasks container built](https://github.com/cockpit-project/cockpituous/actions/runs/725965049) and deployed.

martinpitt:
Tests have seemed fine since then.

PR #411, opened 2021-04-08
tasks: cleanup S3 as part of pruning

allisonkarlitskaya:
At tasks container shutdown, also take a chance to run image-prune on
the S3 stores.

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -80,3 +80,8 @@ done
 # Prune old images
 update_bots
 ./image-prune
+
+# Clean up S3
+for region in eu-central-1 us-east-1; do
+    ./image-prune --s3 "https://cockpit-images.${region}.linodeobjects.com/" || true
+done


tasks/cockpit-tasks
```diff
@@ -80,3 +80,8 @@ done
 # Prune old images
 update_bots
 ./image-prune
+
+# Clean up S3
+for region in eu-central-1 us-east-1; do
+    ./image-prune --s3 "https://cockpit-images.${region}.linodeobjects.com/" || true
```
martinpitt, line 86: Is that alternative truth just to make the integration tests happy? TBH I'd rather get a "bot crashed" notification if pruning fails. If it's just for the tests, could that rather be guarded by some `[ -e the_secret ]`?

allisonkarlitskaya: I was actually thinking about network issues...

allisonkarlitskaya: ...but now that you mention it, I've heard that it's better to beg forgiveness than to ask permission.

martinpitt (approved):
Ack, let's land that to get the cleanup going, and we can fine-tune this.

PR #412, opened 2021-04-16
Add selinux-policy packages into task image

marusak:
As needed for https://github.com/cockpit-project/cockpit/pull/15707

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -3,6 +3,7 @@ LABEL maintainer='cockpit-devel@lists.fedorahosted.org'
 
 # HACK: chromium-headless 88 crashes with some keyDown commands: https://bugs.chromium.org/p/chromium/issues/detail?id=1170634
 # install full chromium for the time being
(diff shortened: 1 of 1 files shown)


tasks/Dockerfile
```diff
@@ -40,6 +40,8 @@ RUN dnf -y update && \
         rpmdevtools \
         rsync \
         sassc \
+        selinux-policy \
+        selinux-policy-devel \
```
martinpitt, line 44: Dou you mind moving that after the `builddep` with a HACK comment pointing to the cockpit PR? We can drop this once it lands in fedora, as then builddep will install it  for us.

martinpitt (changes requested):

tasks/Dockerfile
marusak: Right, good call. Will do.

martinpitt (approved):
Cheers! Want to deploy this yourself, or want me to?

marusak:
nvm, this needs to go into image.
I removed it from quay.io and retaged the old image.

PR #413, opened 2021-05-07
aws: Update RHEL AMI

martinpitt:
Move to the current 8.4 daily. The previous image was still 8.2.

----

I tested this today with an UX demo instance as well as a "big" tasks runner (as e2e broke down last night)


marusak (approved):
What a nice and readable way to declare it :)

PR #414, opened 2021-05-26
metrics: Add prometheus

martinpitt:
Start collecting our CI metrics, so that we can soon deploy a public
Grafana instance.

----

I deployed this to CentOS CI, and it is happily runnning now. To validate the service you can `oc rsh` into e.g. a tasks pod and try `curl http://prometheus:9090/prometheus/`. To validate the import, you can run `oc port-forward metrics-7474b48bb-g6tmz 9999:9090` and poke around in http://localhost:9999, and e.g. graph `test_run_seconds_count`.

### DIFF
--- a/metrics/metrics.yaml
+++ b/metrics/metrics.yaml
@@ -1,0 +1,93 @@
+---
+apiVersion: v1
+kind: List
+items:
+  - kind: Deployment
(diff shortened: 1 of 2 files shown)


marusak (approved):
I guess so? :) Thanks!

PR #418, opened 2021-06-08
tasks: Move to RabbitMQ container on quay.io

martinpitt:
Today we ran into a failure on spawning the webhook container due to
pull request rate limits from CentOS CI. Avoid this in the future, and
use our container copy on quay.io instead.

----

 - [ ] Deploy and validate on CentOS CI

### DIFF
--- a/tasks/cockpit-tasks-webhook.yaml
+++ b/tasks/cockpit-tasks-webhook.yaml
@@ -18,7 +18,7 @@ items:
       spec:
         containers:
           - name: amqp
-            image: docker.io/rabbitmq:3-management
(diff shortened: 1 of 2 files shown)


martinpitt:
I could not make it work with the bitnami container. It gets configured via env vars and ignores rabbitmq.conf at least where we bind-mount it. Timeout, moving to a straight copy on quay.io for now.

[incomplete patch](https://github.com/cockpit-project/cockpituous/files/6621323/cockpituous-bitnami-rabbitmq.patch.txt) (including debug stuff)

marusak (approved):

PR #420, opened 2021-06-10
Convert auxiliary playbooks to roles, simplify setup

martinpitt:
I tested this with a brand new AWS task runner. It's nice to now bring one up with a single command, and without affecting already existing instances.

Now that CentOS CI is back, I also validated that the sync-secrets playbook still updates all CentOS CI (k8s) and AWS (pittishift) runners.


martinpitt:
Argh, [integration test fail](https://github.com/cockpit-project/cockpituous/pull/420/checks?check_run_id=2795065962), need to submit that from origin.

PR #421, opened 2021-06-10
Convert auxiliary playbooks to roles, simplify setup

martinpitt:
I tested this with a brand new AWS task runner. It's nice to now bring one up with a single command, and without affecting already existing instances.

Now that CentOS CI is back, I also validated that the sync-secrets playbook still updates all CentOS CI (k8s) and AWS (pittishift) runners.


martinpitt:
Some more tweaks, just tested IRL with 

    ansible-playbook -i inventory -e hostname=cockpit-aws-tasks-3 aws/launch-tasks.yml

(as we need more capacity right now)

mvollmer (approved):

PR #422, opened 2021-06-14
tasks: Kill libvirtd in between iterations

martinpitt:
Also ensure that there are no leftover VMs.

This may help with

    Failed to connect socket to '/work/.cache/libvirt/libvirt-sock'

errors which have plagued us for a while. Normally libvirtd removes the
socket right before it times out, but either there is a race condition
which we hit often, or something removes stuff in ~/.cache/libvirtd/.

----

This is totally a shot into the dark, to fix failures [like this](https://logs.cockpit-project.org/logs/pull-15934-20210611-093034-a98e957e-fedora-34/log.html).

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -71,6 +71,10 @@ for i in $(seq 1 30); do
     update_bots
     cd "$BOTS_DIR"
 
+    # avoid stale state and sockets
+    virsh list --id --all | xargs --no-run-if-empty virsh destroy
+    pkill libvirtd || true
+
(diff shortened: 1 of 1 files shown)


KKoukiou (approved):

martinpitt:
Container [built](https://github.com/cockpit-project/cockpituous/actions/runs/935196238) and deployed. Let's give it a few days and check if we still get these failures.

allisonkarlitskaya (approved):
GitHub needs a button labelled "strongly approve".

PR #423, opened 2021-06-16
metrics: Add Grafana

martinpitt:
Enable anonymous access. Our raw metrics are public already anyway (on
purpose).

There is still some manual setup required for setting the admin password
and importing the Cockpit CI dashboard. Document these for the time
being.

-----

I already deployed this. See https://grafana-frontdoor.apps.ocp.ci.centos.org/d/ci/cockpit-ci and https://github.com/cockpit-project/cockpituous/tree/grafana/metrics to pre-view the documentation.


martinpitt:
https://grafana-frontdoor.apps.ocp.ci.centos.org/d/ci/cockpit-ci is the proof that this works. @allisonkarlitskaya , @mvollmer , could I get a formal approval so that we can land this? I'd like the cockpituous repo to be in sync with reality.

allisonkarlitskaya (approved):

PR #424, opened 2021-06-16
webhook: add an /inspect-queue endpoint

allisonkarlitskaya:
Fetching http://webhook/inspect-queue should now display the output of
inspect-queue.

This lets us see what's in the queue without having to have secrets
setup locally.

 - [x] uhhh, better wait until pitti gets back

### DIFF
--- a/tasks/webhook
+++ b/tasks/webhook
@@ -168,6 +168,16 @@ class ReleaseHandler(github_handler.GithubHandler):
         publish_to_queue('webhook', event, request)
         return None
 
+    def do_GET(self):
+        if self.path == '/inspect-queue':
+            self.wfile.write(b'HTTP/1.1 200\r\nContent-Type: text/plain; charset=UTF-8\r\n\r\n')
+            try:
+                subprocess.check_call([f'{BOTS_CHECKOUT}/inspect-queue'], stdout=self.wfile)
+            except (subprocess.CalledProcessError, FileNotFoundError) as e:
+                self.wfile.write(b'Failed: ' + bytes(str(e), 'utf-8'))
+        else:
+            self.wfile.write(b'HTTP/1.1 404\r\n\r\nNot found')
+
 
 #
 # main


allisonkarlitskaya:
I tested this locally with a 'hello world' shellscript in `/tmp/home/bots/inspect-queue`, and it seems to work.  It would be nice if you know a better way to test it "for real" before we merge it, though...

martinpitt:
That much ought to be possible with run-local.sh, and curling the webhook pod?

No immediate idea about the tests/tasks failure -- apparently the log file never made it into the sink. Most likely a flake, as the "unit-tests" fake task did run fine. Retrying that one.

martinpitt:
> Retrying that one.

Nevermind, retrying can't work, as we can't make the PR "forget" the already existing unit-tests status. Retrying needs a force-push.

tasks/webhook
```diff
@@ -163,6 +163,16 @@ class ReleaseHandler(github_handler.GithubHandler):
         publish_to_queue('webhook', event, request)
         return None
 
+    def do_GET(self):
+        if self.path == '/inspect-queue':
+            self.wfile.write(b'HTTP/1.1 200\r\nContent-Type: text/plain; charset=UTF-8\r\n\r\n')
+            try:
+                subprocess.check_call([f'{BOTS_CHECKOUT}/inspect-queue'], stdout=self.wfile)
+            except (subprocess.CalledProcessError, FileNotFoundError):
+                self.wfile.write(b'Failed.')
```
martinpitt, line 172: Not much of an error message :-)

martinpitt (dismissed):
FTR, we need to use this sparingly in this form. inspect-queue takes quite some time, and as the handler is synchronous, no other actual webhook requests will be served during that time, and thus time out (GitHub has a relatively short timeout of ~ 10s). It's ok for now, though -- this isn't something that we use often (and personally I just oc rsh to a pod, or check out the secrets)

So :+1 in general, but tests should be green. After that this needs a tasks container rebuild (action) and rolling out to our infra (cockpituous ansible scripts, requires CentOS CI login)

tasks/webhook
allisonkarlitskaya: Added the `str()` of the exception.

tasks/webhook
```diff
@@ -168,6 +168,16 @@ class ReleaseHandler(github_handler.GithubHandler):
         publish_to_queue('webhook', event, request)
         return None
 
+    def do_GET(self):
+        if self.path == '/inspect-queue':
+            self.wfile.write(b'HTTP/1.1 200\r\nContent-Type: text/plain; charset=UTF-8\r\n\r\n')
+            try:
+                subprocess.check_call([f'{BOTS_CHECKOUT}/inspect-queue'], stdout=self.wfile)
+            except (subprocess.CalledProcessError, FileNotFoundError) as e:
+                self.wfile.write(b'Failed: ' + bytes(str(e), 'utf-8'))
```
martinpitt, line 177: That's a curious way of saying `str(e).encode()`, but sure

allisonkarlitskaya: didn't have enough coffee yet :)

The real story: I wondered if I could go directly to `bytes()` from the exception, and tried that first.  Turns out, you can't.  So then I added in the `str()`.

martinpitt (approved):
Green! The bytes() thing is not super-important, so good enough. Next step: Rebuild/push the tasks container from this branch (not master!!) with the [build-tasks workflow](https://github.com/cockpit-project/cockpituous/actions/workflows/build-tasks.yml)

allisonkarlitskaya:
> Green! The bytes() thing is not super-important, so good enough. Next step: Rebuild/push the tasks container from this branch (not master!!) with the [build-tasks workflow](https://github.com/cockpit-project/cockpituous/actions/workflows/build-tasks.yml)

https://github.com/cockpit-project/cockpituous/actions/runs/967322989

allisonkarlitskaya:
> https://github.com/cockpit-project/cockpituous/actions/runs/967322989

Done.  @martinpitt Please deploy :)

martinpitt:
I restarted the webhook pod, and http://webhook-frontdoor.apps.ocp.ci.centos.org/inspect-queue works. I now deployed the new tasks container everywhere else (so watch out for "weird" test regressions due to the newer Fedora)

PR #427, opened 2021-07-06
remove sedery from release-srpm

allisonkarlitskaya:
This bit us on the latest cockpit-ostree release.  Make two changes:

 - use git (not sed) to strip the parts of the tag we don't care about
 - use grep to force the correct format, instead of using sed to attempt to "fix" things

 - [x] rebuild release container: https://github.com/cockpit-project/cockpituous/actions/runs/1004887412
 - [x] test cockpit-ostree release 186.1 with bogus tag body: https://github.com/cockpit-project/cockpit-ostree/actions/runs/1004908828

### DIFF
--- a/release/release-srpm
+++ b/release/release-srpm
@@ -62,13 +62,6 @@ dash_bullets()
             -e 's/^[^ *-]/- /'
 }
 
-# Trim tag message on stdio
-trim_tag()
-{
-    # Remove line with just tag name. Remove signature
-    sed -e '/^\s*'$TAG'\s*$/d' -e '/-----BEGIN/,/-----END/d'
-}
-
 # Format a date for a changelog
 #  $1: The date to format
 changelog_date()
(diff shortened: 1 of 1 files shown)


martinpitt (dismissed):
LGTM, thank you!

release/release-srpm
```diff
@@ -111,6 +97,12 @@ tagemail=%(taggeremail)' "refs/tags/$1")
         tagbody="- Update to upstream $TAG release"
     fi
 
+    # ensure that the body only contains point form
+    if printf "%s\n" "$tagbody" | grep -v '^- ' >&2; then
+        echo '*** tag body not in point form (see above)' "$tagbody" >&2
```
allisonkarlitskaya, line 102: I meant to get rid of "$tagbody" here.

martinpitt (approved):
See above.. and below! :grin:

PR #428, opened 2021-07-07
sink: avoid logging GitHub tokens

allisonkarlitskaya:
Use a regexp to find and blind strings which look like GitHub tokens.

At the point we apply the regexp, we're dealing with complete lines, so
we don't have to worry about the token having arrived in two separate
reads.

### DIFF
--- a/sink/sink
+++ b/sink/sink
@@ -549,6 +549,13 @@ def check_prune(config, logs, dry_run):
                 sys.exit(0)
 
 
+def strip_secrets(line):
+    # Strip out GitHub-style secrets
+    line = re.sub(b'gh[a-z]_[0-9A-Za-z]*', b'gh*_***', line)
+
+    return line
+
+
 def main():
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
Uh, tests are failing of course, because we regenerated the token and didn't update it yet.

sink/sink
```diff
@@ -549,6 +549,13 @@ def check_prune(config, logs, dry_run):
                 sys.exit(0)
 
 
+def strip_secrets(line):
+    # Strip out GitHub-style secrets
+    line = re.sub(b'gh[a-z]_[0-9A-Za-z]*', b'gh*_***', line)
```
martinpitt, line 554: This could have been `{5,}` instead of  `*` to not filter out trivial stuff like "gha_", but we can do with that for the time being.

martinpitt (approved):

PR #431, opened 2021-07-08
workflows: Run deployment tests with default GitHub token

martinpitt:
This avoids another usage of our cockpituous token.

---

 - [x] blocked on https://github.com/cockpit-project/bots/pull/2194
 
I am not sure if we even want that -- it's kinda sorta a validation that our cockpituous token works and has the required privileges. But then again, we usually don't run a cockpituous PR test when we change the token (we usually learn about it because our tests explode :boom: ). So for now this is more of a "let's see if this even works" exercise.

If this works and we want it, I'll do the same to bots -- this test run is the only usage of the cockpituous token by cockpituous.git and bots.git, so it'll make it easier to move this into an environment.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,10 @@ jobs:
 
   tasks:
     runs-on: ubuntu-20.04
+    permissions:
+      # enough permissions for tests-scan to work
+      pull-requests: read
+      statuses: write
     timeout-minutes: 30
     steps:
       - name: Clone repository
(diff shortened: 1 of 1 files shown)


martinpitt:
[Dang](https://github.com/cockpit-project/cockpituous/pull/431/checks?check_run_id=3018877228), not  enough:
```
ERROR:root:tests-scan: Error accessing /orgs/cockpit-project/teams
  Status: 403
  Reason: Forbidden
  Response: {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/reference/teams#list-teams"}
```

None of the [documented GitHub token permissions](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) apply to the org, and this would be a PAT's `read:org` permission.

This comes from [allowlist()](https://github.com/cockpit-project/bots/blob/main/task/github.py#L379) which calls teamIdFromName() which does that blanket `get("/orgs/cockpit-project/teams")`. Apparently we should handle the error in a better way, or even try to not iterate through all the teams -- at first sight this shouldn't be necessary.

martinpitt:
https://github.com/cockpit-project/bots/pull/2193 and https://github.com/cockpit-project/bots/pull/2194 landed now, so this actually works now. Ready for review.

allisonkarlitskaya (approved):
...seems suspiciously similar :thinking:

PR #434, opened 2021-07-14
metrics: Use light Grafana theme by default

martinpitt:
This provides much better contrast, in particular it makes warning
thresholds easier to see.

---

Already rolled out, as usual. Before I just changed that in the admin web UI.


marusak (approved):
Danke!

PR #436, opened 2021-08-05
ansible: Reboot e2e machines in reset-docker.yml

martinpitt:
In all cases so far I actually want to reboot the e2e machines, instead
of restarting the bots right away. This cleans piled up cruft in the
kernel/kvm which tends to cause funny effects/breakage in docker over
time.


marusak (approved):
Right, and on boot tasks are started.

PR #438, opened 2021-09-02
release: Build a node-cache tarball in release-source

martinpitt:
If a (plain Makefile) project defines a `make node-cache` rule, then
call this during release-source. We will gradually add these to our
projects to fulfill Fedora's packaging policy.

Drop the requirement that there is exactly one tarball.

-----

I tested this with cockpit-certificates, see https://github.com/skobyda/cockpit-certificates/pull/65.

I also tested this with cockpit-podman (which does not have such a rule), and it correctly produces a single _release/source/cockpit-podman-34.tar.gz

 - [ ] rebuild container after landing

### DIFF
--- a/release/release-source
+++ b/release/release-source
@@ -101,13 +101,12 @@ prepare()
     else
         # plain Makefile projects
         $MAKE -C $repodir dist || $MAKE -C $repodir dist-gzip
-        archive=$(find $repodir -maxdepth 1 -name "*-${TAG}.tar.*")
-        if [ "$(echo "$archive" | wc -l)" != 1 ]; then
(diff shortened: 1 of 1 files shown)


martinpitt:
https://github.com/skobyda/cockpit-certificates/pull/65 works, so this is ready for landing @mmarusak and @KKoukiou -- can you please have a look?

KKoukiou (approved):

martinpitt:
I [rebuilt the release container](https://github.com/cockpit-project/cockpituous/actions/runs/1216271616), and it [got pushed correctly](https://github.com/cockpit-project/cockpituous/pkgs/container/release)

PR #439, opened 2021-09-03
Two secrets fixes

martinpitt:
@marusak : I tried running sync-secrets from my machine, and ended up with correct permissions (644 instead of 664). So this is a shot into the dark -- at least with this fix the files have the same permissions as before.

I also added a nice little tweak for OpenShift secrets which I learned from @mh21 today, thanks!


martinpitt:
@marusak: FYI, I still can't log into the AWS web UI, but I used the CLI: `aws ec2 terminate-instances --instance-ids i-0cb38715731c2a049`. So it's all good.

marusak (approved):
Thanks!

PR #441, opened 2021-10-01
aws: Add metrics UX demo

martinpitt:
Set up a few resource hogging services/containers for users to find and
identify.

Some parts should be started about 20 to 30 minutes before the test
begins, create an `~admin/demo-prep.sh` script for that. The test
conductor needs to run this manually.

------

I started this with

    ansible-playbook -i inventory aws/launch-demo.yml 
    ansible-playbook -i inventory aws/demo/2021-10-metrics.yml

which launched one machine. I then immediately ran

    ssh admin@ec2-XXX-XXX-XXX.compute-1.amazonaws.com
    sudo ./demo-prep.sh

@andreasn , @marusak: Can you please play around with that instance and see that it works as intended?

### DIFF
--- a/ansible/aws/aws_defaults.yml
+++ b/ansible/aws/aws_defaults.yml
@@ -1,5 +1,5 @@
 ---
 aws_region: us-east-1
 # current RHEL 8
-aws_rhel_ami: ami-0b0af3577fe5e3532
+aws_rhel_ami: ami-03a454637e4aa453d
 aws_key_name: "cockpit-{{ lookup('env', 'USER') }}"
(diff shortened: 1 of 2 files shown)


martinpitt:
Hmm, the "show related logs" still feels rather wonky.. I can see the logs fine in `journalctl`, but they often don't match up well. If you want to reset what the script does, do something like

    sudo systemctl stop chatty big-brain
    sudo podman rm --all -f

Then you can run the demo-prep.sh script again.

andreasn (commented):
Played around with the instance, and looks good to me!

martinpitt:
Ack, thanks @andreasn ! I tore down the test instance. We can land this as-is, and refine if necessary.

marusak (approved):
Thanks!
Admittedly I also struggled with it a bit to figure out the network usage part, but it feels genuine. Let's see what testers will say.

PR #445, opened 2021-11-30
tasks: Add subjectAltName for recently changed e2e domain

martinpitt:
This will fix image upload/download from cockpit-11 after the DNS change.

Currently, uploading fails in CI like this:
```
❱❱❱ curl --cacert ~/upstream/bots/images/files/ca.pem https://cockpit-11.apps.cnfdb2.e2e.bos.redhat.com:8493/
curl: (60) SSL: no alternative certificate subject name matches target host name 'cockpit-11.apps.cnfdb2.e2e.bos.redhat.com'
```

I ran images/generate-image-certs.sh in our secrets git, pushed the result, and re-deployed the secrets. This first failed due to a syntax error, fixed here as well.

Now the above curl command succeeds again.

### DIFF
--- a/ansible/roles/install-secrets-openshift/tasks/main.yml
+++ b/ansible/roles/install-secrets-openshift/tasks/main.yml
@@ -8,5 +8,5 @@
      cmd: "{{ oc_command }} apply -f -"
      stdin: "{{ build_secrets.stdout }}"
 
-- name: Restart all containers to pick up new secrets
-  shell: "{{ oc_command }} get -o name pods | xargs -l -r {{ oc_command }} delete --wait=false"
+ - name: Restart all containers to pick up new secrets
+   shell: "{{ oc_command }} get -o name pods | xargs -l -r {{ oc_command }} delete --wait=false"
--- a/tasks/credentials/openssl.cnf
+++ b/tasks/credentials/openssl.cnf
@@ -51,4 +51,4 @@ extendedKeyUsage = 1.3.6.1.5.5.7.3.2
 basicConstraints = CA:false
 keyUsage = digitalSignature,keyEncipherment
 extendedKeyUsage = 1.3.6.1.5.5.7.3.1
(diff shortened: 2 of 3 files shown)


martinpitt:
Our integration test has some trouble. The first attempt [failed at a temporary github.com DNS resolution](https://github.com/cockpit-project/cockpituous/runs/4363389928?check_suite_focus=true), the second [timed out waiting for the image server](https://github.com/cockpit-project/cockpituous/runs/4363438318?check_suite_focus=true).

With a local `tasks/run-local.sh` this doesn't happen -- I wonder if the container rebuild breaks something? Trying a round without a rebuild.

images/generate-image-certs.sh
```diff
@@ -9,7 +9,7 @@ openssl req -new -newkey rsa:2048 -nodes -keyout server.key -out server.csr \
     -subj '/O=Cockpit/OU=Cockpituous/CN=cockpit-tests' \
     -extensions server_ca_extensions -reqexts server_ca_extensions \
     -config "$OPENSSL_CNF"
-openssl x509 -req -days 365000 -in server.csr -out server.pem \
+openssl x509 -req -days 720 -in server.csr -out server.pem \
```
allisonkarlitskaya, line 12: Everything I know says that this maxes out at a year and a month or so, ~395 days.

https://support.globalsign.com/ssl/general-ssl/397-day-maximum-tls-certificate-validity

Not like it matters or anything, though.  We use this for curl, and in case we use this in a browser, we manually accept it anyway.  I'd say either make it a year ("correctness" for some value of correct), or leave it at 100 ("convenience" for some value of convenient).

allisonkarlitskaya (commented):
I mean, ya, okay... that's fine, but maybe drop the validity tweak.

...but maybe let's talk about it being time to stop doing this stuff.....

images/generate-image-certs.sh
martinpitt: Good point. This is totally unimportant, I dropped the commit.

allisonkarlitskaya (approved):

PR #446, opened 2021-11-30
images: Drop image service on CentOS CI

martinpitt:
Linode stores have proven to be very reliable, and we still have
redundancy on cockpit-11. Let's tear down the image server on CentOS CI
to reduce usage of our custom cert/key and upload token, and save some
bandwidth/time on image refreshes (which upload to all stores).

https://github.com/cockpit-project/bots/pull/2677 drops it as an image
store.

As images are stored on a shared persistent volume in CentOS CI, tasks
containers continue to use this as a cache -- the first tasks container
which needs an image will download it to the cache, and from then on all
others have it available already. We don't need the images container for
that.

 - https://github.com/cockpit-project/bots/pull/2677


martinpitt:
I ran `oc delete -f images/cockpit-images.yaml` and https://images-frontdoor.apps.ocp.ci.centos.org/ is gone.

marusak (approved):

PR #447, opened 2022-01-04
tasks: Go back to standard distro firefox and chromium packages

martinpitt:
Fedora's firefox packages have shipped DevTools support for a while, so
we can go back to testing stable versions instead of nightly ones.

Both of the previous Chromium crashes are fixed in current versions.

----

I tested this in my toolbox, with all four combinations of {chromium,firefox} x {headless,show}. I also ensured that removing the chromium package and only keeping chromium-headless works.

So let's build and roll out a container, we haven't done this in a while.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -1,19 +1,17 @@
 FROM fedora:35
 LABEL maintainer='cockpit-devel@lists.fedorahosted.org'
 
-# HACK: chromium-headless 88 crashes with some keyDown commands: https://bugs.chromium.org/p/chromium/issues/detail?id=1170634, install full chromium for the time being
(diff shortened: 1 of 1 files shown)


martinpitt:
I [rebuilt](https://github.com/cockpit-project/cockpituous/actions/runs/1652661737) and rolled out the new container. https://github.com/cockpit-project/cockpit/pull/16718 tests use the new one, and it has tests on CentOS CI and AWS, chromium and firefox, mobile and desktop. There is some probability of pixel test changes, let's wait with landing this until everything is sorted out.

martinpitt:
New container works nicely, ready to land.

marusak (approved):
oh nice! Great to see these hacks to be gone :tada: 
(I assume you ran both chrome and firefox tests on mobile and desktop)

PR #449, opened 2022-01-05
tasks: add anaconda project's build dependencies to container

KKoukiou:
flatpak implicit build dependency, but if not installed before,
the builddep command fails to resolve it, (at least on the local toolbox
                                           cockpit container)

I tried using builddep command, but it fails resolving some of the
systemd nested deps (systemd is a build dependency of anaconda), but
since systemd is already installed, let's not try to install it again.


KKoukiou:
I tested this `dnf` command added here in a local toolbox based on the cockpit task container and it works fine.

KKoukiou:
Will need to reopen from the same repo intead of my fork.

PR #450, opened 2022-01-06
tasks: add anaconda project's build dependencies to container

KKoukiou:
flatpak implicit build dependency, but if not installed before,
the builddep command fails to resolve it, (at least on the local toolbox
                                           cockpit container)

I tried using builddep command, but it fails resolving some of the
systemd nested deps (systemd is a build dependency of anaconda), but
since systemd is already installed, let's not try to install it again.


marusak:
@KKoukiou has [built the image](https://github.com/cockpit-project/cockpituous/actions/runs/1662599367) and I deployed it on CentosCI and e2e. I am now running tests on https://github.com/cockpit-project/cockpit/pull/16791

marusak (approved):
I ran firefox/chrome/mobile test on both centosci and e2e and all seem fine. Also https://github.com/M4rtinK/anaconda/pull/15 works now.

PR #452, opened 2022-02-11
Add a slowest tests panel to CI dashboard

jelly:
The panel is a bar gauge which lists the slowest tests overall as
filtering is currently not possible. For filtering per project to work
we need to introduce a new metric which has a label for every project we
have. As label_values(top_failures_pct, project) is not the same as
label_values(top_slowest_tests, project).

The second issue is that a bar gauge can't be sorted correctly this is a
Grafana limitation. https://github.com/grafana/grafana/issues/17245

---

The dashboard now looks like 
image

As a follow up we should introduce a new metric (number of tests maybe?) which would have a label for every project we test.

### DIFF
--- a/metrics/cockpit-ci.json
+++ b/metrics/cockpit-ci.json
@@ -8,32 +8,36 @@
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
         "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
         "type": "dashboard"
       }
     ]
   },
   "editable": true,
-  "gnetId": null,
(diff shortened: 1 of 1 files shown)


martinpitt:
Nice! However, I tried to deploy that, and it failed with

    Failed to upgrade legacy queries Datasource ${DS_PROMETHEUS} was not found

There was also a second error message, but both of them went away too quickly, I could only copy&paste the first one.

martinpitt:
Thanks! I deployed this to https://grafana-frontdoor.apps.ocp.ci.centos.org/d/ci/cockpit-ci?orgId=1 . The project selector has no influence on that, is that a known limitation for now?

martinpitt (approved):
Thanks. Let's merge this for now to reflect current reality, and we can always iterate.

PR #455, opened 2022-02-13
tasks: Update to split libvirt daemon, wait for process

martinpitt:
Commit db56f3bda introduced the killing of libvirtd after every test
run, to ensure to start from a clean slate and avoid race conditions.
Howver, in Fedora 35 (commit 864b51b06d), libvirtd got split into
multiple components.

Adjust to `virtqemud` and actually wait until the process is gone, to
avoid a race condition with accessing the socket.

---

We are currently [getting](https://logs.cockpit-project.org/logs/pull-16884-20220128-125225-d023f048-fedora-35-mobile/log) this test failure quite often:

    libvirt: XML-RPC error : Failed to connect socket to '/work/.cache/libvirt/virtqemud-sock': No such file or directory

I have no idea why that happens. However, our original approach in commit db56f3bda stopped working with the upgrade to Fedora 35, and it was racy anyway.

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -73,7 +73,8 @@ for i in $(seq 1 30); do
 
     # avoid stale state and sockets
     virsh list --id --all | xargs --no-run-if-empty virsh destroy
-    pkill libvirtd || true
+    pkill virtqemud || true
+    while pgrep virtqemud >/dev/null; do sleep 0.5; done
 
     # run-queue fails on empty queues; don't poll too often
     timeout 12h ./run-queue ${AMQP_SERVER:+--amqp} ${AMQP_SERVER:-} || slumber


jelly (approved):

martinpitt:
I [triggered an image refresh](https://github.com/cockpit-project/cockpituous/actions/runs/1841146597) and will test it locally in toolbox for a bit, to make sure browsers are not completely broken. Then I'll roll it out onto our infra this evening or tomorrow early morning, to avoid distrupting running tests.

martinpitt:
I locally ran some non-trivial c-machines tests against the updated container with both firefox and chromium, and they both run mostly fine. *However*, there is quite some noise on pixel tests due to the updated chromium. So we need to deploy this in lockstep with pixel updates.

martinpitt:
I reinstated the tag, and sent PRs to update pixels for [cockpit](https://github.com/cockpit-project/cockpit/pull/17013), [c-machines](https://github.com/cockpit-project/cockpit-machines/pull/573), and [c-podman](https://github.com/cockpit-project/cockpit-podman/pull/919)

PR #461, opened 2022-02-23
ansible: Use Fedora CoreOS for webhook, simplify webhook launch

martinpitt:

This will make it easier and *much* faster to set up a fallback webhook, like we needed yesterday.

I ran the playbook as documented, and with no further changes I can see the queues on  http://ec2-3-228-126-27.compute-1.amazonaws.com/inspect-queue

This works really well, I'll convert the tasks runner to FCOS as the next step. I also want to convert the sink, but that requires some additional care as our logs are currently not on a separate volume (argh sins of the past), so they need to be copied first.


martinpitt:
I deleted the current instance again, as we don't currently need it. Of course, feel free to run the playbook yourself to get a feel  for it, and then please delete the instance.

marusak (dismissed):
danke!

ansible/aws/tasks/launch-coreos.yml
```diff
@@ -0,0 +1,57 @@
...
+
+- name: Create EC2 instance
+  ec2:
+    key_name: "{{ aws_key_name }}"
+    region: "{{ aws_region }}"
+    image: "{{ ami_coreos.image_id }}"
+    instance_type: t2.small
+    monitoring: true
+    # persistent network device with stable external IP 3.228.126.27
+    network_interfaces: eni-004f5b4f714f3fda9
```
martinpitt: This does not belong here, move to parameter

ansible/aws/tasks/launch-coreos.yml
```diff
@@ -0,0 +1,57 @@
...
+- name: Get newest AMI
+  set_fact:
+    ami_coreos: "{{ ami_list.images | sort(attribute='creation_date') | last }}"
+
+- name: Create EC2 instance
+  ec2:
+    key_name: "{{ aws_key_name }}"
+    region: "{{ aws_region }}"
+    image: "{{ ami_coreos.image_id }}"
+    instance_type: t2.small
```
martinpitt, line 26: Needs to be parameterized.

martinpitt (commented):
Let me dwell on this a little more

ansible/aws/tasks/launch-coreos.yml
martinpitt: Fixed

ansible/aws/tasks/launch-coreos.yml
martinpitt: Fixed.

martinpitt:
Fixed, re-tested, happier now.

martinpitt:
The parameterization here is good enough now. I sent a follow-up PR #462 that uses this for tasks. It needs a few extensions, but these can happen over there.

ansible/roles/webhook/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+        y = yaml.load(f)
+    files = [item for item in y["items"] if item["metadata"]["name"] == "amqp-config"][0]["data"]
+    for name, contents in files.items():
+        with open(os.path.join('/etc/rabbitmq', name), 'w') as f:
+            f.write(contents)
+    EOF
+
+- name: Clean up existing pod
+  shell: |
+    podman pod rm -f cockpituous || true
```
mh21, line 26: I learned --replace from https://www.redhat.com/sysadmin/podman-features-1 😄

martinpitt: Yes, me too -- and the `-l/--last` thing is even better. However, it seems much less clear to me how `--replace` acts for pods, which are "additive" by nature (unlike containers). For scripts it still seems cleaner to me to first clean up the whole old pod rather than trying to replace individual containers in a running one.

marusak (approved):

PR #462, opened 2022-02-23
ansible: Use Fedora CoreOS for tasks

martinpitt:
Similarly to commit 9cccfd9, move launch-tasks.yml to launch-coreos.

Adjust the device name, as the root volume is now called /dev/xvda instead of
/dev/sda1 with CoreOS and the new ec2_instance API.

---

 - Builds on top of PR #461 

I launched this, it started and immediately grabbed some work, for example [this](https://logs.cockpit-project.org/logs/pull-17035-20220223-084238-42d9c17e-fedora-35-devel/log.html) or [this run](https://logs.cockpit-project.org/logs/pull-17035-20220223-084238-42d9c17e-fedora-35-devel/log.html)

### DIFF
--- a/ansible/aws/launch-tasks.yml
+++ b/ansible/aws/launch-tasks.yml
@@ -5,51 +5,28 @@
   vars_files: aws_defaults.yml
 
   tasks:
-    - name: Create EC2 instance
-      ec2:
-        key_name: "{{ aws_key_name }}"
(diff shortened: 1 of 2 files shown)


martinpitt:
I validated that deploying the webhook still works with this changes.

marusak (approved):
Danke

PR #463, opened 2022-02-24
sink: Use separate volume, move to Fedora CoreOS

martinpitt:
This is rolled out, and seems to work fine. E.g. [this test](https://logs.cockpit-project.org/logs/pull-17038-20220224-094403-b4e3bd1f-rhel-8-6/log.html) was just started, and is logging to the new sink.

I created the separate volume this early morning (while no test was running) and copied over everything. So from now on we can upgrade/redeploy/hack on the sink without losing logs. While I was at it, I also increased the log storage size from 40 to 100 GB, just in case we ever want to store core dumps again :wink:


ansible/roles/sink-ssh-gateway/tasks/main.yml
```diff
@@ -1,7 +1,4 @@
 ---
-- hosts: tag_Name_cockpit_sink
-  tasks:
-
   - name: Create sshgw user
     user:
       name: sshgw
```
martinpitt, line 4: Oops, I should have outdented this file for consistency. It apparently seems to work anyway, though.

martinpitt: Fixed

marusak (approved):
Danke :)

PR #466, opened 2022-02-25
Set explicit resource requests/limits for our k8s pods

martinpitt:
sink and webhook did not set any resources, but were relying on the
cluster defaults which are beyond our control.

Set explicit CPU request for the task container, to match the
TEST_JOBS=5 parallelism. Also request a minimum of 16 GiB RAM, to help
k8s with the node allocation.

See https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/

----

With a hat tip to @inakimalerba for explaining this on the cyborg infra workshop! :tophat:


martinpitt:
I redeployed our CentOS CI resources with these new yamls.

sink/sink-centosci.yaml
```diff
@@ -39,6 +39,13 @@ items:
             - name: config
               mountPath: /run/config
               readOnly: true
+            resources:
+              limits:
+                memory: 1G
+                cpu: 2
+              requests:
+                memory: 200Mi
+                cpu: "100m"
```
marusak, line 48: I must say, that I love how 3 values with units are all defined differently. One is in 10 base, one in 2 base and another one is string in `"` :)

marusak (approved):
OOI where do these number come from? Why is it for example `300Mi`?

sink/sink-centosci.yaml
martinpitt: You are right, I cleaned that up.

martinpitt:
@marusak: Memory requirements for the tasks container actually have some science behind it: (VM RAM + browser) x TEST_JOBS plus a little spare room. We have used the same limit on our e2e machines (see install-service.sh).

For webhook and sink I looked at OpenShift's metrics page, added some spare room, plus some educated guessing. These may well need a little adjustment in the future, but now at least we have some defined limit.

marusak (approved):
Danke :)

PR #468, opened 2022-03-02
tasks: Add lcov

mvollmer:
So that we can generate coverage reports during CI.


martinpitt:
Please send this from a branch on origin, so that the "tasks" test can run properly. That doesn't work from forks. Thanks!

PR #469, opened 2022-03-02
tasks: Add lcov

mvollmer:
So that we can generate coverage reports during CI.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -19,6 +19,7 @@ RUN dnf -y update && \
         gnupg \
         intltool \
         jq \
+        lcov \
         libappstream-glib \
         libvirt-daemon-kvm \
         libvirt-client \


martinpitt (approved):
Thanks! I'll trigger a container rebuild. I'll roll this out late tonight, to avoid interrupting everyone's tests all the time, okay?

There's always the chance of a new browser breaking something, so we need to watch out.

martinpitt:
[container refresh workflow](https://github.com/cockpit-project/cockpituous/actions/runs/1923327795)

FTR, if I forget, the bots will also pull the new image after the "30x tasks" iteration cycle (which usually takes a day or less) Maybe let's just do that this time, unless you are very anxious to see this life.

mvollmer:
> Maybe let's just do that this time, unless you are very anxious to see this life.

Let's wait, thanks!

martinpitt:
Argh, [failed](https://github.com/cockpit-project/cockpituous/runs/5394619718?check_suite_focus=true)! That's my stupidity, sending a fix

PR #471, opened 2022-03-03
tasks: Kill virtqemud harder, fix image pull failure on RHEL 8

martinpitt:
Several of our tasks containers got stuck in an endless pgrep/sleep
loop, as the running virtqemud ignored SIGTERM and even SIGQUIT.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -7,16 +7,18 @@ jobs:
     runs-on: ubuntu-latest
     environment: quay.io
     timeout-minutes: 30
+    # needs to use docker, as podman-built images are not accepted to RHEL 8's docker any more
+    env:
+      DOCKER: docker
     steps:
       - name: Clone repository
         uses: actions/checkout@v2
 
-      # use podman as root, so that it uses overlayfs; user with vfs takes too long and uses too much space
       - name: Log into container registry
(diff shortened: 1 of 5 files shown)


martinpitt:
Also, trying to pull yesterday's refresh fails spectacularly:
```
# docker pull quay.io/cockpit/tasks
Using default tag: latest
Trying to pull repository quay.io/cockpit/tasks ... 
missing signature key
```

I'm not sure if that is a problem of the new build, or quay.io has an outage, etc. I'll try with the [new build](https://github.com/cockpit-project/cockpituous/actions/runs/1926427681) that I just triggered for this branch, if it fails as well I'll try to set back the tag to the version before yesterday (without lcov)

martinpitt:
The new build fails the same way. `podman pull` works, but `docker pull` on RHEL 8.5 fails. `docker pull quay.io/cockpit/tasks:2022-02-14` still works, so this is somehow a problem with new container builds. This needs a workaround, but for now I moved :last back to :2022-02-14

martinpitt:
New container built, and I confirm that I can pull it both on RHEL 8 with docker and on Fedora 35 with podman. I am now running it in toolbox. As bots are still catching up from last night's e2e outage, I don't want to interrupt them now. But they will auto-update to the new one over the day.

In the meantime, I'll run a few tests locally with firefox and chromium and will yell if anything goes wrong.

marusak (approved):

PR #473, opened 2022-03-14
tasks: Avoid unnecessary qemu-kvm desktop/storage/network dependencies

martinpitt:
libvirt-daemon-kvm is a meta-package that pulls in the qemu-kvm
metapackage (which has tons of desktop-only dependencies), plus all of
the storage/network libvirt drivers which we also don't need for
testing. This reduces the image size from 2.55 to 2.22 GB.

See https://libvirt.org/kbase/rpm-deployment.html

-----

 - Depends on https://github.com/cockpit-project/bots/pull/3064

I locally built a container from this and ran a Firefox and Chromium test.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -21,7 +21,7 @@ RUN dnf -y update && \
         jq \
         lcov \
         libappstream-glib \
-        libvirt-daemon-kvm \
+        libvirt-daemon-driver-qemu \
         libvirt-client \
         libvirt-python3 \
         libXt \
@@ -37,6 +37,7 @@ RUN dnf -y update && \
         python3-pycodestyle \
         python3-pika \
         python3-pillow \
+        qemu-kvm-core \
         rpm-build \
         rpmdevtools \
         rsync \


martinpitt:
I [rebuilt the container](https://github.com/cockpit-project/cockpituous/actions/runs/1980170785), it'll roll itself out over the next day.

berrange:
Beware that in using qemu-kvm-core you are leaving out potentially quite a lot of potentially very useful features  - eg all the audio backends, USB passthrough, virtio-gpu display devices which should be the default for modern OS.  As QEMU grows more modules, qemu-kvm-core will have increasingly fewer features. So be warned that this can make your setup less reliable, as we might modularize something you are relying on existing, causing a regression.

martinpitt:
Thanks @berrange for the warning! I'm aware of that, but we really test stuff in "server mode". If we ever encounter a test which wants to add some audio etc. hardware, we can just selectively add the dependencies.

martinpitt:
This is rolled out now. It required updating some pixel tests in https://github.com/cockpit-project/cockpit/pull/17138 , but otherwise behaves well.

marusak (approved):

PR #475, opened 2022-03-31
ansible: Set up DNS caching on e2e machines

martinpitt:
The e2e network's DNS server has been rather unreliable recently. Set up
a local dnsmasq to mitigate this.

Force TTLs to be at least one hour.

---

I rolled this out. Let's give it a day or two to see whether it makes a difference.


martinpitt:
Positive data point: The services image refresh in https://github.com/cockpit-project/bots/pull/3153 ran through at last.

jelly (approved):

PR #477, opened 2022-04-12
tasks: Introduce elastic cloud runner mode

martinpitt:
Let the cockpit-tasks main runner keep track of whether `run-queue` ever
succeeded (i.e. picked something from the queue and ran it), and
otherwise exit with 100. Tell the systemd unit that 100 is a successful
exit code, to avoid confusing log noise.

Let all cockpit-tasks@*.service run in a specific (instead of the
automatic anonymous) `idle-poweroff.slice`, so that we can react to all
tasks containers shutting down.

When calling the installation script with `$IDLE_POWEROFF`, configure
the slice to automatically power off the machine once all cockpit-tasks
instances exited cleanly (we don't want this on failures, so that we can
ssh in and examine them). Use the `poweroff-immediate` heavy hammer
there, to avoid potential hangs on shutdown -- there is nothing to
rescue from the instance anyway.

Plumb that through the AWS Ansible role and document it.

----

 - [x] [Refresh container](https://github.com/cockpit-project/cockpituous/actions/runs/2153138538) and deploy it to infra
 - [x] Test refreshed container: [initial run](https://logs.cockpit-project.org/logs/pull-17236-20220412-072642-ad7de160-arch/log.html), more thoroughly in https://github.com/cockpit-project/cockpit/pull/17238
 - [x] Work around `podman --device` regression: https://bugzilla.redhat.com/show_bug.cgi?id=2074402
 - [x] Full end-to-end test
 - [ ] Rework idle detection work to cooperate across containers
 - Requires #478 
 - Follow-up: Investigate setting up an auto-terminate cloud alarm, to shorten the 15 mins it takes EC2 to recognize the shutdown

### DIFF
--- a/ansible/aws/README.md
+++ b/ansible/aws/README.md
@@ -59,6 +59,8 @@ Create and configure the instance:
 
 If you run more than one at a time, set a custom host name with `-e hostname=cockpit-aws-tasks-2` or similar, so that GitHub test statuses remain useful to identify where a test runs.
 
+There is also an "elastic" mode where the tasks bots keep running until they all run out of work (i.e. all iterations had an empty queue). Use that for situations where AWS instances act as extra high-demand capacity instead of being the primary runners. Enable that mode with `-e idle_poweroff=1`.
+
 Public log sink/server setup
 ----------------------------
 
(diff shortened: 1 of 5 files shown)


martinpitt:
I don't have motivation to work on this anytime soon. We don't currently have much elastic infrastructure, except for PSI -- and even there running idle workers isn't very expensive. Let's shelve this for now.

PR #478, opened 2022-04-12
ansible: Terminate AWS instances on shutdown

martinpitt:
This avoids hidden costs or accidentally blocked volume/IP resources,
and allows us to move to an elastic model especially for the tasks
runners -- then they can decide on their own for how long they want to
run.

---

Split out from PR #477 -- I did comprehensive testing there.


allisonkarlitskaya (approved):
I'm going to pretend like I understand what's going on here.  Sounds nice :)

PR #479, opened 2022-04-12
tasks: Add hack for podman --device regression

martinpitt:
In recent versions, `podman run --device=/dev/kvm` does not work any
more, it fails access with "Operation not permitted". Run with
`--privileged` for the time being to work around that.

---

I saw this as part of testing #477, but it's completely independent. We need this case we need to spin up AWS tasks for real due to some e2e outage.

### DIFF
--- a/tasks/install-service
+++ b/tasks/install-service
@@ -17,6 +17,8 @@ systemctl stop 'cockpit-tasks@*.service'
 if RUNC=$(which podman 2>/dev/null); then
     UNIT_DEPS=''
     DEVICES="--device=/dev/kvm"
+    # HACK: https://bugzilla.redhat.com/show_bug.cgi?id=2074402
+    DEVICES="$DEVICES --privileged"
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (commented):
I have to admit I'm not a big fan of this change.

Can we just hold off on updating the container until podman gets their act together?

martinpitt:
@allisonkarlitskaya It's *not* about the cockpit/tasks container -- this regression affects podman on the host. We could find and pin down an earlier Fedora CoreOS build (before mid March). But I suppose I can also just keep this PR in stock and land it when we have a desperate need to bring up an AWS runner.

FTR, I just tested the fix half an hour ago, so one of the next coreos nightlies should pick it up. With that, this PR isn't important any more, but I didn't know that yet this morning.

martinpitt:
https://bodhi.fedoraproject.org/updates/FEDORA-2022-2067702f06 needs to land. It's stuck for some reason, but as soon as that goes in, we can drop this.

PR #480, opened 2022-04-13
tasks: Avoid test interruption when deploying a new container

martinpitt:
Our cockpit-tasks script already listens to SIGTERM for a graceful
shutdown, i.e. finish the current task and then stop the main loop.

Make use of this when we deploy a new container, instead of bluntly
killing them:

 - For OpenShift, `oc delete` already SIGTERMs the container, but the
   default grace period of 30s is way too short. Bump it to one hour,
   which is our agreed-upon maximum test runtime.

 - For systemd, signal cockpit-tasks instances (which we can see from
   the host)

There is no big downside: After running the deploy script, the existing
running containers will not de-queue new tasks, i. e. all
subsequent tests will run on the new image. But we stop interrupting
running tests, having to re-start them and running our test stats.

----

I tested this both on OpenShift (config change rolled out already) and on e2e. Given how simple this is, why didn't we do this years ago already? :shrug:

### DIFF
--- a/ansible/maintenance/deploy-tasks-container.yml
+++ b/ansible/maintenance/deploy-tasks-container.yml
@@ -20,5 +20,5 @@
     # avoid docker_image module, that requires python-docker
     command: docker pull quay.io/cockpit/tasks
 
-  - name: Restart systemd controlled tasks containers
-    command: systemctl restart cockpit-tasks@*
+  - name: Tell tasks containers to drain and restart
+    command: pkill -ex cockpit-tasks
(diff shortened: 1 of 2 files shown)


martinpitt:
See https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace and https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination for details.

marusak (approved):
Thank you!

PR #482, opened 2022-04-26
tasks: npm install robustification

martinpitt:
We are currently getting an awful lot of [npm install failures](https://logs.cockpit-project.org/logs/pull-3277-20220426-052741-a5434ecf-fedora-rawhide-boot-rhinstaller-anaconda/log.html) on our CI. Let's try to improve this.

 - [x] Deploy [tasks refresh](https://github.com/cockpit-project/cockpituous/actions/runs/2224947474)
 - [x] test run with new container: https://github.com/cockpit-project/bots/pull/3299 (including lots of `npm install` runs wit chromium), and a [cockpit firefox run](https://logs.cockpit-project.org/logs/pull-17281-20220426-071312-9cfbfa09-fedora-35-firefox/log.html)

### DIFF
--- a/tasks/cockpit-tasks
+++ b/tasks/cockpit-tasks
@@ -21,6 +21,10 @@ if [ -r /secrets/npm-registry.crt ]; then
     export NODE_EXTRA_CA_CERTS=/secrets/npm-registry.crt
 fi
 
+# prone to timeouts and errors with lots of parallel containers
+npm config set fetch-retries 6
+npm config set fetch-timeout 600000
+
 # set up S3 keys for OpenShift secrets volume
 if [ ! -d /secrets/s3-keys ]; then
     # then our container symlink will point into the void, replace it with a directory and set up all files that we can find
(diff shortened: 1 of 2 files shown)


tasks/cockpit-tasks
```diff
@@ -21,6 +21,10 @@ if [ -r /secrets/npm-registry.crt ]; then
     export NODE_EXTRA_CA_CERTS=/secrets/npm-registry.crt
 fi
 
+# prone to timeouts and errors with lots of parallel containers
+npm config set fetch-retries 6
+npm config set fetch-timeout 600000
```
jelly, line 26: ah, that's milliseconds, so one minute? And interesting, it's lower then the wopping 5 minutes! default.

jelly (approved):

tasks/cockpit-tasks
martinpitt: Right, see https://docs.npmjs.com/cli/v7/using-npm/config#fetch-timeout

PR #485, opened 2022-05-12
ansible: Enable ntpd on e2e servers

martinpitt:
It was not running on several servers, which caused significant clock
drift. This caused S3 requests to fail, as request signature validity
is time dependent.

---

That is what caused these [RHEL image download errors](https://logs.cockpit-project.org/logs/pull-17334-20220512-071017-8ff38376-rhel-8-6/log.html)


allisonkarlitskaya (approved):
Thanks!

PR #487, opened 2022-05-18
tasks: Test image upload/download with a fake image instead of --state, and with S3 store

martinpitt:
`image-upload --state` does not currently work with S3 stores, as the
checksum is not part of the file name. We can and should fix this later,
but for now, test what an actual image upload would look like. Fake a
"testimage" for that.

Set up a https://min.io/ S3 container, and use it to test image
uploading and downloading with an S3 store.

---

Need to go via the bots here, as locally this doesn't work for me. The upload fails with code 500, and the images container says
```
[crit] 4#4: *3 open() "/cache/images/0000000002" failed (75: Value too large for defined data type), client: 10.0.2.100, server: cockpit-tasks, request: "PUT /testimage-e258d248fda94c63753607f7c4494ee0fcbe92f1a76bfdac795c9d84101eb317.qcow2 HTTP/1.1", host: "cockpituous:8443"
```

This is something weird between file systems, podman, etc. I already tried to locally rebuild the images container with Fedora 35, but that didn't help. I don't want to debug this right now.

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -189,12 +189,18 @@ test_image() {
             sleep 5
         done
 
-        # test image-upload
         cd bots
-        echo world  > /cache/images/hello.txt
-        ./image-upload --store https://cockpituous:8443 --state hello.txt
+
+        # fake an image
+        echo world  > /cache/images/testimage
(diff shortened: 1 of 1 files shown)


martinpitt:
Nice, [this works](https://github.com/cockpit-project/cockpituous/runs/6488893014?check_suite_focus=true). Now adding S3..

PR #488, opened 2022-05-18
tasks: Switch run-local.sh to testing logs in S3

martinpitt:
https://github.com/cockpit-project/bots/pull/3403 switches tests-scan
over to log to S3. Adjust run-local.sh's test_pr() for that. Create a
separate "logs" S3 bucket to mimic what happens in production.
    
Drop checking the "status" file, as the S3 stream does not do this.

-----

This should land shortly after https://github.com/cockpit-project/bots/pull/3403, and then without the XXX commit. But let's get this working first.

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -219,6 +219,8 @@ test_image() {
 
         # test image-upload to S3
         ./image-upload --store http://localhost.localdomain:9000/images/ testimage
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Thanks.  This stuff is deeply cool :)

PR #490, opened 2022-05-21
e2e S3 sink for production

martinpitt:
Configure/test `run-local.sh`'s S3 store to be suitable for production, and add scripts/Ansible to set up a production S3 on cockpit-11. I already deployed that. However, to use it, we still need some bots changes.

We also still need to keep the old images container until we move image refresh logs to S3 as well. After that, the prize money is [this commit](https://github.com/cockpit-project/cockpituous/commit/af43988407951b64e3da03e670bccf12f5148461) which removes all the images/sink stuff.

 - [x]  Fix bots to use our CA for self-hosted S3 stores, and declare S3 store on cockpit-11: https://github.com/cockpit-project/bots/pull/3417

~After that lands, I'll drop the XXX commit and squash the fixup.~ **done**

### DIFF
--- a/images/install-s3-service
+++ b/images/install-s3-service
@@ -1,0 +1,60 @@
+#!/bin/sh
+
+# deploy this script with:
+# ansible -i inventory -m include_role -a name=s3-systemd e2e_s3
+set -eufx
+
+CACHE=/var/cache/cockpit-tasks
+SECRETS=/var/lib/cockpit-secrets
+
+if RUNC=$(which podman 2>/dev/null); then
+    UNIT_DEPS=''
+else
+    RUNC=$(which docker)
+    UNIT_DEPS="Requires=docker.service
+After=docker.service"
+fi
+
+systemctl stop cockpit-s3.service || true
+
+cat <<EOF > /etc/systemd/system/cockpit-s3.service
+[Unit]
+Description=Cockpit S3 image server
+$UNIT_DEPS
+
+[Service]
+Restart=always
+RestartSec=60
+ExecStartPre=-$RUNC rm -f cockpit-s3
+ExecStartPre=/bin/rm -rf $CACHE/.minio.sys/
+ExecStartPre=/bin/sh -c 'dd if=/dev/urandom bs=12 count=1 status=none | base64 > /run/cockpit-s3.rootpassword.txt'
+ExecStart=/bin/sh -xc "$RUNC run --name=cockpit-s3 --publish=443:9000 -e MINIO_ROOT_USER=minioadmin -e MINIO_ROOT_PASSWORD=\$(cat /run/cockpit-s3.rootpassword.txt) -v $SECRETS/tasks/server.key:/root/.minio/certs/private.key:ro -v $SECRETS/tasks/server.pem:/root/.minio/certs/public.crt:ro -v $CACHE:/data:rw quay.io/minio/minio server /data"
+ExecStartPost=/usr/local/lib/setup-s3.sh
+ExecStartPost=/bin/rm -f /run/cockpit-s3.rootpassword.txt
+ExecStop=$RUNC rm -f cockpit-s3
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+cat <<EOF > /usr/local/lib/setup-s3.sh
+#!/bin/sh
+set -eu
+read s3user s3key < "$SECRETS/tasks/s3-keys/e2e.bos.redhat.com"
+$RUNC run --interactive --rm --network=host \
+    -v "$SECRETS"/webhook/ca.pem:/etc/pki/ca-trust/source/anchors/ca.pem:ro \
+    --entrypoint /bin/sh quay.io/minio/mc <<EOC
+set -e
+update-ca-trust
+until mc alias set minio https://localhost minioadmin \$(cat /run/cockpit-s3.rootpassword.txt); do sleep 1; done
+# this fails with non-zero as the directory already exists; that's fine
+mc mb minio/images || true
+mc policy set download minio/images
+mc admin user add minio/ '\$s3user' '\$s3key'
+mc admin policy set minio/ readwrite user='\$s3user'
+EOC
+EOF
+chmod a+x /usr/local/lib/setup-s3.sh
+
+systemctl daemon-reload
+systemctl enable --now cockpit-s3.service
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -12,6 +12,11 @@ SECRETS=$DATADIR/secrets
 IMAGES=$DATADIR/images
 IMAGE_PORT=${IMAGE_PORT:-8080}
 S3_PORT=${S3_PORT:-9000}
+# S3 address from inside cockpituous pod
+S3_URL_POD=http://localhost.localdomain:9000
+# S3 address from host
+S3_URL_HOST=http://localhost.localdomain:$S3_PORT
+IMAGES_URL_POD=https://cockpituous:8443
 
 # CLI option defaults/values
 PR=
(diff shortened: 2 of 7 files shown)


martinpitt:
Meh, [using cockpituous.pod hostname](https://github.com/cockpit-project/cockpituous/pull/490/commits/101efb2f0e08ae1632c9c20fff3f06a94b4bebe5) to address the S3 store would have been nice, and works locally. But the `podman run --add-host` option [fails in the workflow](https://github.com/cockpit-project/cockpituous/runs/6536393659?check_suite_focus=true). So back to `localhost.localdomain`. :man_shrugging:

tasks/run-local.sh
```diff
@@ -209,7 +209,7 @@ cleanup_containers() {
 
 test_image() {
     # test image upload
-    podman exec -i cockpituous-tasks timeout 30 sh -ec '
+    podman exec -i cockpituous-tasks timeout 30 sh -exc '
```
allisonkarlitskaya, line 185: `u` for good measure?

tasks/run-local.sh
```diff
@@ -12,6 +12,9 @@ SECRETS=$DATADIR/secrets
 IMAGES=$DATADIR/images
 IMAGE_PORT=${IMAGE_PORT:-8080}
 S3_PORT=${S3_PORT:-9000}
+S3_URL_POD=http://localhost.localdomain:9000
```
allisonkarlitskaya: This could use a comment.  It took me a fair bit of thinking before I realised what the point of having two of these was.

tasks/run-local.sh
```diff
@@ -281,19 +288,20 @@ test_pr() {
     ./inspect-queue --amqp localhost:5671;"
 
     LOGS_URL="$S3_URL_HOST/logs/"
+    CURL="curl --cacert $SECRETS/ca.pem --silent --fail --show-error"
 
     # wait until the unit-test got run and published, i.e. until the non-chunked raw log file exists
     for retry in $(seq 60); do
-        LOG_MATCH="$(curl --silent --show-error $LOGS_URL| grep -o "pull-${PR}-[[:alnum:]-]*-unit-tests/log<")" && break
+        LOG_MATCH="$($CURL $LOGS_URL| grep -o "pull-${PR}-[[:alnum:]-]*-unit-tests/log<")" && break
```
allisonkarlitskaya, line 239: This is begging for a `python3 -m lib.s3 get` subcommand or so.

allisonkarlitskaya (approved):
Presumably there's a `ci-secrets` commit somewhere as well? :)

I mostly glanced over this.

Minor stuff below.  You could also implement the suggestions as followups (or not).

tasks/run-local.sh
martinpitt: You mean `ls`? But the next few usages of `$CURL` would be a `get` indeed. I'd keep this as a follow-up though, as it needs changes to bots.

tasks/run-local.sh
martinpitt: Ack, added a comment.

tasks/run-local.sh
martinpitt: Good point, added as separate commit.

images/install-s3-service
```diff
@@ -0,0 +1,60 @@
...
+    --entrypoint /bin/sh quay.io/minio/mc <<EOC
+set -e
+update-ca-trust
+until mc alias set minio https://localhost minioadmin \$(cat /run/cockpit-s3.rootpassword.txt); do sleep 1; done
+# this fails with non-zero as the directory already exists; that's fine
+mc mb minio/images || true
+mc policy set download minio/images
+mc admin user add minio/ '\$s3user' '\$s3key'
+mc admin policy set minio/ readwrite user='\$s3user'
+EOC
```
allisonkarlitskaya, line 55: I think this is the first time I've seen this.

images/install-s3-service
```diff
@@ -0,0 +1,60 @@
...
+[Unit]
+Description=Cockpit S3 image server
+$UNIT_DEPS
+
+[Service]
+Restart=always
+RestartSec=60
+ExecStartPre=-$RUNC rm -f cockpit-s3
+ExecStartPre=/bin/rm -rf $CACHE/.minio.sys/
+ExecStart=/bin/sh -xc "$RUNC run --name=cockpit-s3 --publish=443:9000 -e MINIO_ROOT_USER=minioadmin -e MINIO_ROOT_PASSWORD=\$(cat /run/cockpit-s3.rootpassword.txt) -v $SECRETS/tasks/server.key:/root/.minio/certs/private.key:ro -v $SECRETS/tasks/server.pem:/root/.minio/certs/public.crt:ro -v $CACHE:/data:rw quay.io/minio/minio server /data"
```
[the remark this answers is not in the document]
allisonkarlitskaya: Was it your intention to capture the password in the unit file?  I guess in that case you shouldn't escape the `$`.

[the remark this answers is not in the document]
martinpitt: We don't even publish the web console port, or even set it (it's going to be a random one). To make this cleaner, I can add an ExecStartPost= which rm's the password file, WDYT?

images/install-s3-service
martinpitt: Like, how to configure S3 user policies? Me too, if there is a more elegant way, I'm all ears. I was following https://docs.min.io/minio/baremetal/reference/minio-mc-admin/mc-admin-user.html

images/install-s3-service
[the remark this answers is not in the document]
martinpitt: Ah, I see what you mean, thanks! I'll think about that.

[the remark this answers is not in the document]
martinpitt: Fixed by moving the password generation into ExecStartPre=. I deployed this to cockpit-11, rebooted it, container starts up and https://cockpit-11.apps.cnfdb2.e2e.bos.redhat.com/images/ responds as expected.

images/install-s3-service
```diff
@@ -0,0 +1,59 @@
...
+[Unit]
+Description=Cockpit S3 image server
+$UNIT_DEPS
+
+[Service]
+Restart=always
+RestartSec=60
+ExecStartPre=-$RUNC rm -f cockpit-s3
+ExecStartPre=/bin/rm -rf $CACHE/.minio.sys/
+ExecStartPre=/bin/sh -c 'dd if=/dev/urandom bs=12 count=1 status=none | base64 > /run/cockpit-s3.rootpassword.txt'
```
allisonkarlitskaya, line 30: I'm not gonna say I like it, but at least I think it'll work :)

images/install-s3-service
allisonkarlitskaya: I mean the double-nested here-document :)

images/install-s3-service
```diff
@@ -0,0 +1,59 @@
...
+#!/bin/sh
+set -eu
+read s3user s3key < "$SECRETS/tasks/s3-keys/e2e.bos.redhat.com"
+$RUNC run --interactive --rm --network=host \
+    -v "$SECRETS"/webhook/ca.pem:/etc/pki/ca-trust/source/anchors/ca.pem:ro \
+    --entrypoint /bin/sh quay.io/minio/mc <<EOC
+set -e
+update-ca-trust
+until mc alias set minio https://localhost minioadmin \$(cat /run/cockpit-s3.rootpassword.txt); do sleep 1; done
+# this fails with non-zero as the directory already exists; that's fine
```
allisonkarlitskaya, line 49: Can you delete the password file here?

allisonkarlitskaya (changes requested):

images/install-s3-service
martinpitt: I added the deletion to the unit instead. More symmetric and robust IMHO.

allisonkarlitskaya: I agree.  I was going to propose the same, but I don't understand enough about systemd to know if the post-execs are sync or async :)

allisonkarlitskaya (approved):

PR #499, opened 2022-06-13
e2e: add "reboot" racadm fragment

allisonkarlitskaya:
I used this today.  Maybe I'll use it again. :)

For the record, these get used like this:

```
./run-fragment idrac-cockpit-9 fragments/reboot.racadm
```


martinpitt (approved):
Thanks!

PR #502, opened 2022-06-22
tasks: Use "podman network"

martinpitt:
All our tasks machines run RHEL ≥ 8.6 or Fedora CoreOS now, so we can
rely on `podman network`. Use a bridge instead of slirp for more
performance.

----

This is the approch that we had used with docker for years. But our initial EC2 instances ran RHEL 8.4 or earlier, which did not have `podman network` yet, so I resorted to slirp.

I had this sitting on my disk for 2½ years, and just spotted this again when reviewing local branches. I rolled this out to e2e, and so far it seems happy. But let's give it a day or two of field evaluation.

### DIFF
--- a/ansible/e2e/setup.yml
+++ b/ansible/e2e/setup.yml
@@ -86,9 +86,7 @@
   roles:
     - ci-data-cache
     - install-secrets-dir
-    - role: tasks-systemd
-      vars:
-        instances: 4
+    - tasks-systemd
 
 - name: Set up image server
   hosts: e2e_s3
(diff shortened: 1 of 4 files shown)


martinpitt:
https://github.com/cockpit-project/cockpit-machines/pull/721 , https://github.com/cockpit-project/bots/pull/3544 , and https://github.com/cockpit-project/cockpit/pull/17469 all ran on the new configuration, works fine.

marusak (approved):
Thanks!

PR #508, opened 2022-07-22
Move PSI instances to Fedora CoreOS, add script for mass-deployment

martinpitt:
See individual commits for details. I rolled this out, using `deploy-all.sh`.


marusak (approved):
nice, thanks!

PR #509, opened 2022-07-25
Fix webhook EC2 deployment, move primary webhook to EC2

martinpitt:
Our CentOS CI project will go away soon, so let's move to an EC2 instance as our primary webhook/queue. I deployed this already ([hello](http://ec2-3-228-126-27.compute-1.amazonaws.com/inspect-queue)), but didn't activate it yet in bots nor our projects.

 - [x] Coordinate switchover time with @ptoscano, when to move bots + sub-man project settings.
 - [x] Land https://github.com/cockpit-project/bots/pull/3660 after this

### DIFF
--- a/ansible/aws/README.md
+++ b/ansible/aws/README.md
@@ -77,17 +77,16 @@ points to the stable IP 54.89.13.31 of that instance.
 
 Webhook setup
 -------------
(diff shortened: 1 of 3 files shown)


martinpitt:
@ptoscano: So what you'll need to do here is to change sub-man and sub-man-cockpit project settings to change the webhook URL (which requires project admin). We need to coordinate that. When would be a good time for you?

PR #513, opened 2022-07-28
ansible: Set up image cache in PSI

martinpitt:
Currently, downloading images from Linode to PSI is really slow, and it
has to happen on each instance separately. To mitigate that, install a
local S3 image store on the first instance.
    
We don't have any stable IPs or DNS names there, so we have to manage
that as a dynamic "extra" image store entry. Write that to the cockpit
cache directory and bind-mount it into the tasks container as
~/.config/cockpit-dev/image-stores.

----

I rolled this out to PSI, the [image store](http://10.0.190.195/images/) is up. You need to be behind the Red Hat VPN to see this.

 - Requires support for ~/.config/cockpit-dev/image-stores and a small s3 fix in bots: https://github.com/cockpit-project/bots/pull/3670


martinpitt:
The bots support PR https://github.com/cockpit-project/bots/pull/3670 is ready for review. I fully tested image upload/download there, on the already rolled out infra.

martinpitt:
@marusak : I'd appreciate if you could test the new `ansible/openstack-ssh-config` script and see whether it works for you and is sufficiently documented.

ansible/psi/README.md
```diff
@@ -41,4 +41,6 @@ You can run [deploy-all.sh](./deploy-all.sh) to mass-deploy all instances. Exist
 
 All cloud/PSI specific parameters are in [psi_defaults.yml](./psi_defaults.yml), please edit/extend that instead of hardcoding cloud specifics in roles or playbooks.
 
-The instances run our usual [users role](../roles/users/), so if you are in the [ssh-keys](../roles/users/tasks/ssh-keys) list, you can SSH to the instances with `ssh core@10.X.X.X`.
+SSH access
+----------
+The instances run our usual [users role](../roles/users/), so if you are in the [ssh-keys](../roles/users/tasks/ssh-keys) list, you can SSH to the instances. Run [`./openstack-ssh-config`](../openstack-ssh-config) to generate an SSH configuration file from the dynamic inventory, and follow the instructions. Then you can do e. g. `ssh rhos-01-1`.
```
marusak, line 54: I followed this and it generates `/home/mmarusak/.ssh/ansible-openstack.config`. However:
```
 ssh rhos-01-1
Bad owner or permissions on /home/mmarusak/.ssh/ansible-openstack.config
```

```
$ ls -l /home/mmarusak/.ssh/ansible-openstack.config
-rw-rw-r--. 1 mmarusak mmarusak 1985 Aug  2 11:39 /home/mmarusak/.ssh/ansible-openstack.config

```

marusak:
Nice! I have not tried to deploy the cache as that is already set up. I only have one comment about wrong permissions when following the documentation. Since this step is one of those things one would do very rarely I am fine if we just write it down that you need to adjust the permissions - ideally in the output that says where I should add the one line.

ansible/psi/README.md
martinpitt: Nice find! Mine is 644, which works. I suppose your umask is "002" (mine is "022"), so new files are group writeable by default.

I updated the commit to remove group/other write permissions.

marusak (approved):
Thank you!

PR #517, opened 2022-08-08
tasks: Install flake8

martinpitt:
cockpit-machines recently moved to using flake8 by way of cockpit's
static-code. But that runs in the tasks container, and thus is currently
skipped. Let's make sure it runs again.

 - Blocks on https://github.com/cockpit-project/cockpit-machines/pull/788
 - [x] [refresh](https://github.com/cockpit-project/cockpituous/actions/runs/2816767772) and deploy
 - [x] Run cockpit and cockpit-machines chromium and firefox tests to validate new container: https://github.com/cockpit-project/bots/pull/3708

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -34,9 +34,8 @@ RUN dnf -y update && \
         origin-clients \
         psmisc \
         procps-ng \
-        python3-pyflakes \
         python3 \
-        python3-pycodestyle \
+        python3-flake8 \
         python3-pika \
         python3-pillow \
         qemu-kvm-core \


martinpitt:
Indeed [cockpit-machines is now running flake8 static tests](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3708-20220808-093654-c2d5747e-rhel-8-7-cockpit-project-cockpit-machines/log.html). They look a little odd, we should probably revert the `--tap` change or teach log.html to present that differently. But at least we won't break it again now.

marusak (approved):

PR #520, opened 2022-09-14
Replace `kind: List` k8s resource with multi-document YAML

martinpitt:
The kubernetes API does not actually have a "List" kind [1], it's just
some internal implementation detail of `kubectl` and `oc`. E.g.
`podman play kube` does not know about it.

Thus convert all our Lists to multi-document YAML format, with `---`
separators.

Adjust YAML parsing in tasks/run-local.sh accordingly.

[1] https://kubernetes.io/docs/reference/using-api/api-concepts/

----

I re-deployed all changed resources to make sure they still work fine.


martinpitt:
@jelly : Friendly review ping?

marusak (approved):
Interesting, I though that everyone was using List. Thanks!

PR #523, opened 2022-10-05
tasks: mc policy is deprecated

jelly:
minio renamed `mc policy` to `mc anonymous`.


jelly:
It was changed two days ago! https://github.com/minio/mc/releases/tag/RELEASE.2022-10-01T07-56-14Z

martinpitt (approved):
Whee, thank you! That's what we get for following `:latest`, but honestly I think it's still better that way. It didn't break too often so far (this was the first time)

PR #525, opened 2022-10-12
Add CJK fonts to task container

M4rtinK:
This way, we should be able to see CJK fonts instead of boxes in pixel test images, making it more likely to spot various issues involving non-latin languages.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -18,6 +18,7 @@ RUN dnf -y update && \
         git \
         git-lfs \
         gnupg \
+        google-noto-cjk-fonts-common \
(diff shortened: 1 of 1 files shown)


martinpitt:
@jelly : This needs some piloting. Re-propose from origin, refresh and roll out the container, and ensure that it doesn't regress (not only because of the new package, but due to general package updates -- usually new browser versions). The [last update](https://quay.io/repository/cockpit/tasks?tab=tags) happened a month ago.

jelly:
Re-opened here https://github.com/cockpit-project/cockpituous/pull/526

PR #527, opened 2022-10-13
.github: update checkout action to v3

jelly:
actions/checkout@v3 uses nodejs 16 as runtime instead of the deprecated nodejs 12 runtime.


martinpitt:
Can you please re-propose from origin? TBH it's best to just delete your fork, it's rarely useful (I haven't had one for a long time).

PR #519, opened 2022-09-02
Mostly revert "tasks: add anaconda project's build dependencies to container"

martinpitt:
Since https://github.com/rhinstaller/anaconda/pull/4287 , Anaconda RPMs
are built inside the test VM, not in the container. Just keep libtool
and gobject-introspection-devel, as anaconda needs them for running
`./configure`.

This mostly reverts commit e3afeb4f6a6308c8df1a653b50e116666b08f30b.

-----

This reduces the [compressed container size](https://quay.io/repository/cockpit/tasks?tab=tags) from 777 MB to 723 MB, and the uncompressed size from 2.3 GB to 2.1 GB.

 - [x] [Build](https://github.com/cockpit-project/cockpituous/actions/runs/3319836314) and deploy
 - [x] Test anaconda against new container: https://github.com/rhinstaller/anaconda/pull/4396
 - [x] Test cockpit PR against new container to check for regressions: https://github.com/cockpit-project/cockpit/pull/17839

PR #529, opened 2022-10-14
Drop Anaconda build-deps from task container

M4rtinK:
This reverts commit e3afeb4f6a6308c8df1a653b50e116666b08f30b.

Since https://github.com/rhinstaller/anaconda/pull/4287 has been merged, the Anaconda RPMs are now built in a VM. This has been working well, so I think we can drop the Anaconda build deps from the task container.

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -51,28 +51,6 @@ RUN dnf -y update && \
     curl -o /tmp/cockpit.spec -s https://raw.githubusercontent.com/cockpit-project/cockpit/main/tools/cockpit.spec | \
     dnf -y builddep /tmp/cockpit.spec && \
     rm /tmp/cockpit.spec && \
-    dnf -y install \
-        audit-libs-devel \
-        libtool \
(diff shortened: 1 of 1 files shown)


marusak:
We will need to build and deploy this. Also it is somewhat duplicate of https://github.com/cockpit-project/cockpituous/pull/519

I am happy to take a look or maybe @jelly is?

M4rtinK:
> Right. @M4rtinK, I tested this in [rhinstaller/anaconda#4308](https://github.com/rhinstaller/anaconda/pull/4308) and it [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4308-20220902-061014-0e47a4b4-fedora-rawhide-boot/log.html) at least due to some missing build dependencies. We need to find out which ones Anaconda needs, and keep these.
> 
> I propose to do that in #519, as the branch needs to be on origin for technical reasons. Thus closing this as a duplicate.

Sounds good to me, thanks! :)

PR #530, opened 2022-10-18
tasks: Hardcode github.com IP address

martinpitt:
DNS servers in e2e are rather flaky, and tests often fail on resolving github.com. Its IP does not seem to change that often, so hardcode it into tasks container's /etc/hosts.

---

As per discussion with @jelly . I rolled this out on e2e. If we go with this, we should limit it to e2e.


martinpitt:
This fell by the wayside when I rolled out the change from #537. Let's put this into the attic, and dust it off if/when we need it again.

PR #531, opened 2022-10-26
tasks: Move to Fedora 37, drop container script labels

martinpitt:
I built this locally and verified that I can build and test cockpit with chromium and firefox.

  - [x] [build](https://github.com/cockpit-project/cockpituous/actions/runs/3326429810) and deploy
  - [x] test with anaconda: https://github.com/rhinstaller/anaconda/pull/4399
  - [x] test with c-machines: https://github.com/cockpit-project/cockpit-machines/pull/838
  - [x] https://github.com/cockpit-project/cockpit/pull/17847

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -7,15 +7,12 @@ jobs:
     runs-on: ubuntu-22.04
     environment: quay.io
     timeout-minutes: 30
-    # needs to use docker, as podman-built images are not accepted to RHEL 8's docker any more
-    env:
-      DOCKER: docker
     steps:
       - name: Clone repository
         uses: actions/checkout@v3
 
       - name: Log into container registry
-        run: docker login -u ${{ secrets.QUAY_BOTUSER }} -p ${{ secrets.QUAY_TOKEN }} quay.io
+        run: podman login -u ${{ secrets.QUAY_BOTUSER }} -p ${{ secrets.QUAY_TOKEN }} quay.io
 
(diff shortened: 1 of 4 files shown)


martinpitt:
What the... why does the [rebuild](https://github.com/cockpit-project/cockpituous/actions/runs/3326351200/jobs/5500036953) try to install stuff from rawhide!? My local build is fine, it uses f37 as the `FROM` line tells it to. Same as in [tests/tasks](https://github.com/cockpit-project/cockpituous/actions/runs/3326314113/jobs/5499911376#step:5:18) here, that was also as expected.

.github/workflows/build-images.yml
```diff
@@ -7,15 +7,12 @@ jobs:
     runs-on: ubuntu-22.04
     environment: quay.io
     timeout-minutes: 30
-    # needs to use docker, as podman-built images are not accepted to RHEL 8's docker any more
```
martinpitt, line 10: FTR, we don't need them in RHEL 8 any more, our infra now runs on Fedora CoreOS.

martinpitt:
There's one regression, cockpit's `TestRunTest.testRetry` now [fails](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17839-20221026-054748-c5ecd1ca-arch/log.html#307-2). But this is subtle and easy enough to fix. I sent a PR, blocking on that.

marusak (approved):
Thank you!

PR #534, opened 2022-11-23
Drop sink

martinpitt:
We changed the last sink consumer in bots to put logs to S3 and report results directly to GitHub: https://github.com/cockpit-project/bots/pull/4075

Thus now we don't need all that complicated sink machinery any more. Drop the sink code, `$TEST_PUBLISH`, and installation of the `.ssh/config` secret -- with that, the `id_rsa{,.pub}` secrets can be dropped as well.

This also gets rid of one persistent EC2 instance (formerly logs.cockpit-project.org).

The only thing that is still left is exposing our metrics data (prometheus and test-results.db) on the CentOS CI sink pod, thus keep sink/centosci.yaml for the time being and just remove the sink logic part from it. Replacing that will take some more design discussion first.

Fix tasks/run-local.sh to not expect the "OK" from the sink unit tests
any more. Instead, verify that pyflakes was called.

----

"Let that sink out!"

broken sink

 - [x] drop EC2 sink instance, network, and volume
 - [x] drop CentOS CI  sink-ssh service
 - [x] fix [integration test](https://github.com/cockpit-project/cockpituous/actions/runs/3531847489/jobs/5925479995)
 - [x] [refresh images container](https://github.com/cockpit-project/cockpituous/actions/runs/3532205795), tear down metrics+sink on CentOS CI and re-deploy it, validate that it still works: [CI weather report](https://logs-https-frontdoor.apps.ocp.ci.centos.org/tests.html) and [grafana](https://grafana-frontdoor.apps.ocp.ci.centos.org/d/ci/cockpit-ci?orgId=1)
 - [x] drop obsolete ssh_config/id_* secrets, re-deploy
 - [x] [refresh tasks container](https://github.com/cockpit-project/cockpituous/actions/runs/3532204534) , deploy it
 - [x] ensure that refreshed tasks container still works for cockpit: https://github.com/cockpit-project/cockpit/pull/17947
 - [x] ensure that refreshed tasks container still works for image refreshes: https://github.com/cockpit-project/bots/issues/4093

PR #535, opened 2022-11-25
Drop cockpit/images container

martinpitt:
The remaining cleanup after #534. I already did `oc delete -f sink/sink-centosci.yaml` and I deployed the simple nginx. https://logs-https-frontdoor.apps.ocp.ci.centos.org/tests.html works as before.

Once this is approved:

 - [ ] Delete https://quay.io/repository/cockpit/images
 - [ ] Delete `tasks/htpasswd` secret
 - [ ] Roll out new secrets and restart tasks containers after approval, to pick up `server` → `s3-server` cert change name


martinpitt:
@marusak , @allisonkarlitskaya : Friendly review ping? I need to do some more infra changes which will require a container rebuild/restart to fix the never-ending npm cache corruptions. I don't want to have more than one intrusive PR in flight. Thanks!

allisonkarlitskaya (approved):
This looks pretty reasonable.

Happy to be rid of this increasingly-inaccurately named container :)

PR #537, opened 2022-12-01
tasks: Re-drop NPM cache

martinpitt:
Revert commit 8789af70e58aa. It has turned out that caching this over long times is not reliable, the cache keeps getting corrupted every few days.

cockpit-{podman,machines} now also use git-based node_modules/ tracking, so `npm install` runs are down to cockpit-{composer,certificates}, subscription-manager, and anaconda. The first three don't get a lot of PRs, and if anaconda starts running into npm install failures in PRs again, let's move it to tools/node_modules.

---

This stops [this kind of failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4116-20221129-230354-8be0c45c-rhel-9-1-candlepin-subscription-manager-subscription-manager-1.29.30/log.html), which we see in every other image refresh. It was an idea, it failed. :cry: 

I rolled this out to all tasks containers.


jelly (approved):

jelly:
Sounds good! I somewhere on my list had the plan to update cockpit-ostree to node-modules and as the different was also a bit annoying when updating the Makefile for starterkit changes.

PR #538, opened 2022-12-06
tasks: Drop pyflakes assumption in run-local.sh

martinpitt:
bots does not use pyflakes any more, and moved to flake8. It also does
not run its test with `set -x`, so the command doesn't actually appear
in the log. Just drop the assumption, it's not important and just wrong
now.

-----

test/cockpituous from https://github.com/cockpit-project/bots/pull/4129 repeatedly fails, this fixes it.

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -261,7 +261,6 @@ test_pr() {
     echo "--------------- test log end -------------"
     assert_in '<html>' "$LOG_HTML"
     assert_in 'Running on:.*cockpituous' "$LOG"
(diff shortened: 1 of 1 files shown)


martinpitt:
I [hacked](https://github.com/cockpit-project/bots/pull/4130/commits/60443ec7cdd979be33623dac334b8a629404421c) bots to run against this branch in https://github.com/cockpit-project/bots/pull/4130  and tests pass now.

marusak (approved):
Thank you!

PR #539, opened 2022-12-12
metrics: Move to new CentOS CI OCP

martinpitt:
The project name on the new OpenShift cluster is now "cockpit". The
PVC name changed, and the claim now only guarantees 5 GiB.

Add a paragraph about the CI weather to README.

---

I rolled this out to validate our [onboarding request](https://pagure.io/centos-infra/issue/991#comment-831950), and it works. I also copied over our `test-results.db`. Of course the [Grafana board](https://grafana-cockpit.apps.ocp.cloud.ci.centos.org/d/ci/cockpit-ci) is still rather empty, as the prometheus database was reset.

The weather report also works.

Currently, prometheus is not being updated, and the [statistics queue](http://webhook-cockpit.apps.ocp.cloud.ci.centos.org/inspect-queue) is growing. This is because the old CentOS CI cluster cannot talk to the webhook/AMQP queue on the new CentOS CI cluster, so I just shut down the tasks container there. I will soon replace this with some stripped down tasks container which *only* processes the statistics queue (see https://github.com/cockpit-project/bots/pull/4162). But let's land these other two pieces for now.

### DIFF
--- a/metrics/prometheus-claim.yaml
+++ b/metrics/prometheus-claim.yaml
@@ -2,11 +2,11 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: prometheus-db
-  namespace: frontdoor
+  name: prometheus-data
+  namespace: cockpit
 spec:
   accessModes:
   - ReadWriteOnce
   resources:
     requests:
-      storage: 10Gi
+      storage: 5Gi
--- a/metrics/README.md
+++ b/metrics/README.md
@@ -19,7 +19,7 @@ regularly read these metrics and store them in a database, and [Grafana](https:/
 
        metrics/deploy.sh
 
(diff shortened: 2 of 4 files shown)


metrics/prometheus-claim.yaml
```diff
@@ -2,11 +2,11 @@
...
-  namespace: frontdoor
+  name: prometheus-data
+  namespace: cockpit
 spec:
   accessModes:
   - ReadWriteOnce
   resources:
     requests:
-      storage: 10Gi
+      storage: 5Gi
```
KKoukiou, line 12: This change is delibarate ? (Not included in the commit msg or wondering if it's missed to explain, accident or just trivial)

martinpitt: It is deliberate, see the [onboarding request](https://pagure.io/centos-infra/issue/991#comment-831950). I adjusted the commit message accordingly, thanks for pointing out!

KKoukiou (approved):

PR #542, opened 2023-02-08
tasks: add python3-pip and esbuild to the tasks container

allisonkarlitskaya:
- [x] https://github.com/cockpit-project/cockpituous/actions/runs/4142420285
 - [x] Deploy to our CI
 - [ ] fix all the things:
   - [x] https://github.com/cockpit-project/cockpit-ostree/pull/329
   - [x] https://github.com/cockpit-project/cockpit-machines/pull/947
   - [x] https://github.com/cockpit-project/cockpit-podman/pull/1202
   - [ ] https://github.com/cockpit-project/cockpit/pull/18319
   - [x] validate [firefox run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18319-20230210-104814-82f0f4e9-fedora-37-firefox/log.html)

### DIFF
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -19,6 +19,7 @@ RUN dnf -y update && \
         git-lfs \
         gnupg \
         gobject-introspection-devel \
+        golang-github-evanw-esbuild \
         google-noto-cjk-fonts-common \
         intltool \
         jq \
--- a/tasks/Dockerfile
+++ b/tasks/Dockerfile
@@ -42,6 +42,7 @@ RUN dnf -y update && \
         python3-flake8 \
         python3-pika \
         python3-pillow \
+        python3-pip \
         python3-wheel \
         qemu-kvm-core \
         rpm-build \


martinpitt:
We haven't refreshed the tasks container in a while, so it's possible that there's some pixel fallout or worse. But +1 for the actual change of course.

allisonkarlitskaya:
ya... the run is still going and there are an awful lot of pixels changes already...

martinpitt:
@KKoukiou : FYI, I added another commit here to install the esbuild package. IMHO this is much better than pulling built binaries into node_modules/. esbuild is packaged in both Fedora and Debian (thus fine as a build dependency), and for RHEL we can just continue to ship the pre-built upstream dist/ files, and don't have to re-bundle during rpm build time.

martinpitt:
Deployed. Now we'll need to run a few tests against the new container to mop up pixel changes and possible other trouble.

jelly:
Only Cockpit left \o/

tasks/Dockerfile
```diff
@@ -19,6 +19,7 @@ RUN dnf -y update && \
         git-lfs \
         gnupg \
         gobject-introspection-devel \
+        golang-github-evanw-esbuild \
```
KKoukiou, line 22: So this won't help me much, as it's installed a very old version of the bundler which we can't use. We need esbuid >= 17.7 which is not available in fedora. The package there was not updated with esbuild version bump since almost a year.

martinpitt: And we need the latest crack? Well, this was purely opportunistic to play around with that, and compare "OS package" vs. "bundled". No commitment to anything right now.

martinpitt (approved):
The container works well enough that we want to go ahead with this. Thanks @jelly !

PR #555, opened 2023-07-24
tasks: Drop CentOS CI OpenShift deployment

martinpitt:
Current CentOS CI runs on AWS which doesn't support /dev/kvm. There is no k8s-on-bare-metal anywhere on the horizon, so drop this dead code. If it ever comes back, it will need adjustment anyway.

---

:sob: This was by far the best way to deploy and maintain our CI. We can't have nice things..


jelly (approved):
:cry:

PR #556, opened 2023-08-24
task: copy token if provided for interactive mode

jelly:


tasks/run-local.sh
```diff
@@ -86,7 +86,8 @@ EOF
         (mkdir -p tasks; cd tasks; $ROOTDIR/local-s3/generate-s3-cert.sh)
 
         # dummy token
-        echo 0123abc > "$SECRETS"/webhook/.config--github-token
+        [ -z "$TOKEN" ] || cp -fv "$TOKEN" "$SECRETS"/webhook/.config--github-token
+        [ -z "$TOKEN" ] && echo 0123abc > "$SECRETS"/webhook/.config--github-token
```
martinpitt, line 90: I am really allergic against this `&&` construct, as it's so weird in shell.; couldn't this just become a standard `if`/`then`/`else`?

jelly: Yes.

PR #561, opened 2023-10-27
local-s3: update-ca-trust is no longer part of the minio image

jelly:


jelly:
This feel hackish @martinpitt, but I think not only minion's cli utility use this cert but also curl right?

So bind mounting the ca cert into $HOME/.minio/certs/ca likely won't help?
https://min.io/docs/minio/linux/operations/network-encryption.html#self-signed-internal-and-private-certificates

martinpitt (approved):
Thanks! This also breaks the production instance, so let's land this. Yes, it's a bit hackish, but in essence that cating is all that update-ca-trust does.

martinpitt:
Deployed, seems to work :tada:

PR #562, opened 2023-11-10
metrics: Add podman deployment

martinpitt:
This makes working on metrics easy for development and experimentation.

---

After that I'll update our boards a bit wrt. changed queue time SLO


metrics/deploy-podman.sh
```diff
@@ -0,0 +1,45 @@
...
+podman kube down "${MYDIR}/metrics.yaml"
+for filter in grafana prometheus-config; do
+    podman volume ls -q --filter "name=$filter" | xargs --no-run-if-empty podman volume rm
+done
+
+# adjust the k8s deployment to work for podman play kube:
+#  - grafana-dashboards ConfigMap gets built dynamically in metrics/deploy-k8s.sh, but
+#    it is much more flexible and ergonomic to just mount the dashboards directory
+#  - replace grafana-admin secret with static "foobar" admin password
+patch -o- "$MYDIR/metrics.yaml" - <<EOF | podman kube play -
```
jelly, line 17: I assume we cannot do an include?

jelly: So seeing this works with secrets and `podman secret` being a thing, maybe we can use that instead?

That however does not solve the `hostPath` issue.

jelly: https://www.redhat.com/sysadmin/podman-kubernetes-secrets

martinpitt: Nice! I have an even higher opinion of `podman play kube` now, it has come a long a great way in the last few years.

metrics/deploy-podman.sh
```diff
@@ -0,0 +1,41 @@
...
++++ metrics/metrics.yaml
+@@ -75,9 +73,8 @@ spec:
+           configMap:
+             name: grafana-provisioning-dashboards
+         - name: grafana-dashboards
+-          configMap:
+-            # this is not defined here, but gets built from *.json files in ./deploy.sh
+-            name: grafana-dashboards
++          hostPath:
++            path: metrics/dashboards
```
jelly, line 37: I wonder if we can use `--configMap` but I am not 100% sure how that works so I'll leave it up to you.

https://docs.podman.io/en/v3.4.1/markdown/podman-play-kube.1.html#configmap-path

jelly (approved):

PR #564, opened 2023-11-13
tasks: Check image pruning in run-local.sh

martinpitt:
This validates https://github.com/cockpit-project/bots/pull/5542

----

 - [x] Remove the [extra commit](https://github.com/cockpit-project/cockpituous/commit/0721b4494c399001a0c920b897a6bb4338ebc3da) that tests the bots branch instead of main

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -226,7 +226,17 @@ test_image() {
         cd bots
         ./image-download --store '$S3_URL_POD'/images/ testimage
         grep -q "^world" /cache/images/testimage-*.qcow2
+        rm --verbose /cache/images/testimage*
         '
+
+    # validate image pruning on s3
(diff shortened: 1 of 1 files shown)


martinpitt:
> Error: unknown shorthand flag: 't' in -t0

Meh, old podman. I'll drop this commit.

martinpitt:
With the extra "use branch" commit, the [test passes](https://github.com/cockpit-project/cockpituous/actions/runs/6852889693/job/18632467433?pr=564#step:6:239). Will drop that commit once https://github.com/cockpit-project/bots/pull/5542 lands.

martinpitt:
Actually, I already dropped it now, to prove that the test fails and reproduces the bug. We can just retry it after landing the bots PR.

jelly (approved):

PR #566, opened 2023-11-16
Add fedora-logos in the Dockerfile

KKoukiou:
Needed by anaconda-webui RPM.

### DIFF
--- a/tasks/Containerfile
+++ b/tasks/Containerfile
@@ -11,6 +11,7 @@ RUN dnf -y update && \
         dbus-glib \
         diffstat \
         expect \
+        fedora-logos \
         firefox \
         fpaste \
         gcc-c++ \


martinpitt (approved):
Thanks! I'll trigger a build, but these refreshes tend to break pixel tests and such, so we'll need to work through all our projects tomorrow. But that's ok.

martinpitt:
Triggered a build: https://github.com/cockpit-project/cockpituous/actions/runs/6893507798 . This will trickle into the bots by tomorrow, but I can also make sure to re-deploy it synchronously.

martinpitt:
@KKoukiou While it's harmless, I still wonder why -- shouldn't this be a *runtime* dependency, not a build-time one? And even if it's a build-time one, build should happen in a VM, not on the host. You'd *never* rpm-install anaconda-webui.rpm inside the tasks container.

I refreshed my local cockpit/tasks toolbox with this rebuild, and locally ran the podman tests on f39.

martinpitt:
I rolled this out as it doesn't seem to cause too many regressions (and I hope these are naughties), and it's good to refresh the container from time to time anyway. So you should be unblocked, but please fix the build/binary dependencies of this. Not landing yet, until we discussed that.

KKoukiou:
You 're right, build is happening in the VM, this is not needed.

PR #568, opened 2023-12-05
Completely remove Anaconda build dependencies

KKoukiou:
We don't build anaconda-core from sources anymore. We do that for the anaconda-webui only, but this is not needed for it.

Followup for 593d3d18629adc0b0cea17e332a8c5bec6394652.


martinpitt (approved):
Thanks! CI doesn't work from a branch, but I think we can just commit that without actually refreshing the tasks container.

PR #570, opened 2024-01-09
PSI/e2e infra updates

martinpitt:
This got rolled out already.

### DIFF
--- a/ansible/psi/deploy-all.sh
+++ b/ansible/psi/deploy-all.sh
@@ -5,9 +5,12 @@ set -eu
 # cd to ansible/
 cd "$(dirname $(dirname $(realpath $0)))"
 
+# first instance name; can be set by env
+FIRST=${FIRST:-1}
+
 # number of instances; limited by quota
 NUM=35
 
-seq $NUM | parallel --line-buffer -j4 ansible-playbook -i inventory/ -e instance_name='rhos-01-{}' psi/launch-tasks.yml
+seq $FIRST $NUM | parallel --line-buffer -j4 ansible-playbook -i inventory/ -e instance_name='rhos-01-{}' psi/launch-tasks.yml
 
 ansible-playbook -i inventory/ psi/image-cache.yml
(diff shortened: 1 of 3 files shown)


ansible/psi/deploy-all.sh
```diff
@@ -5,9 +5,12 @@ set -eu
...
+FIRST=${FIRST:-1}
+
 # number of instances; limited by quota
 NUM=35
 
-seq $NUM | parallel --line-buffer -j4 ansible-playbook -i inventory/ -e instance_name='rhos-01-{}' psi/launch-tasks.yml
+parallel -i -j4 ansible-playbook -i inventory/ -e instance_name='rhos-01-{}' psi/launch-tasks.yml -- $(seq $FIRST $NUM)
 
-ansible-playbook -i inventory/ psi/image-cache.yml
+ansible-playbook -f20 -i inventory/ psi/image-cache.yml
```
jelly, line 16: non-blocking nit, `--forks` would be what I prefer reading here. 20 sounds like a lot then again never seen ansible-playbook runs "eat resources" with the default 5.

jelly (dismissed):

ansible/psi/deploy-all.sh
martinpitt: On the client side it's really just a bunch of `ssh` processes. I find 20 to be just fine for all our cockpituous cases, we don't do anything expensive on the caller side.

But I'll change to `--forks` indeed.

mvollmer (commented):
I have no idea what's going on here. :-)

martinpitt:
@mvollmer commits 1, 2, 4 should be obvious; 3 is the broken networking that I reported to the admins, and so far changing the network to `shared_net_2` has worked. It's more or less just "try them and see what works". I'm happy to talk you through commit 5 "Configure local image S3 store in e2e" if you are interested.

jelly (approved):
Thanks

PR #571, opened 2024-01-22
add tools to be able improve ISO for rhel4edge

jscotka:


tasks/Containerfile
```diff
@@ -27,14 +29,17 @@ RUN dnf -y update && \
         libvirt-client \
         libvirt-python3 \
         libXt \
+        mtools \
         nc \
         net-tools \
         nodejs-devel \
         npm \
+        ostree \
```
martinpitt, line 37: What do you need this for? I'm rather sceptical that this will work in an unprivileged container. You can run `toolbox create --image quay.io/cockpit/tasks -c cockpit` to test it.

tasks/Containerfile
```diff
@@ -48,10 +53,12 @@ RUN dnf -y update && \
         rpmdevtools \
         rsync \
         socat \
+        syslinux \
         strace \
         tar \
         vim-enhanced \
-        virt-install && \
+        virt-install \
+        xorriso && \
```
martinpitt, line 61: The `&&` needs to move a line up.

tasks/Containerfile
```diff
@@ -48,10 +53,12 @@ RUN dnf -y update && \
         rpmdevtools \
         rsync \
         socat \
+        syslinux \
```
martinpitt, line 56: This is certainly a curious thing to have in the tasks container -- certainly the rhel4edge builds from image builder already have a boot loader? We want to test these images, not apply a dozen hacks to fix their boot. What's that for?

martinpitt (changes requested):
Thanks! Some questions though. Deploying a new tasks container is a two-day effort, so I'd like to sort this out first.

tasks/Containerfile
jscotka: it is based on dependencies inside podman image what was used before: https://github.com/loadtheaccumulator/fleet-iso-util/

martinpitt: But https://github.com/loadtheaccumulator/fleet-iso-util/blob/main/fleetkick.sh doesn't use ostree anywhere?

PR #574, opened 2024-02-15
ansible: Document how to run a playbook against a local VM

martinpitt:
Add an example inventory for our standard cockpit bots `c` machine. It should be obvious from there how to extend it.


allisonkarlitskaya (approved):
Thanks!

PR #576, opened 2024-02-19
tasks: add python3-aiohttp and aioresponses

allisonkarlitskaya:
This seems to be the standard Python asyncio HTTP library. python3-aioresponses is a useful mocking framework for use with pytest.

We're going to be using these soon.


martinpitt (approved):
`await ci.run(); await merge()`

allisonkarlitskaya:
I triggered `image:fedora-39` on bots/main as a way of validating this and it came back almost entirely green (aside known flakes).

PR #578, opened 2024-02-27
tasks: Document running single container locally in podman

martinpitt:
Rendered view: https://github.com/cockpit-project/cockpituous/blob/local-doc/tasks/README.md#running-single-container-locally


mvollmer (approved):
Nice, this works!

PR #581, opened 2024-02-29
ansible: Clean up old container images in deploy-tasks-container.yml

martinpitt:
Our machines have piled up a few dozen GB of old tasks containers. This will keep the current and the new image, and get rid of all older ones.

----

I validated that this works while deploying the new tasks container for https://github.com/cockpit-project/cockpit/pull/20107#issuecomment-1970732087

### DIFF
--- a/ansible/maintenance/deploy-tasks-container.yml
+++ b/ansible/maintenance/deploy-tasks-container.yml
@@ -12,6 +12,10 @@
   gather_facts: false
 
   tasks:
+  - name: Clean up old container images
+    # this will fail due to at least one used image, but it's opportunistic anyway
+    shell: podman rmi --all || true
+
   - name: Pre-pull current container image to avoid long downtime
     command: podman pull quay.io/cockpit/tasks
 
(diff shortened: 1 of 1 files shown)


ansible/maintenance/deploy-tasks-container.yml
```diff
@@ -18,6 +18,10 @@
   - name: Tell tasks containers to drain and restart
     command: pkill -ex cockpit-tasks
 
+  - name: Clean up old container images
+    # this will fail due to at least one used image, but it's opportunistic anyway
+    shell: podman rmi --all || true
```
mvollmer, line 23: So, even if this would remove all images, that would still be correct, right? We don't rely on the current image being protected by being "in use". Yep, just talking to myself...

mvollmer: No, wait.  We _do_ pull the new image before this. Shouldn't this be the other way around?

mvollmer (changes requested):

ansible/maintenance/deploy-tasks-container.yml
martinpitt: Doesn't matter much -- at that point the old containers are still running, so they will keep the *current* image busy. The point is to remove the older ones. Keeping the current and new images around is fine, and even useful in case we want to quickly revert.

mvollmer (changes requested):
Let's prune first and pull after.

mvollmer (approved):

PR #583, opened 2024-03-04
tasks: Bind the host's podman API socket

martinpitt:
tasks: Bind the host's podman API socket

This paves the way for spawning per-job tasks containers from the
container (via `job-runner`).

Getting the permissions right is unfortunately annoyingly complicated, as the
host's socket has 660 permissions, but the tasks container runs as uid 1111.
Ideally we could use something like

    -v "${XDG_RUNTIME_DIR:-/run}/podman/podman.sock:/podman.sock:idmap=gids=$(id -g)-1111-1"

but that fails with "invalid mappings", and is generally not well documented.
`--mount=type=bind,[...],idmap --uidmap [...]` does not work either.

So resort to file permissions for the host's `podman.sock`:
 - For local developers in `run-local.sh`, make the socket world
   writable. This does not actually hurt for a human developer: Its
   directory (/run/user/uid) is not accessible for any other user.

 - For production, set the socket group to `1111`. That doesn't matter
   much, the secrets are all already chmod'ed to the container user, and
   these machines don't do anything else. `setfacl` would be more
   targeted, but it isn't installed in Fedora IoT/CoreOS.

-----

I rolled this out to one production machine:

    ansible -i inventory -m include_role -a name=tasks-systemd -e instances=1 -l rhos-01-1 openstack_tasks

Et voilà:
```
❱❱❱ ssh rhos-01-1 sudo podman exec -it cockpit-tasks-1 podman-remote ps
Warning: Permanently added '10.0.203.194' (ED25519) to the list of known hosts.
CONTAINER ID  IMAGE                         COMMAND               CREATED        STATUS        PORTS                 NAMES
51ef06c0f09c  quay.io/minio/minio:latest    server /data          4 days ago     Up 4 days     0.0.0.0:80->9000/tcp  cockpit-s3
ea0bd1fbcc99  quay.io/cockpit/tasks:latest  /usr/local/bin/co...  5 minutes ago  Up 5 minutes                        cockpit-tasks-1
```

Once this lands, I'll apply it to all our production machines.

### DIFF
--- a/tasks/install-service
+++ b/tasks/install-service
@@ -36,6 +36,8 @@ touch "$IMAGE_STORES"
 cat <<EOF > /etc/systemd/system/cockpit-tasks@.service
 [Unit]
 Description=Cockpit Tasks %i
+Requires=podman.socket
+After=podman.socket
 
 [Service]
 Environment="TEST_JOBS=${TEST_JOBS:-8}"
@@ -53,6 +55,8 @@ ExecStartPre=-/usr/bin/podman network rm cockpit-tasks-%i
 ExecStartPre=/usr/bin/chcon -R -l s0 \${TEST_CACHE}/images/
 ExecStartPre=/usr/bin/flock /tmp/cockpit-image-pull podman pull quay.io/cockpit/tasks
 ExecStartPre=/usr/bin/podman network create cockpit-tasks-%i
+# idmapped mount would be better, but did not figure out how
+ExecStartPre=/usr/bin/chgrp 1111 %t/podman/podman.sock
 ExecStart=/usr/bin/podman run --name=cockpit-tasks-%i --hostname=${CONTAINER_HOSTNAME} \
     --device=/dev/kvm --network=cockpit-tasks-%i \
     --memory=24g --pids-limit=16384 --shm-size=1024m ${TMPVOL:-} \
@@ -61,8 +65,10 @@ ExecStart=/usr/bin/podman run --name=cockpit-tasks-%i --hostname=${CONTAINER_HOS
     --volume=\${TEST_SECRETS}/webhook:/run/secrets/webhook:ro \
     --volume=${IMAGE_STORES}:/work/.config/cockpit-dev/image-stores:ro \
     --volume=/etc/npmrc:/etc/npmrc:ro \
+    --volume=%t/podman/podman.sock:/podman.sock:rw \
     --env=COCKPIT_GITHUB_TOKEN_FILE=/run/secrets/webhook/.config--github-token \
     --env=COCKPIT_S3_KEY_DIR=/run/secrets/tasks/s3-keys \
+    --env=CONTAINER_HOST=unix:///podman.sock \
     --env=TEST_JOBS=\${TEST_JOBS} \
     --env=TEST_NOTIFICATION_MX=\${TEST_NOTIFICATION_MX} \
     --env=TEST_NOTIFICATION_TO=\${TEST_NOTIFICATION_TO} \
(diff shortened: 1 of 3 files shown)


martinpitt:
[eh what](https://github.com/cockpit-project/cockpituous/actions/runs/8136575508/job/22233168910?pr=583)? `tasks/run-local.sh: 305: podman-remote: not found`. We added that two weeks ago in commit 89b9700c6da, and in my local tasks container it's very much available. The workflow also didn't rebuild the container. `podman run -it --rm quay.io/cockpit/tasks podman-remote --help` works fine. :exploding_head:

martinpitt:
Ugh, [running the test on GitHub](https://github.com/cockpit-project/cockpituous/actions/runs/8136854481/job/22233961091?pr=583) is difficult. First of all, the `setfacl` did succeed, but the container still gets an EPERM. If I hack this with `chmod 666`, then it fails with
```
Error: unable to connect to Podman socket: server API version is too old. Client "4.0.0" server "3.4.4"
```

https://github.com/containers/podman/discussions/17362 doesn't make me very hopeful, but there is somebody who backported the Debian bookworm packages to Ubuntu 22.04: https://launchpad.net/~quarckster/+archive/ubuntu/containers . This works very well indeed!

martinpitt:
Rebased after #585

tasks/install-service
```diff
@@ -61,8 +65,11 @@ ExecStart=/usr/bin/podman run --name=cockpit-tasks-%i --hostname=${CONTAINER_HOS
     --volume=\${TEST_SECRETS}/webhook:/run/secrets/webhook:ro \
     --volume=${IMAGE_STORES}:/work/.config/cockpit-dev/image-stores:ro \
     --volume=/etc/npmrc:/etc/npmrc:ro \
+    --volume=%t/podman/podman.sock:/podman.sock:rw \
```
allisonkarlitskaya, line 68: Why not use the "normal" location on the inside, and save ourselves the envvar?

tasks/install-service
```diff
@@ -61,8 +65,11 @@ ExecStart=/usr/bin/podman run --name=cockpit-tasks-%i --hostname=${CONTAINER_HOS
...
     --env=COCKPIT_GITHUB_TOKEN_FILE=/run/secrets/webhook/.config--github-token \
     --env=COCKPIT_S3_KEY_DIR=/run/secrets/tasks/s3-keys \
+    --env=CONTAINER_HOST=unix:///podman.sock \
+    --env=NPM_REGISTRY=\${NPM_REGISTRY} \
```
allisonkarlitskaya, line 72: That looks like a rebase error...

allisonkarlitskaya (changes requested):

tasks/install-service
martinpitt: Heh, I spent about two hours on that. There's no `$XDG_RUNTIME_DIR` in the tasks container, so podman defaults to some `/tmp/podman-1111/whatever` path which doesn't feel very predictable. Plus, if you actually bind-mount it there, it then wants a /work/rundir/podman/... thing. This is by far the easiest and most robust version that I found.

tasks/install-service
martinpitt: Oh yes it is, will fix.

martinpitt: Done.

martinpitt:
@allisonkarlitskaya Simplified as discussed, fixed the rebase error, and I redeployed it on rhos-01-1.  I re-validated that I can access podman there.

tasks/install-service
allisonkarlitskaya: I mean `/run/podman/podman.sock`.  I guess that's only the default when you're root, though?

martinpitt: But we aren't, we are uid 1111 :grin:

allisonkarlitskaya (approved):
Better, thanks!

PR #584, opened 2024-03-04
tasks: Add local mock PR run to run-local.sh

martinpitt:
This is an "almost end to end" test which runs fully locally and requires no
GitHub interaction or token. Add a little mock GitHub server (utilizing bots'
`task.test_mock_server`) to respond to the required GitHub APIs for handling an
"opened PR" webhook event.

This is useful for playing around with our AMQP queue/job execution engine
locally (when using the mock GH server in interactive mode), or
validating changes to run-queue/AMQP structure.

 - [ ] Requires #585

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ tasks-shell:
 	$(DOCKER) run -ti --rm \
 		--shm-size=1024m \
 		--volume=$(CURDIR)/tasks:/usr/local/bin \
-		--volume=$(TASK_SECRETS):/secrets:ro \
+		--volume=$(TASK_SECRETS):/run/secrets/tasks/:ro \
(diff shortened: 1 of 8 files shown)


martinpitt:
This is getting close -- the [inner test fails](https://github.com/cockpit-project/cockpituous/actions/runs/8141672656/job/22249531068?pr=584) because that `run-queue` is running in parallel with the `cockpit-tasks` main loop. We want to get rid of that anyway, though. I tried `kill -STOP 1`, but that unfortunately doesn't work (we can STOP the `sleep 10`, but ugh).

martinpitt:
OK, #585 is working, let's merge this new test there. No reason to drag around two PRs.

PR #585, opened 2024-03-04
tasks: Drop cockpit-tasks as main entry point, move tasks secrets from /secrets to /run/secrets/tasks, drop setup-tasks, add local mock PR test

martinpitt:
This is various general cleanup and making the tasks container a bit more flexible. I like the new "fully local almost end-to-end PR test" which makes hacking on our infra easy to validate (the new job runner infra).


tasks/install-service
```diff
@@ -62,7 +62,7 @@ ExecStart=/usr/bin/podman run --name=cockpit-tasks-%i --hostname=${CONTAINER_HOS
     --device=/dev/kvm --network=cockpit-tasks-%i \
     --memory=24g --pids-limit=16384 --shm-size=1024m ${TMPVOL:-} \
     --volume=\${TEST_CACHE}/images:/cache/images:rw \
-    --volume=\${TEST_SECRETS}/tasks:/secrets:ro \
+    --volume=\${TEST_SECRETS}/tasks:/run/secrets/tasks:ro \
```
allisonkarlitskaya, line 65: I wonder if we could think a step ahead here and:

 - add `/run/secrets/github-token` as the 'authoritative' location for the token file
 - add `/run/secrets/s3-keys` as the location for the s3-keys directory

...and adjust the symlinks to point to those things.  That's where we'll want to have those two things inside of containers spawned by `job-runner`, under control of the "secrets" attribute in the job definition.

Of course, we'd also continue to have the volume mount for `secrets/webhook` and `secrets/tasks` since we still use those at present.  We might want to start thinking about what to do more generally with "external config" such as the npm-registry certificate, though... it's not really a secret.

tasks/run-local.sh
```diff
@@ -176,7 +176,9 @@ EOF
         -e AMQP_SERVER=$AMQP_POD \
         -e S3_LOGS_URL=$S3_URL_POD/logs/ \
         -e SKIP_STATIC_CHECK=1 \
-        quay.io/cockpit/tasks:${TASKS_TAG:-latest} ${INTERACTIVE:+sleep infinity}
+        quay.io/cockpit/tasks:${TASKS_TAG:-latest} bash
```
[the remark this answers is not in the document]
martinpitt: I started with `sleep infinity`, but `bash` as pid 1 has the advantage that it reaps zombies. `cockpit-tasks` relies on that in its `pkill virtqemud; while pgrep` loop, and generally it's a bit irritating to pile up zombies in the container.

tasks/run-local.sh
```diff
@@ -176,7 +176,9 @@ EOF
...
+
+    podman exec -i cockpituous-tasks setup-tasks
```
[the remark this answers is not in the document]
martinpitt: Maybe -- that's what I'm not yet clear about. E.g. do we want to run it in the per-job instance containers? Do they need the bots checkout? (I figure that part is the step child). I think we need to talk about that.

In general I'm sympathetic about making this an entry point, I thought about it myself, too.

Makefile
```diff
@@ -22,7 +22,7 @@ tasks-shell:
 	$(DOCKER) run -ti --rm \
 		--shm-size=1024m \
 		--volume=$(CURDIR)/tasks:/usr/local/bin \
-		--volume=$(TASK_SECRETS):/secrets:ro \
+		--volume=$(TASK_SECRETS):/run/secrets/tasks/:ro \
```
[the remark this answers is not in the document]
martinpitt: Yeah, I know :/

tasks/install-service
martinpitt: No objection to the two extra volumes (and doing the same in `run-local.sh`). Do you want to do that in this PR?

I'm not sure if we can commit the npm registry internal RH mirror cert to a public place -- it's not really secret, but it applies to RH internal infrastructure only. TBH I wouldn't change that right now without at least some more checking.

tasks/run-local.sh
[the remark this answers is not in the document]
martinpitt: Doing `-ec "setup-tasks; bash" WFM, but see the entry-point discussion below.

tasks/run-local.sh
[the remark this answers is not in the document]
martinpitt: My gut feeling: Make `setup-tasks` an entry point, but revert "Move initial bots clone to setup-tasks" and live with the duplication in run-local.sh. Then this only does secrets (symlink/OpenShift unpacking) setup, which we really want everywhere.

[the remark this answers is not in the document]
allisonkarlitskaya: My gut feeling: I hate `setup-tasks`.  I think we have three good tools at our disposal and we need to use them:

 - the container image itself
 - the ability to arbitrarily bind-mount stuff in place
 - the fact that we write almost all of the software that ultimately is running inside of our container and interacting with these config bits

We should be able to construct our entire "world" out of the first two, and to the extent that its not possible, deal with it by way of the last one.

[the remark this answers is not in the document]
allisonkarlitskaya: Add to the above list, of course: 
 - environment variables

tasks/install-service
allisonkarlitskaya: I'd like to see the github-token and s3-keys bit in this PR.  No time like the present!

The npm registry cert is a bit more interesting and it dovetails into the question about what to do with the npm config bits that current live in `setup-tasks` (which needs to die, I'm increasingly convinced)...

tasks/run-local.sh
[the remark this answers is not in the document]
martinpitt: - We can replace the `npm config`s with a ~/.npmrc which we bind-mount as a volume, and keep the custom registry in cockpitous.git for our bots; that part is relatively simple
- We can directly mount ~/.config/cockpit-dev/s3-keys and  ~/.config/github-token on `podman` based bots.
- `~/.git-credentials` could be replaced with a volume if we duplicate the secret and its setup as a bind mount.

That leaves the OpenShift user and credentials setup, which *really* needs to be dynamic. We can split it out into `setup-openshift` and start the container through it as custom entry point (or command) in the yaml.

As I said, we can move bots back, we presumably don't need it in the job runners, just the monitor.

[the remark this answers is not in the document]
allisonkarlitskaya: I think `.git-credentials` isn't actually used, for the record.  My local testing showed that bots (ie: `tasks/__init__.py`) is smart enough to setup the credentials for itself based on the API key from the usual file.

I have no idea what the OpenShift stuff is for so I'm happy to defer to you on that point, but ya, my instinct there was that this is the sort of thing that can be handled in a script elsewhere, if it's required.

And ya, for npm, I just played with that and got:

```
fetch-retries=6
fetch-timeout=600000
fetch-retry-mintimeout=60000
maxsockets=3
registry=https://xyz
```

in my `~/.npmrc`, but also notice that `/etc/npmrc` is a thing, and this is probably a better fit.  I bet we can get the certificate specified via this file, too...

Which brings me to my next point: I wonder if we should have something like /run/external-config/ with files like `npmrc` in it, and symlink those from /etc in the container image.  This would be our primary mechanism for injecting external configuration for everything but secrets (in the `job-runner` sense of the term).  We don't need to do this now, but in the long-run it's probably going to be somewhat cleaner than a dozen bindmounts...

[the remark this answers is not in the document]
martinpitt: >  I bet we can get the certificate specified via this file, too...

That unfortunately needs the silly `$NODE_EXTRA_CERTS` env variable, but that's already handled differently. I.e. we can put that cert file wherever we like (including `/run/external-config/`), just need to point to it.

I'm not sure about .git-credentials. It may be a leftover from the time when we did `npm-update`, `po-refresh` etc. bots in our CI, as we do the [exact same thing in the workflows](https://github.com/cockpit-project/cockpit/blob/main/.github/workflows/weblate-sync-po.yml#L45). So we can try to drop it, and put it back if necessary :grin:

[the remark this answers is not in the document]
martinpitt: Tested on a bot:
```
git clone https://github.com/cockpit-project/starter-kit
cd starter-kit
rpm -rf ~/.npm
unset NODE_EXTRA_CERTS
npm install  # → hangs

npm config set cafile /secrets/npm-registry.crt
npm install # → works
```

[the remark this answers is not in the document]
allisonkarlitskaya: You can also set `ca` directly as the content of the certificate and avoid having to bring it in as a secret..

tasks/run-local.sh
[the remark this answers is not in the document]
martinpitt: setup-tasks is gone, and I added a comment why we want `bash`.

tasks/run-local.sh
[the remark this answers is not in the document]
martinpitt: setup-tasks is gone, and I implemented the things discussed above.

tasks/install-service
martinpitt: I didn't yet do this, instead used the env variables. Let's stabilize this version first, then I can take another look.

martinpitt:
@allisonkarlitskaya Reworked, and I also tested the webhook in an interactive deployment with `tasks/run-local.sh -i -t ~/.config/cockpit-dev/github-token`. Then `podman logs cockpituous-webhook` shows that it properly scans our projects, and `bots/inspect-queue --amqp localhost:5671` in the tasks container shows the scanned results.

At this point I'd like at least a cursory re-review. When you are by and large happy, I can try to deploy this to prod to see what I missed :wink:

tasks/install-service
allisonkarlitskaya: Sure.  With the envvars, I care a lot less.

tasks/webhook
```diff
@@ -74,9 +74,9 @@ def distributed_queue(amqp_server):
         logging.error('Please format amqp_server as host:port')
         sys.exit(1)
 
-    context = ssl.create_default_context(cafile='/run/secrets/webhook/ca.pem')
-    context.load_cert_chain(keyfile='/run/secrets/webhook/amqp-client.key',
-                            certfile='/run/secrets/webhook/amqp-client.pem')
+    context = ssl.create_default_context(cafile=os.path.join(WEBHOOK_SECRETS, 'ca.pem'))
```
allisonkarlitskaya, line 62: lis's list of favourite path mangling technologies (in order):
 - pathlib.Path
 - f-strings
 - ...
 - os.path.join()

Not worth getting into a fight over, though ;)

tasks/install-service
```diff
@@ -65,8 +60,7 @@ ExecStart=/usr/bin/podman run --name=cockpit-tasks-%i --hostname=${CONTAINER_HOS
     --volume=\${TEST_SECRETS}/tasks:/run/secrets/tasks:ro \
     --volume=\${TEST_SECRETS}/webhook:/run/secrets/webhook:ro \
     --volume=${IMAGE_STORES}:/work/.config/cockpit-dev/image-stores:ro \
-    --env=NPM_REGISTRY=\${NPM_REGISTRY} \
-    --env=NODE_EXTRA_CA_CERTS=${NODE_EXTRA_CA_CERTS:-} \
+    --volume=/etc/npmrc:/etc/npmrc:ro \
```
allisonkarlitskaya, line 63: This is *so* much better.  Thank you.

tasks/setup-tasks
```diff
@@ -2,19 +2,10 @@
...
-if ! whoami && [ -w /etc/passwd ]; then
-    echo "user:x:$(id -u):0:random uid:/work:/sbin/nologin" >> /etc/passwd
-    export HOME=/work
-fi
-
 # Set up secrets
 if [ -d /run/secrets/tasks ]; then
     ln -snf /run/secrets/tasks/s3-keys ~/.config/cockpit-dev/s3-keys
     ln -snf /run/secrets/webhook/.config--github-token ~/.config/github-token
-    git config --global credential.helper store
```
allisonkarlitskaya: The commit message here is incorrect.  `image-refresh` still pushes branches, for example, but it's covered by this:

https://github.com/cockpit-project/bots/blob/6b7864a1e6d76bfd29b700cdf09a807cb5e3a55e/task/__init__.py#L287

tasks/build-secrets
```diff
@@ -16,7 +16,9 @@ metadata:
 data:
 EOF
 cd "$BASE/tasks"
-for f in $(find -type f -o -type l); do
+# This intentionally does not capture subdirs like tasks/s3-keys/. As OpenShift secret volumes don't have subdirectories,
+# these need to be created as a separate volume if and when we ever need that.
+for f in $(find -maxdepth 1 -type f -o -type l); do
     printf '  %s: %s\n' "$(echo $f | sed 's!^./!!; s!/!--!g')" "$(base64 --wrap=0 $f)"
```
allisonkarlitskaya, line 22: I guess you can remove half of the path mangling seddery here now?

You could also rewrite your find command like so:

  `find * -type f -o -type l -maxdepth 0`

to get rid of the other half of the sed.  That would be a substantial simplification.

Probably also works:

```sh
for f in *; do
  if [ -f $f ]; then
    printf '  %s: %s\n' "$f" "$(base64 --wrap 0 "$f")"
  fi
done
```

(`test(1)` dereferences links for file-type operations)

tasks/cockpit-tasks
```diff
@@ -2,8 +2,6 @@
 
 set -eux
 
-setup-tasks
```
allisonkarlitskaya, line 5: yay!!

tasks/Containerfile
```diff
@@ -64,7 +64,7 @@ RUN dnf -y update && \
     dnf clean all && \
     pip install ruff
 
-COPY setup-tasks cockpit-tasks install-service webhook github_handler.py /usr/local/bin/
```
allisonkarlitskaya, line 67: yay!!

tasks/github_handler.py
```diff
@@ -16,7 +16,7 @@ def check_sig(self, request):
         See https://developer.github.com/webhooks/securing/
         '''
         # load key
-        keyfile = os.path.expanduser('~/.config/github-webhook-token')
+        keyfile = os.getenv("COCKPIT_GITHUB_WEBHOOK_TOKEN_FILE", os.path.expanduser('~/.config/github-webhook-token'))
```
allisonkarlitskaya, line 19: grumble, grumble

tasks/run-local.sh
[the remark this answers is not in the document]
allisonkarlitskaya: Educational!  Thanks!

allisonkarlitskaya (approved):
This looks amazing.  The usual minor gripes, but to be honest, if you want to land it as-is, go right ahead.

tasks/webhook
martinpitt: Path() is too much overhead for my taste, so I picked your second favorite :grin:

tasks/setup-tasks
martinpitt: Thanks, adjusted the commit message. Also, this is still subject to real-life testing.

tasks/build-secrets
martinpitt: Thanks for pointing out -- I wanted to do that, then it dropped from my  attention. `for f in *` is awkward as it doesn't consider hidden files; same for `find *`. I wanted to move it into a helper, but as we treat the metrics secret special (trim line endings), I don't bother -- this is rather unimportant anyway.

I diffed the before/after output, and it's identical.

tasks/build-secrets
```diff
@@ -16,8 +16,10 @@ metadata:
 data:
 EOF
 cd "$BASE/tasks"
-for f in $(find -type f -o -type l); do
-    printf '  %s: %s\n' "$(echo $f | sed 's!^./!!; s!/!--!g')" "$(base64 --wrap=0 $f)"
+# This intentionally does not capture subdirs like tasks/s3-keys/. As OpenShift secret volumes don't have subdirectories,
+# these need to be created as a separate volume if and when we ever need that.
+for f in $(find -maxdepth 1 -type f -o -type l); do
+    printf '  %s: %s\n' "${f#./}" "$(base64 --wrap=0 $f)"
```
allisonkarlitskaya, line 22: I was going to suggest "that thing you can do with shell parameter expansion that drops prefixes" as another option, but I was too lazy to look it up ;)

allisonkarlitskaya (approved):
Thanks for the tweaks.

allisonkarlitskaya (approved):

martinpitt:
I rolled this out to production, and spotted a few typos in the bits that run-local.sh can't reach (mostly OpenShift). I retried some tests in https://github.com/cockpit-project/cockpit/pull/20117 after the rollout, and they seem happy both on [e2e](https://cockpit-logs.us-east-1.linodeobjects.com/pull-20117-20240305-131158-fc59195a-fedora-39-firefox-other/log.html) and [rhos](https://cockpit-logs.us-east-1.linodeobjects.com/pull-20117-20240305-131157-fc59195a-fedora-coreos-networking/log.html).

PR #586, opened 2024-03-05
tasks: Create job-runner.toml in run-local.sh

martinpitt:
This still needs to happen for production, but allows us to start
testing https://github.com/cockpit-project/bots/pull/6017.

----
 
 - [ ] Requires #583


tasks/run-local.sh
```diff
@@ -178,6 +203,7 @@ EOF
     # use bash as pid 1 to mop up zombies
     podman run -d -it --name cockpituous-tasks --pod=cockpituous \
         -v "$SECRETS"/tasks:/run/secrets/tasks:ro,z \
+        -v "$SECRETS"/tasks/job-runner.toml:/work/.config/cockpit-dev/job-runner.toml:ro,z \
```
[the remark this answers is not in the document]
martinpitt: ah nice! Let's do that.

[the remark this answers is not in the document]
allisonkarlitskaya: One more, that I regret not mentioning during the last PR:

`job-runner.toml` has

```toml
[container]
command = 'podman-remote'
args = ['--url=unix:///run/podman/podman.sock']
```

...as a way to drop another envvar.

martinpitt:
@allisonkarlitskaya that change should be its own commit (or even PR), and also needs to add `--url` to `test_podman()` in run-local.sh.

tasks/run-local.sh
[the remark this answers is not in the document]
martinpitt: Sets JOB_RUNNER_CONFIG and mounts to /run/secrets/ now. The --url=unix bit is part of #587

tasks/run-local.sh
```diff
@@ -108,6 +108,55 @@ EOF
     fi
 }
 
+create_job_runner_config() {
+    # we never want to push to real GitHub branches in this test
+    run_args="'--volume=$MYDIR/mock-git-push:/usr/local/bin/git:ro'"
```
martinpitt, line 113: Let's add the --security-opt treatment for this.

ansible/roles/tasks-systemd/job-runner.toml
```diff
@@ -0,0 +1,25 @@
...
+token = [{file="/run/secrets/webhook/.config--github-token"}]
+
+[logs.s3]
+# bots lib/stores.py LOG_STORE
+url = 'https://cockpit-logs.us-east-1.linodeobjects.com/'
+key = [{file="/run/secrets/tasks/s3-keys/cockpit-logs.us-east-1.linodeobjects.com"}]
+
+[container]
+command = ['podman-remote', '--url=unix:///podman.sock']
+run-args = ['--device=/dev/kvm']
```
allisonkarlitskaya, line 14: That's going to need to get a bit bigger.  In addition to the issue about the cache dir you pointed out, at least these are missing:
 - `--memory=24g`
 - `--pids-limit=16384`
 - `--shm-size=1024m`
 - `${TMPVOL}` moral equivalent

martinpitt: Done. Confirmed in podman exec, and I also did a podman run to get the full joy of 8x parallelism: https://cockpit-logs.us-east-1.linodeobjects.com/test-podman-1/log.html .

ansible/roles/tasks-systemd/tasks/main.yml
```diff
@@ -40,6 +40,54 @@
...
+      command = ['podman-remote', '--url=unix:///podman.sock']
+      run-args = [
+          # resources
+          '--device=/dev/kvm',
+          '--memory=24g',
+          '--pids-limit=16384',
+          '--shm-size=1024m',
+          # qcow overlays on tmpfs
+          '--tmpfs=/tmp:size=14g',
+          '--env=TEST_OVERLAY_DIR=/tmp',
```
martinpitt, line 68: @allisonkarlitskaya FYI, I just added this so that we can eventually drop https://github.com/cockpit-project/cockpituous/blob/accb1fdc6aff599ab66a4915d0a3c952a48ccddb/tasks/Containerfile#L81 which is a bit ugly, too disconnected from the other half (i.e. the `--tmpfs`), and won't work with other containers.

I also reordered these by purpose instead of primarily on type, which makes it a bit easier to maintain/understand.

ansible/roles/tasks-systemd/tasks/main.yml
```diff
@@ -40,6 +40,54 @@
...
 
+- name: Create job-runner configuration
+  copy:
+    dest: /etc/job-runner.toml
+    content: |
+      [logs]
+      driver='s3'
+
+      [forge.github]
+      token = [{file="/run/secrets/webhook/.config--github-token"}]
```
allisonkarlitskaya, line 51: (feel free to say no) can you get us an extra `--volume` for /var/secrets/github-token and use it from that location?

ansible/roles/tasks-systemd/tasks/main.yml
```diff
@@ -40,6 +40,54 @@
...
+
+      [logs.s3]
+      # bots lib/stores.py LOG_STORE
+      url = 'https://cockpit-logs.us-east-1.linodeobjects.com/'
+      key = [{file="/run/secrets/tasks/s3-keys/cockpit-logs.us-east-1.linodeobjects.com"}]
```
allisonkarlitskaya, line 56: ditto for this (ie: `/run/secrets/s3-keys/cockpit-logs.us-east-1.linodeobjects.com`)

ansible/roles/tasks-systemd/tasks/main.yml
```diff
@@ -40,6 +40,54 @@
...
+          '--env=TEST_OVERLAY_DIR=/tmp',
+          # local image stores
+          '--env=COCKPIT_IMAGE_STORES_FILE=/config/image-stores',
+          '--volume=/var/cache/cockpit-tasks/image-stores:/config/image-stores:ro',
+          # generic secrets
+          '--volume=/var/lib/cockpit-secrets/tasks/npm-registry.crt:/run/secrets/tasks/npm-registry.crt:ro',
+          # various configuration
+          '--volume=/etc/npmrc:/etc/npmrc:ro',
+          '--volume=/var/cache/cockpit-tasks/images:/cache/images:rw',
+          '--env=TEST_JOBS={{ TEST_JOBS | default(8) }}',
```
allisonkarlitskaya, line 77: How about `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL`?  I'm not sure if we also need `COMMITTER` but I assume it'll default to author if it's not set?

ansible/roles/tasks-systemd/tasks/main.yml
allisonkarlitskaya: I feel like a bit more vertical whitespace would help keeping the sections visually separated from each other...

allisonkarlitskaya (approved):
This can go as-is if you disagree with my suggestions.  I have a feeling you'll be doing another round anyway, though...

tasks/run-local.sh
martinpitt: Done.

PR #587, opened 2024-03-06
Various cleanups

martinpitt:
This is a grab-bag of small unrelated stuff. @allisonkarlitskaya I took in and fixed your `$CONTAINER_HOST` change from #586. Can you please test `tasks/run-local.sh` with this? With any luck it fixes it. If not, I'm also happy to add `--security-opt=label=disable` instead like in commit 84e82706bb71bad36e36cd93783a553e79179808

### DIFF
--- a/tasks/Containerfile
+++ b/tasks/Containerfile
@@ -69,7 +69,7 @@ COPY cockpit-tasks install-service webhook github_handler.py /usr/local/bin/
 RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user --home-dir /work && \
     groupadd -g 1001 -r github && useradd -r --no-create-home -g github -u 1001 github && \
     mkdir -p /usr/local/bin /cache/images /cache/github && \
-    mkdir -p /work/.config /work/.config/cockpit-dev /work/.ssh /work/.cache /work/.rhel && \
+    mkdir -p /work/.ssh /work/.cache && \
     printf '[user]\n\t\nemail = cockpituous@cockpit-project.org\n\tname = Cockpituous\n[cockpit "bots"]\n\timages-data-dir = /cache/images\n' >/work/.gitconfig && \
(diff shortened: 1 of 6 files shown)


tasks/Containerfile
```diff
@@ -78,6 +78,7 @@ RUN groupadd -g 1111 -r user && useradd -r -g user -u 1111 user --home-dir /work
     echo 'user ALL=NOPASSWD: /usr/bin/chmod 666 /dev/kvm' > /etc/sudoers.d/user-fix-kvm
 
 ENV LANG=C.UTF-8 \
+    HOME=/work \
```
allisonkarlitskaya, line 81: That feels wrong.  This is going to have weird effects when we try to enter the container as `-u root`, for example.  Why doesn't this get set normally?

Checking this locally (against the current image, without this change):

```
$ podman run --rm quay.io/cockpit/tasks:latest sh -c 'echo $HOME'
/work
```

allisonkarlitskaya: Also: it occurs to me that now that we no longer have any "homedir path API" bindmounts, we could probably get rid of a great deal of the above special handling of creating `/work` and everything inside of it and let the homedir land in the default location.

Just about the only thing that looks actually useful is the gitconfig, and I have to think we could do something about that by setting some environment variables when we're running the bots... (or just set the environment variables on the container in the usual way when creating it)...

tasks/run-local.sh
```diff
@@ -174,7 +174,7 @@ EOF
         sleep 3
     done
 
-    # Run tasks container in the backgroud
+    # Run tasks container in the background
```
allisonkarlitskaya, line 177: Oh.  Good :)

allisonkarlitskaya (changes requested):

allisonkarlitskaya (approved):

PR #588, opened 2024-03-06
tasks: Add integration test for image-refresh via issue-scan

martinpitt:
This covers the real thing in between getting the "issue" webhook data
and uploading the image to the S3 server, and hence `issue-scan`,
`image-refresh`, `image-create`, S3 image credentials, etc.

Extend mock-github for the operations that `image-refresh` does, like
converting an issue to a pull, posting test statuses, etc. These aren't
asserted: it's  possible in principle with keeping state in the mock,
but these aren't the parts that are sensitive to us changing our bots
infrastructure. Also, these are better done in bots' unittests than this
integration test.

----

TODO:
 - [x] validate the uploaded image

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -246,20 +246,20 @@ test_image() {
 }
 
 test_mock_pr() {
-    podman cp "$MYDIR/mock-github-pr" cockpituous-tasks:/work/bots/mock-github-pr
+    podman cp "$MYDIR/mock-github" cockpituous-tasks:/work/bots/mock-github
     podman exec -i cockpituous-tasks sh -euxc "
         cd bots
         # test mock PR against our checkout, so that cloning will work
         SHA=\$(git rev-parse HEAD)
 
         # start mock GH server
-        PYTHONPATH=. ./mock-github-pr cockpit-project/bots \$SHA &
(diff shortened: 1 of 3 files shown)


tasks/run-local.sh
```diff
@@ -340,6 +342,76 @@ test_pr() {
...
+    LOGS_URL="$S3_URL_HOST/logs/"
+    CURL="curl --cacert $SECRETS/ca.pem --silent --fail --show-error"
+    LOG_MATCH="$($CURL $LOGS_URL| grep -o "image-refresh-foonux-[[:alnum:]-]*/log<")"
+    LOG="$($CURL "${LOGS_URL}${LOG_MATCH%<}")"
+    echo "--------------- mock image-refresh test log -----------------"
+    echo  "$LOG"
+    echo "--------------- mock image-refresh test log end -------------"
+    assert_in 'Running on:.*cockpituous' "$LOG"
+    assert_in './image-create.*foonux' "$LOG"
+    assert_in 'Uploading to' "$LOG"  # TODO: make this more precise
```
martinpitt: Ah, forgot about that -- fixed.

allisonkarlitskaya (approved):

PR #589, opened 2024-03-07
run-local.sh: Various cleanups, prepare for job-runner

martinpitt:
Broken out of #586 -- these parts apply to the current infra. I re-triggered https://github.com/cockpit-project/bots/pull/6017 and it is green now :green_heart: .

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -16,6 +16,8 @@ S3_URL_POD=https://localhost.localdomain:9000
 S3_URL_HOST=https://localhost.localdomain:$S3_PORT
 # AMQP address from inside the cockpituous pod
 AMQP_POD=localhost:5671
+# mock GitHub API running in tasks pod
+GHAPI_URL_POD="http://127.0.0.7:8443"
 
 # CLI option defaults/values
 PR=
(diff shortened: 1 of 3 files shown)


tasks/run-local.sh
```diff
@@ -257,8 +259,8 @@ test_mock_pr() {
         # start mock GH server
         PYTHONPATH=. ./mock-github cockpit-project/bots \$SHA &
         GH_MOCK_PID=\$!
-        export GITHUB_API=http://127.0.0.7:8443
-        until curl --silent \$GITHUB_API; do sleep 0.1; done
+        export GITHUB_API=$GHAPI_URL_POD
+        until curl --silent --fail \$GITHUB_API/repos/cockpit-project/bots; do sleep 0.1; done
```
allisonkarlitskaya, line 263: I almost commented here that hardcoding this was bad, for example, but that's obviously non-sense in this case.

Sometimes I think it would be better to have named this repository something different.  I always get confused about things on the real GitHub (which should allow for generalization based on which fork is running the tests) vs. our mocked-up stuff.

This is officially a drive-by.

PR #590, opened 2024-03-08
tasks: Lots of cleanup

martinpitt:
Rendered README: https://github.com/cockpit-project/cockpituous/blob/cleanups/tasks/README.md

 - [x] Tests blocked on https://github.com/cockpit-project/bots/pull/6035
 - [x] Blocked on finding a proper solution for #6045 , i.e. getting rid of the old machinery
 - [x] Rebase after #598, it's going to conflict (textually, not conceptually)
 - [x] [Refresh the container](https://github.com/cockpit-project/cockpituous/actions/runs/8262958933) .  ~Blocked on https://github.com/cockpit-project/bots/pull/6035 and https://github.com/cockpit-project/cockpit/pull/20149~
 - [x] Deploy this on OpenShift and validate
 - [x] Deploy this to rhos/e2e and validate


ansible/roles/tasks-systemd/tasks/main.yml
```diff
@@ -63,18 +63,29 @@
           '--memory=24g',
           '--pids-limit=16384',
           '--shm-size=1024m',
+
           # qcow overlays on tmpfs
           '--tmpfs=/tmp:size=14g',
           '--env=TEST_OVERLAY_DIR=/tmp',
+
+          # image cache
+          '--env=COCKPIT_IMAGES_DATA_DIR=/cache/images',
```
allisonkarlitskaya, line 72: I feel like I'm missing the corresponding change in `run-local.sh`

ansible/roles/tasks-systemd/tasks/main.yml
```diff
@@ -63,18 +63,29 @@
...
           '--volume=/var/cache/cockpit-tasks/image-stores:/config/image-stores:ro',
+
           # generic secrets
           '--volume=/var/lib/cockpit-secrets/tasks/npm-registry.crt:/run/secrets/tasks/npm-registry.crt:ro',
+
           # various configuration
           '--volume=/etc/npmrc:/etc/npmrc:ro',
-          '--volume=/var/cache/cockpit-tasks/images:/cache/images:rw',
           '--env=TEST_JOBS={{ TEST_JOBS | default(8) }}',
+          '--env=GIT_COMMITTER_NAME=Cockpituous',
```
allisonkarlitskaya, line 85: Ditto about `run-local.sh`

tasks/cockpit-tasks
```diff
@@ -45,19 +40,10 @@ for i in $(seq 1 30); do
     update_bots
     cd "$BOTS_DIR"
 
-    # avoid stale state and sockets
```
allisonkarlitskaya, line 48: I love that we don't have to worry about this sort of stuff anymore...

tasks/install-service
```diff
@@ -40,43 +38,29 @@ Requires=podman.socket
...
 # HACK: sometimes images get an MCS category which makes them undeletable by the container
-ExecStartPre=/usr/bin/chcon -R -l s0 \${TEST_CACHE}/images/
+ExecStartPre=/usr/bin/chcon -R -l s0 ${CACHE}/images/
 ExecStartPre=/usr/bin/flock /tmp/cockpit-image-pull podman pull quay.io/cockpit/tasks
-ExecStartPre=/usr/bin/podman network create cockpit-tasks-%i
 # idmapped mount would be better, but did not figure out how
 ExecStartPre=/usr/bin/chgrp 1111 %t/podman/podman.sock
+# job-runner doesn't need /images, but we still need it for the run-queue store-tests task
 ExecStart=/usr/bin/podman run --name=cockpit-tasks-%i --hostname=${CONTAINER_HOSTNAME} \
-    --device=/dev/kvm --network=cockpit-tasks-%i \
```
allisonkarlitskaya, line 61: This is starting to look nice :)

allisonkarlitskaya (changes requested):
commit message: consistentcy

commit message: we're also running tests-scan and friends directly in the main container (in addition to statistics)

ansible/roles/tasks-systemd/tasks/main.yml
martinpitt: It's [right here](https://github.com/cockpit-project/cockpituous/pull/590/commits/fe9946094e4366a2cd55ea833084f674fd89ef97#diff-889d7681af39462e3296875439af2aa5a56e43a0085c273b9e80d84b01b93b32R252), in this very commit.

ansible/roles/tasks-systemd/tasks/main.yml
martinpitt: Right -- doesn't do anything, but I added it for completeness. Same for the CentOS CI deployment.

martinpitt:
OK, this works now. I adjusted the commit messages, too. However, I'd like to queue it after #598, as it's gonna conflict (a little bit), and is less important.

martinpitt:
@allisonkarlitskaya This is the "first stage" review -- it may likely get some minor fixups when I deploy it, I just wanted us to agree before I change anything in prod

ansible/roles/tasks-systemd/tasks/main.yml
allisonkarlitskaya: Ah, I didn't understand how this works differently (ie: args given to outside container and then forwarded inside, with job-runner.toml created by shellscript).

I feel like it would be easier to understand (in `run-local.sh`) if the never-changing parts of the run-args are listed directly in the toml file and we only use the variable substitution for the parts we change when we're mocking.

allisonkarlitskaya (commented):
I think these are all nice cleanups — but let's see how the rollout goes.  Please see the comments about how job-runner.toml in run-local.sh could be easier to understand...

One thing I'd wish for is less user-setup stuff in the containerfile.

martinpitt:
Updated for two issues that I overlooked. The /dev/kvm was a funny one.. I deployed this now, and sent https://github.com/cockpit-project/bots/pull/6079 to confirm both an image refresh and a test run, and it's happy now.

PR #592, opened 2024-03-11
tasks: Test posted statuses in mock PR and image refresh

martinpitt:
Teach `mock-github` to write a log of POST requests, and add some initial assertions. They are not very precise yet (that'll get cleaned up in the Python rewrite soon), but at least ensure that we post the statuses to the expected SHA.

This enables us to create an integration test for a cross-project test in a future commit.

### DIFF
--- a/tasks/mock-github
+++ b/tasks/mock-github
@@ -19,10 +19,15 @@ from task.test_mock_server import MockHandler, MockServer
 
 repo = None
 sha = None
+log = None
 
 
 class Handler(MockHandler):
     def do_GET(self):
+        if log is not None:
+            print("GET", self.path, file=log)
+            log.flush()
+
         if self.path in self.server.data:
             self.replyJson(self.server.data[self.path])
         elif self.path.startswith(f'/repos/{repo}/pulls?'):
@@ -39,6 +44,10 @@ class Handler(MockHandler):
             self.send_error(404, 'Mock Not Found: ' + self.path)
 
     def do_POST(self):
+        if log is not None:
+            print("POST", self.path, self.rfile.read1().decode(), file=log)
+            log.flush()
+
         if self.path.startswith(f'/repos/{repo}/statuses/{sha}'):
             self.replyJson({})
         # new SHA from mock-pushed PR #2 for image-refresh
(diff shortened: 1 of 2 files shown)


tasks/mock-github
```diff
@@ -39,6 +40,10 @@ class Handler(MockHandler):
             self.send_error(404, 'Mock Not Found: ' + self.path)
 
     def do_POST(self):
+        if log:
+            print("POST", self.path, self.rfile.read1().decode(), file=log)
```
allisonkarlitskaya, line 44: ...why not (also) GET?

tasks/mock-github
```diff
@@ -39,6 +40,10 @@ class Handler(MockHandler):
             self.send_error(404, 'Mock Not Found: ' + self.path)
 
     def do_POST(self):
+        if log:
```
allisonkarlitskaya, line 43: `... is not None`

tasks/mock-github
```diff
@@ -19,6 +19,7 @@ from task.test_mock_server import MockHandler, MockServer
 
 repo = None
 sha = None
+log = None
```
allisonkarlitskaya, line 22: I wonder if you'd like to add some more types...

allisonkarlitskaya (approved):
I think this is fine.  Usual suggestions below.

I'm also not a fan of the tone of the second paragraph of the commit message (although this is something I feel like I've been guilty of in the past).  It's not immediately clear to me if the commit in question is doing the thing that it's talking about or just "enabling" it (with the expectation that it will be done in a future commit).

tasks/mock-github
martinpitt: This is mostly meant to validate actions. GET requests don't change the state, they are already sufficiently defined in terms of which paths `do_GET` handles in the test, and they are also somewhat of an implementation detail. I can add them to the log of course, but I wouldn't have anything to assert on them. For human inspection it's nice though, I agree.

Added.

tasks/mock-github
martinpitt: Done.

tasks/mock-github
martinpitt: TBH not today when we have a crapton of stuff to catch up with. I'd like to reserve this to the next weeks -- I also want to rewrite run-local.sh as proper pytest.

martinpitt:
@allisonkarlitskaya Reworded the commit message and fixed some of your suggestions. Thanks!

allisonkarlitskaya (approved):

PR #594, opened 2024-03-11
ansible: Define "image-download" job secret

martinpitt:
Tests will need that to download private (RHEL) images from the stores.

For now this is the same directory as the upload secret, but at some point we should split them.

----

This goes along with https://github.com/cockpit-project/bots/pull/6062

I added the secret splitting to the pilot board as enhancement, so that we don't forget.

I did *not* roll this out yet, will do after review.

### DIFF
--- a/ansible/roles/tasks-systemd/tasks/main.yml
+++ b/ansible/roles/tasks-systemd/tasks/main.yml
@@ -79,7 +79,12 @@
 
       [container.secrets]
       # these are *host* paths, this is podman-remote
+      # FIXME: Split the upload/download secrets
       image-upload=[
+          '--volume=/var/lib/cockpit-secrets/tasks/s3-keys/:/run/secrets/s3-keys:ro',
+          '--env=COCKPIT_S3_KEY_DIR=/run/secrets/s3-keys',
+      ]
+      image-download=[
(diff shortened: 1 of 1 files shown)


ansible/roles/tasks-systemd/tasks/main.yml
```diff
@@ -79,10 +79,15 @@
 
       [container.secrets]
       # these are *host* paths, this is podman-remote
+      # FIXME: Split the upload/download secrets
       image-upload=[
           '--volume=/var/lib/cockpit-secrets/tasks/s3-keys/:/run/secrets/s3-keys:ro',
           '--env=COCKPIT_S3_KEY_DIR=/run/secrets/s3-keys',
       ]
+      image-download=[
```
jelly, line 87: At first I was a bit confused, as alternative we could have `s3-bucket-token`. But this might be a bit too much of a bikeshed :)

jelly (approved):

ansible/roles/tasks-systemd/tasks/main.yml
martinpitt: The point is that we eventually want to have *different* tokens for reading (enough for tests) and writing (for image refreshes). So we should name them by purpose. Also, S3 is an implementation detail at this point.

PR #597, opened 2024-03-12
tasks: Add cockpituous refs/heads/main mock API endpoint

martinpitt:
job-runner needs that for cross-project testing [1]. A clean way would be to actually resolve "main" to a current SHA in the test, but as nothing actually cares, just rely on the `main` fallback for now -- nothing cares whether this is an actual SHA, as most things in git are happy with a "ref" -- and we don't use that to post statuses (and that would fail because the mock doesn't provide these paths).

[1] https://github.com/cockpit-project/bots/pull/6066

---

This unblocks https://github.com/cockpit-project/bots/pull/6066


martinpitt:
@allisonkarlitskaya If this is too smelly for you, I'm happy to put in the SHA resolution, but I'm more of a "build it only when needed" kind of guy :grin:

PR #598, opened 2024-03-12
Properly fix podman.socket permissions in tasks containers

martinpitt:

I rolled this out onto rhos-01-1, switched it back to `setenforce 1`, and confirmed that `sudo podman exec -it cockpit-tasks-1 podman-remote --url unix:///podman.sock ps` works and also doesn't leave any "AVC denied" in  `journalctl -f`.

### DIFF
--- a/ansible/psi/image-cache.yml
+++ b/ansible/psi/image-cache.yml
@@ -24,8 +24,8 @@
         src: self-hosted
         dest: "/var/lib/cockpit-secrets/tasks/s3-keys/{{ hostvars[groups['psi_s3'][0]].ansible_host }}"
         state: link
-        owner: 1111
-        group: 1111
+        owner: cockpituous
+        group: cockpituous
 
     - name: Tell tasks containers to drain and restart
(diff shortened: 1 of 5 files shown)


jelly (approved):
I can't say if the SELinux code is correct but if it's already deployed / tested then :+1:

martinpitt:
@jelly I only deployed/tested it on rhos-01-1 before, I wanted to wait for the review. I fully deployed it, and unfortunately I was [missing a trivial bit](https://github.com/cockpit-project/cockpituous/compare/f95744d9fb2fb557ced3b7b137f00c791a96a85e..83d7f8c2863e220e1254223720ddca716da16fee) -- the `file:` module doesn't know how to create the directory of the file you want to create. I didn't notice that as I manually created that directory for the initial experimentation, and forgot to delete it afterwards.

I rolled it out everywhere with that fix, and re-enabled SELinux on all boxes.

jelly (approved):

PR #600, opened 2024-03-13
Fix statistics processing in OpenShift tasks

martinpitt:
Sorry about that. I rolled this out, and confirmed that statistics processing now works.


allisonkarlitskaya (approved):
Sure.

PR #601, opened 2024-03-13
tasks: Stop reacting to opened issue

martinpitt:
This causes a duplicate image-refresh job for auto-generated image refresh issues. These generate both an "opened" and "labelled" event at the same tme, so that we'll get two `issues-scan` webhook queue entries, and the API query in `issue-scan` will see a "bot" labeled issue for both responses.

It is enough to react to the "labeled" event.


allisonkarlitskaya (approved):
Yes!  Thanks for finding the cause of this!

martinpitt:
... and deployed

PR #603, opened 2024-03-15
workflows: build a multi-arch tasks container

allisonkarlitskaya:
Build a tasks container for arm64 in addition to amd64.

We do this by splitting the build process into multiple jobs.  For the arm one, we make use of the arm runners from BuildJet.  At the end, we write the manifest and upload that with the "public" tag name.

I [kicked off a workflow run](https://github.com/cockpit-project/cockpituous/actions/runs/8294855484), which produced the 2024-03-15 tag [on quay](https://quay.io/repository/cockpit/tasks?tab=tags&tag=latest):
image

I'm taking that for a spin now in https://github.com/cockpit-project/starter-kit/pull/812


allisonkarlitskaya:
Out of all of the things that I thought might be difficult, I didn't imagine that it would be getting the `:latest` tag pushed.

Anyway — this is working, and we now have a `2024-03-15` and corresponding `latest` tag on quay, with the "temp" tags neatly tucked away in the [cockpit/tasks-tmp](https://quay.io/repository/cockpit/tasks-tmp?tab=history) repository.

.github/workflows/build-tasks.yml
```diff
@@ -1,21 +1,68 @@
...
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        build:
+          - label: amd64
+            runner: buildjet-2vcpu-ubuntu-2204
+          - label: arm64
+            runner: buildjet-2vcpu-ubuntu-2204-arm
```
martinpitt, line 22: I'm surprised that we don't need any secret/token for this? They just bill it to the org?

.github/workflows/build-tasks.yml
```diff
@@ -1,21 +1,68 @@
...
+          file: tasks/container/Containerfile
 
-      - name: Log into container registry
-        run: podman login -u ${{ secrets.QUAY_BOTUSER }} -p ${{ secrets.QUAY_TOKEN }} quay.io
+  manifest:
+    needs: build
 
-      - name: Build tasks container
-        run: make tasks-container
+    runs-on: buildjet-2vcpu-ubuntu-2204
```
martinpitt, line 52: OOI, does that not work on the regular GH runners?

martinpitt (commented):
This looks great, thanks for figuring out all the details!

I'm a bit concerned about the [integration test failure](https://github.com/cockpit-project/cockpituous/actions/runs/8296339259/job/22705177096?pr=603) -- although it's not quite clear what this has to do with setting an empty `LOG_MATCH`. As a first thing, perhaps you could do a no-change force-push to run them again? (We can't just retry as the "unit-tests" context already exists).

.github/workflows/build-tasks.yml
allisonkarlitskaya: No.  I signed us up and we get an initial 5$ credit.  We need to add a credit card once that runs out.

.github/workflows/build-tasks.yml
allisonkarlitskaya: It does.  I used buildjet because I was getting tired of waiting for the GH runners to pick up the job (they were being slow today).  I'm happy to put it back, if you prefer.

.github/workflows/build-tasks.yml
allisonkarlitskaya: I wasn't totally clear: as part of the signup, I added their app to our org.

.github/workflows/build-tasks.yml
martinpitt: Nah, it's fine, I was just curious.

martinpitt (approved):
✈️

PR #604, opened 2024-03-18
workflows: move tasks container to ghcr.io

allisonkarlitskaya:
This speeds things up (due to having the runner closer to the registry) and simplifies credentials management: we'll be able to drop our quay.io environment after this.

Move the amd64 and manifest jobs back to GitHub so they can run on GitHub infra, closer to the container registry.

I already did a run of this and it worked nicely.  See https://github.com/cockpit-project/cockpituous/actions/runs/8307247489 and https://github.com/orgs/cockpit-project/packages

I also started to try porting the container-file-updater-script to GitHub's API.  That was fairly straight-forward, but with an important caveat: in its first iteration, it was not possible to use ghcr at all without a GitHub token, even on public repos.  Since then, they've carved out enough of an exception to allow `podman pull` to function, but iterating over available tags via the REST API (or the docker protocol one) is still not possible without a token... :(

### DIFF
--- a/.github/workflows/build-tasks.yml
+++ b/.github/workflows/build-tasks.yml
@@ -8,8 +8,8 @@ on:
 # where different runners pick different tags (eg: for builds near midnight, or
 # in different time zones).  The manifest job picks the final name.
 env:
-  tag: quay.io/cockpit/tasks
-  tmptag: quay.io/cockpit/tasks-tmp:${{ github.run_id }}.${{ github.run_attempt }}
+  tag: ghcr.io/${{ github.repository_owner }}/tasks
+  tmptag: ghcr.io/${{ github.repository_owner }}/tasks-tmp:${{ github.run_id }}.${{ github.run_attempt }}
 
 jobs:
   build:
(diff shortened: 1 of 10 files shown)


martinpitt:
A bit early to review the precise diff, as this needs lots of other places changed (Ansible, `install-service`, integration tests, documentation, etc.) But I do agree to moving it to ghcr.io.

@allisonkarlitskaya how about this, to avoid having to log in?

```
❱❱ skopeo list-tags --tls-verify --no-creds docker://ghcr.io/cockpit-project/tasks
{
    "Repository": "ghcr.io/cockpit-project/tasks",
    "Tags": [
        "2024-03-16",
        "latest"
    ]
}
```

allisonkarlitskaya:
> A bit early to review the precise diff, as this needs lots of other places changed (Ansible, `install-service`, integration tests, documentation, etc.) But I do agree to moving it to ghcr.io.
> 
> @allisonkarlitskaya how about this, to avoid having to log in?
> 
> ```
> ❱❱ skopeo list-tags --tls-verify --no-creds docker://ghcr.io/cockpit-project/tasks
> {
>     "Repository": "ghcr.io/cockpit-project/tasks",
>     "Tags": [
>         "2024-03-16",
>         "latest"
>     ]
> }
> ```

This is wonderful, thanks!!

martinpitt (approved):
LGTM! After this and https://github.com/cockpit-project/bots/pull/6095 we need to update our projects to pull in the new tasks container, and updating their container update workflow for the renamed script.

martinpitt:
I deployed that change to our infra now.

PR #605, opened 2024-03-18
tasks: Speed up test_pr()

martinpitt:
The `[ -t 0 ]` check in `cockpit-tasks` doesn't do anything any more, as the integration tests now run that script with `podman exec`, without an attached tty. This is also a bit awkward for the upcoming pytest rewrite, so replace it with an environment variable.

### DIFF
--- a/tasks/run-local.sh
+++ b/tasks/run-local.sh
@@ -435,7 +435,7 @@ test_pr() {
 
     # run the main loop in the background; we could do this with a single run-queue invocation,
     # but we want to test the cockpit-tasks script
-    podman exec -i cockpituous-tasks cockpit-tasks &
+    podman exec -i --env=SLUMBER=1 cockpituous-tasks cockpit-tasks &
     TASKS_PID=$!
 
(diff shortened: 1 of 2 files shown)


tasks/run-local.sh
```diff
@@ -435,7 +435,7 @@ test_pr() {
 
     # run the main loop in the background; we could do this with a single run-queue invocation,
     # but we want to test the cockpit-tasks script
-    podman exec -i cockpituous-tasks cockpit-tasks &
+    podman exec -i --env=SLUMBER=5 cockpituous-tasks cockpit-tasks &
```
allisonkarlitskaya, line 438: ...or 1 :)

allisonkarlitskaya (dismissed):

tasks/run-local.sh
martinpitt: err, right :sweat_smile: Done

allisonkarlitskaya (approved):
yip

PR #609, opened 2024-03-21
tasks: Install types-PyYAML into container

martinpitt:
This is a prerequisite for proper mypy testing of the upcoming pytest in PR #606

### DIFF
--- a/tasks/container/Containerfile
+++ b/tasks/container/Containerfile
@@ -67,7 +67,7 @@ RUN dnf -y update && \
     dnf -y builddep --setopt=install_weak_deps=False /tmp/cockpit.spec && \
     rm /tmp/cockpit.spec && \
     dnf clean all && \
-    pip install ruff
+    pip install ruff types-PyYAML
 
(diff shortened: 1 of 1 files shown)


martinpitt:
@allisonkarlitskaya Heh, I understood. In the end this is tiny, so let's do that. More fodder for our container-update bots :grin:

tasks/container/Containerfile
```diff
@@ -67,7 +67,7 @@ RUN dnf -y update && \
     dnf -y builddep --setopt=install_weak_deps=False /tmp/cockpit.spec && \
     rm /tmp/cockpit.spec && \
     dnf clean all && \
-    pip install ruff
+    pip install ruff types-PyYAML
```
allisonkarlitskaya, line 70: Okay, but one more and we start using the `\`s :)

allisonkarlitskaya (approved):

tasks/container/Containerfile
martinpitt: haha, deal!

PR #610, opened 2024-04-01
ansible: Move npm CA file back to env variable

martinpitt:
Commit dbd7643e5c958 caused a regression: The `cafile` npmrc option *only* uses the given CA, while `$NODE_EXTRA_CA_CERTS` uses the given CA *in addition*. That broke fetching npm modules directly from github.com, which needs the system default CA lists.

---

Blocks https://github.com/cockpit-project/bots/pull/6148 and also https://github.com/cockpit-project/bots/pull/6160. See the former for debugging notes.


allisonkarlitskaya (approved):
Oh look, yaml!

:eyes:

PR #615, opened 2024-04-08
tasks: Introduce elastic cloud runner mode

martinpitt:
As the EC2 bare metal instances are expensive, we don't want them to run
permanently, but only in times of high demand. They should then
terminate themselves when the queue is running out of work.

Define "run out of work" as "the number of job-runner entries in the
AMQP queue drops below 10". At that level, our permanent PSI runners can
keep up. This is are more robust global criterion than checking if
`run-queue` encountered an empty queue, as that is more prone to
terminating only *some* of the instances while some others keep picking
up brand new queue entries.

Introduce an "idle poweroff" mode in which the `cockpit-tasks` main loop
exits with code 100 instead of slumbering when work is running low.
Configure the slice to automatically power off the machine once all
cockpit-tasks instances exited cleanly (we don't want this on failures,
so that we can ssh in and examine them). Use the `poweroff-immediate`
heavy hammer there, to avoid potential hangs on shutdown -- there is
nothing to rescue from the instance anyway.

Plumb that through the AWS Ansible role and document it.

-----

There was surprisingly little to fix -- only a self-inflicted incompatibility with Fedora CoreOS.

I tested the auto-poweroff approach locally with the new "localvm" playbook. The AMQP queue is currently empty, so the containers ran out of work and the VM powered off immediately.

 - [x] built on top of #616
 - [ ] [build the tasks container](https://github.com/cockpit-project/cockpituous/actions/runs/8594245809) to pick up the `IDLE_POWEROFF` handling
 - [x] Test-deploy this to EC2 in standard mode (it's been a while and it might need more adjustments)
 - [ ] Figure out why metal instance doesn't terminate after poweroff
 - [ ] Test-deploy in `-e idle_poweroff=1` mode

### DIFF
--- a/ansible/aws/README.md
+++ b/ansible/aws/README.md
@@ -57,6 +57,8 @@ Create and configure the instance:
 
 If you run more than one at a time, set a custom host name with `-e hostname=cockpit-aws-tasks-2` or similar, so that GitHub test statuses remain useful to identify where a test runs.
 
+There is also an "elastic" mode where the tasks bots keep running until the AMQP queue runs low. Use that for situations where AWS instances act as extra high-demand capacity instead of being the primary runners. Enable that mode with `-e idle_poweroff=1`.
+
 Webhook setup
 -------------
 AWS runs our primary webhook. Deploy or update it with:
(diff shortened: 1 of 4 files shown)


martinpitt:
Meh - I tested `idle_poweroff` on AWS, and while it does power down the  machine, it doesn't actually terminate it, i.e. the `instance_initiated_shutdown_behavior: terminate` flag doesn't seem to work on metal instances? :cry: 

Putting this on the shelf then, but we at least need the other three commits.

martinpitt:
This was already proposed two years ago in #477, and Lis didn't like it. So, manual for now.

PR #616, opened 2024-04-08
Fix EC2 instances, add local tasks playbook testing

martinpitt:
I rolled this out to EC2 and validated that the task runner comes up fine. With the demise of our e2e machies we'll most probably need this soon.


allisonkarlitskaya (approved):
I love the smell of yaml in the morning

PR #618, opened 2024-04-08
Split s3 server certificate out of tasks secrets

martinpitt:
Tasks containers don't need, and therefore should not have, the local minio S3 server certificate.

Our ci-secrets.git repo already moved the S3 certificate out of `tasks/` into the top level `s3-server/` directory. Adjust the deployment scripts and integration test accordingly.


martinpitt:
I deployed this to production, and the image server on rhos-01-1 works fine. The cert is also gone from cockpit-tasks* containers.

martinpitt:
@allisonkarlitskaya gentle review ping?

PR #619, opened 2024-04-23
Fedora 40, install ruff from distro packages

allisonkarlitskaya:
- [x] rebuild tasks container when landing: https://github.com/cockpit-project/cockpituous/actions/runs/8831813342

### DIFF
--- a/tasks/container/Containerfile
+++ b/tasks/container/Containerfile
@@ -50,12 +50,14 @@ RUN dnf -y update && \
         python3-pytest-asyncio \
         python3-pytest-timeout \
         python3-pyyaml \
+        python3-types-pyyaml \
         python3-vulture \
         python3-wheel \
         qemu-kvm-core \
         rpm-build \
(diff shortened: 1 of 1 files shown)


tasks/container/Containerfile
```diff
@@ -50,12 +50,14 @@ RUN dnf -y update && \
         python3-pytest-asyncio \
         python3-pytest-timeout \
         python3-pyyaml \
+        python3-types-pyyaml.noarch \
```
martinpitt, line 53: The .noarch suffix is weird here, let's please drop that.

martinpitt (changes requested):
Cool, I wanted to do the same thing this morning :grin:

tasks/container/Containerfile
allisonkarlitskaya: Copy-paste error.

martinpitt (approved):
Cheers!

PR #620, opened 2024-06-11
tasks: container: add python3-pcp as dependency for tests

jelly:
Required for: https://github.com/cockpit-project/cockpit/pull/20049

### DIFF
--- a/tasks/container/Containerfile
+++ b/tasks/container/Containerfile
@@ -43,6 +43,7 @@ RUN dnf -y update && \
         python3-build \
         python3-flake8 \
         python3-mypy \
+        python3-pcp \
         python3-pika \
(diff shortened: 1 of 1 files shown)


tasks/container/Containerfile
```diff
@@ -43,6 +43,7 @@ RUN dnf -y update && \
         python3-build \
         python3-flake8 \
         python3-mypy \
+	python3-pcp \ # Temporary PCP dependency until we port pcp to the Python bridge and add it in the spec file
```
allisonkarlitskaya, line 46: I don't think that this is legit...

jelly:
https://github.com/cockpit-project/cockpituous/pull/621

PR #621, opened 2024-06-11
tasks: container: temporarily add python3-pcp as dependency for tests

jelly:
Required for: https://github.com/cockpit-project/cockpit/pull/20049


martinpitt (approved):
Not sure about the "temporarily", but it's not important.

PR #623, opened 2024-10-04
tasks: Install lorax

KKoukiou:
This is required by anaconda tests as they are using `mkksiso` to adjust the test ISO file. [1]

[1]: https://github.com/rhinstaller/anaconda-webui/pull/462


mvollmer:
The linting issues in the unit tests are unrelated, but worth investigating.

jelly (approved):
Linting failure is due to a newer ruff in the GitHub workflow but our tasks container uses an older ruff so this is sadly impossible to fix for now.

PR #626, opened 2024-10-22
tasks: Fix image cache in local podman container, fix ruff 0.7 errors

martinpitt:
These days we have to explicitly set `$COCKPIT_IMAGES_DATA_DIR`.

Drop the explicit "bash" command, it's not necessary any more.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,3 +62,7 @@ select = [
     "W",       # warnings (mostly whitespace)
     "YTT",     # flake8-2020
 ]
+
+ignore = [
+    "PT001", "PT023", # different versions can't make up their mind how to write decorators
+]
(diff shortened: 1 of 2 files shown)


martinpitt:
Wohoo, we haven't run cockpituous CI tasks in a while -- [new ruff fun](https://github.com/cockpit-project/cockpituous/actions/runs/11453587734/job/31866312529?pr=626). Fixed (ruff could do it automatically)

martinpitt:
[Meh](https://github.com/cockpit-project/cockpituous/actions/runs/11453642210/job/31866463521?pr=626) -- this now says exactly the opposite. I cannot reproduce this, the tasks container's ruff-0.6.7-1.fc40.x86_64 says "all passed". Let's disable that rule.

jelly (approved):

PR #627, opened 2024-11-25
ansible: Temporarily disable RH internal NPM mirror

martinpitt:
It's been down since yesterday, see https://issues.redhat.com/browse/SPRE-880

---

See https://github.com/cockpit-project/cockpit-ostree/pull/738 and https://github.com/cockpit-project/bots/pull/7136 

We don't necessarily have to land this, I am just filing this for transparency. I rolled this out.


martinpitt:
Let's use this for the pilot board to re-enable the internal mirror once it's fixed.

martinpitt:
Wow, that was a fast fix. I re-deployed main, and this is obsolete.

PR #629, opened 2024-12-06
workflows: Auto-refresh tasks container every week

martinpitt:
We tend to forget about refreshing the tasks container, and thus fall
behind on browser and toolchain updates.

Run these every Saturday, so that they are ready for the Sunday night
runs of our projects' tasks-container-update workflows.

### DIFF
--- a/.github/workflows/build-tasks.yml
+++ b/.github/workflows/build-tasks.yml
@@ -26,18 +26,17 @@ jobs:
 
     steps:
       # NB: no podman on buildjet arm runners
-      -
-        name: Log in to container registry
+      - name: Log in to container registry
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      -
-        name: Checkout
+
+      - name: Checkout
         uses: actions/checkout@v4
-      -
-        name: Build and push
+
+      - name: Build and push
         uses: docker/build-push-action@v5
         with:
           push: true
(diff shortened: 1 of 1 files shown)


martinpitt:
If this lands today, it can run tomorrow morning, and I'll watch/fix up the results on Monday.

martinpitt:
> This is all gated in the end right?

Yes, all our projects have a .cockpit-ci/container which pins down a particular tag. And a "tasks-container-update" workflow which tries to update to the new one, and will tell us about changed browser pixel tests or other incompatibilities. I.e. gated on the project level.

> Just wondering how we detect deploy failures?

Yes, at least I get email about it. Also, even if nobody listens, we are no worse off than right now -- every now and then the pilot updates it manually, and then on https://github.com/cockpit-project/cockpituous/actions/workflows/build-tasks.yml we see the failures.

jelly:
> > This is all gated in the end right?
> 
> Yes, all our projects have a .cockpit-ci/container which pins down a particular tag. And a "tasks-container-update" workflow which tries to update to the new one, and will tell us about changed browser pixel tests or other incompatibilities. I.e. gated on the project level.
> 
> > Just wondering how we detect deploy failures?
> 
> Yes, at least I get email about it. Also, even if nobody listens, we are no worse off than right now -- every now and then the pilot updates it manually, and then on https://github.com/cockpit-project/cockpituous/actions/workflows/build-tasks.yml we see the failures.

Sounds good, I just don't want the burden of keeping track of this to fall all on you. So I'll try to keep an eye and put it on the pilot board when relevant.

PR #634, opened 2024-12-16
tasks: Delay webhook startup; metrics: Adjust config to Prometheus 3

martinpitt:
Some fallout from my CI wrestling today (for OpenSSL 3 certs, this was unrelated Yak shaving).

### DIFF
--- a/metrics/metrics.yaml
+++ b/metrics/metrics.yaml
@@ -94,9 +94,6 @@ data:
       alertmanagers:
       - static_configs:
         - targets: []
-        scheme: http
-        timeout: 10s
-        api_version: v1
     scrape_configs:
     - job_name: ci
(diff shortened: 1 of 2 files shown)


jelly (approved):
The sleep 10 feels a bit hacky, couldn't the webhook just retry a few times. Anyway, lgtm.

martinpitt:
> couldn't the webhook just retry a few times

Yes, absolutely -- this was a quick hack to unbreak the setup. I'll look into updating the webhook script itself. It already has a retry loop, which apparently doesn't work. Dang, I should have captured the precise backtrace..

martinpitt:
See commit 06c2717be744b9281f76934b524c66b73421a40f. I put it on my TODO list to fix this properly.

PR #636, opened 2024-12-17
webhook: Retry OSError on queue connection failure

### LINKED ISSUE #635: amqp/webhook startup race
Starting the AMQP and webhook container in parallel very often runs into webhook not yet being able to connect to the AMPQ server, and thus getting trapped in a crash loop until it wins the race.

This happens when deploying `oc create -f tasks/cockpit-tasks-webhook.yaml` to OpenShift.

Yesterday I did a quick hack in commit 169b53bac1998ed59997e2c27cd50baf92502d91 to delay the webhook startup. I already tried to fix that in commit 06c2717be744b9281f76934b524c66b73421a40f two years ago, but yesterday it happened again. I was distracted by a lot of other fires, so I didn't capture the precise error. Let's fix this properly.

martinpitt:
Commit 06c2717be74 introduced a retry loop for connecting to AMQP, as
these two containers start at the same time in the OpenShift deployment.

However, these days the webhook often fails with `ssl.SSLEOFError`
(with parents `SSLError` and `OSError`), which doesn't fit `AMQPError`.
So intercept this as well.

With that we can revert the static sleep from commit 169b53bac1998
again.

Fixes #635

----

 - [x] Rebuild container: https://github.com/cockpit-project/cockpituous/actions/runs/12369349737
 - [x] Quick local run (we just did the last refresh yesterday):
 - [x] Redeploy on OpenShift and ensure that it works:

### DIFF
--- a/tasks/cockpit-tasks-webhook.yaml
+++ b/tasks/cockpit-tasks-webhook.yaml
@@ -39,7 +39,7 @@ spec:
           ports:
             - containerPort: 8080
               protocol: TCP
-          command: [ "sh", "-ec", "sleep 10; exec webhook" ]
+          command: [ "webhook" ]
           volumeMounts:
           - name: webhook-secrets
             mountPath: /run/secrets/webhook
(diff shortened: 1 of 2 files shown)


jelly (approved):
Thanks

martinpitt:
Rebuilt, tested locally, deployed, works :heavy_check_mark:

PR #637, opened 2025-01-20
tasks: Install full chromium

martinpitt:
The headless version has several bugs, at least [1] and [2]. Also, headless and full Chromium render texts slightly differently, so we can't debug/fix pixel tests with `TEST_SHOW_BROWSER=pixels`.

This is a net 100 MB size increase of the tasks container, and will break all pixel tests once, but long-term it is a better browser.

[1] https://issues.redhat.com/browse/COCKPIT-1159
[2] https://github.com/cockpit-project/cockpit-files/pull/911#issuecomment-2602064380

### DIFF
--- a/tasks/container/Containerfile
+++ b/tasks/container/Containerfile
@@ -7,7 +7,7 @@ RUN dnf -y update && \
         adobe-source-code-pro-fonts \
         byobu \
         chromedriver \
-        chromium-headless \
+        chromium \
         curl \
         dbus-daemon \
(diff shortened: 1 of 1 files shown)


martinpitt:
Building in https://github.com/cockpit-project/cockpituous/actions/runs/12866891639

martinpitt:
Doing the first refresh in https://github.com/cockpit-project/cockpit-files/pull/912

PR #639, opened 2025-01-23
Fedora wiki report deps

KKoukiou:


martinpitt (approved):
Thanks! I rebuild the tasks container now, then you need to pull in the new tag into a-webui

PR #644, opened 2025-02-26
ansible: Add fedora-wiki-staging tasks secret

KKoukiou:
For posting Anaconda nightly compose test results to the Fedora Wiki staging instance.


martinpitt (approved):
Thanks! I commited the new secret and rolled it out. I'll roll out that new config.

PR #652, opened 2025-04-16
Fedora 42 upgrade

jelly:
This is semi-deployed: nodes 38-42 are now Fedora 42 the rest is still Fedora 40.


martinpitt (approved):
Progress! Part of me wants NUM=42, though :grin:

jelly:
This is now almost deployed, I only haven't replaced rhos-01-1 yet.

PR #654, opened 2025-05-19
mock-github: adjust .cockpit-ci/container location

jelly:
Since githubusercontent is now rate limited bots changed to obtaining the .cockpit-ci/container file from the GitHub rest API.

### DIFF
--- a/tasks/mock-github
+++ b/tasks/mock-github
@@ -11,6 +11,7 @@
 # first to process webhook → tests-scan → public, second to actually run it
 
 import argparse
+import base64
 import json
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


tasks/mock-github
```diff
@@ -40,6 +40,8 @@ class Handler(MockHandler):
             })
         elif self.path == f'/{repo}/{sha}/.cockpit-ci/container':
             self.replyData('ghcr.io/cockpit-project/tasks')
+        elif self.path == f'/repos/{repo}/contents/.cockpit-ci/container?ref={sha}':
+            self.replyData('ghcr.io/cockpit-project/tasks')
```
allisonkarlitskaya, line 44: That seems wrong.  Isn't it going to be expecting a JSON document?

jelly: It is, I made a mess last night :)

tasks/mock-github
```diff
@@ -40,6 +41,10 @@ class Handler(MockHandler):
             })
         elif self.path == f'/{repo}/{sha}/.cockpit-ci/container':
             self.replyData('ghcr.io/cockpit-project/tasks')
```
jelly, line 43: Kept the old path because I think we can't easily land this green otherwise.

allisonkarlitskaya: I agree, but maybe you change the subject of the commit message to reflect that you're adding a new endpoint...

allisonkarlitskaya (approved):

PR #657, opened 2025-06-17
tasks: Remove unnecessary packages

martinpitt:
- dbus-glib and libXt were added in commit 23afb2d197f0e9 for firefox nightly builds. But we've switched back to the Fedora package long ago which explicitly depends on these.

 - expect hasn't been used by our image build scripts any more for ages.

 - gcc-c++ and gnupg aren't used by any current project.

 - git-core is enough, we don't need the documentation and extra modules.

### DIFF
--- a/tasks/container/Containerfile
+++ b/tasks/container/Containerfile
@@ -10,20 +10,15 @@ RUN dnf -y update && \
         chromium \
         curl \
         dbus-daemon \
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested this locally, admittedly only with c-files. But these seem harmless enough, and we get the projects gated via container update PRs anyway.

PR #668, opened 2025-09-10
tasks: add vhost-device-vsock to the container

allisonkarlitskaya:
test.thing wants to depend on this to avoid using vsock on the host machine, avoiding all kids of problems.


allisonkarlitskaya:
New container is built and published:
 - https://github.com/cockpit-project/cockpituous/actions/runs/17617041079
 - https://github.com/cockpit-project/cockpituous/pkgs/container/tasks/511467178?tag=2025-09-10

PR #671, opened 2025-09-15
ansible: set `post = true` in job-runner.toml

allisonkarlitskaya:
We're about to change the default of this to `false` to enable more of the job-runner internals to be used from other scripts.


martinpitt (approved):
Assuming green

PR #677, opened 2025-11-13
local-s3: Switch to Hummingbird minio containers

martinpitt:
The official image is deprecated and not updated any more [1][2], so we need to
switch to a maintained image. Project Hummingbird just got a new one [3], so
switch to that.

This requires some adjustments to volume paths, as the Hummingbird image runs
as non-root user and thus does not have `HOME=/root`.

Related: https://issues.redhat.com/browse/HUM-167

[1] https://github.com/minio/minio/commit/9e49d5e7a648f00e
[2] https://github.com/minio/minio/issues/21647
[3] https://quay.io/repository/hummingbird/minio

----

Goes on top of #676. I rolled this out and verified on rhos-01-2 with `sudo podman exec -it -w /work/bots cockpit-tasks-1 python3 -m lib.s3 ls http://10.0.203.212/images/`.

### DIFF
--- a/local-s3/install-s3-service
+++ b/local-s3/install-s3-service
@@ -14,7 +14,7 @@ systemctl stop cockpit-s3.service || true
 
 if [ -z "${DISABLE_TLS:-}" ]; then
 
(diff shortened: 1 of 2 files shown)


jelly (approved):
Thanks!

PR #695, opened 2026-05-07
Switch to Hummingbird community images

jelly:
The old images are no longer available publicly.

### DIFF
--- a/ansible/roles/webhook/cockpituous-webhook.service
+++ b/ansible/roles/webhook/cockpituous-webhook.service
@@ -22,7 +22,7 @@ ExecStart=/usr/bin/podman run \
     --tmpfs /var/lib/rabbitmq \
     -v /etc/rabbitmq:/etc/rabbitmq:ro,z \
     -v /var/lib/cockpit-secrets/webhook:/run/secrets/webhook:ro,z \
-    docker.io/rabbitmq
+    docker.io/rabbitmq:4.2.6
 
 ExecStart=/usr/bin/podman run \
     --cgroups=no-conmon \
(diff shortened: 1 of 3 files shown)


jelly:
```
pika.exceptions.ConnectionClosedByBroker: (541, 'INTERNAL_ERROR - Feature `transient_nonexcl_queues` is deprecated.\nBy default, this feature is not permitted anymore.\nThe feature will be removed from a future major RabbitMQ version, regardless of the configuration; actual version to be determined.\nTo...')
```

Fun never stops

ansible/roles/webhook/cockpituous-webhook.service
```diff
@@ -22,7 +22,7 @@ ExecStart=/usr/bin/podman run \
     --tmpfs /var/lib/rabbitmq \
     -v /etc/rabbitmq:/etc/rabbitmq:ro,z \
     -v /var/lib/cockpit-secrets/webhook:/run/secrets/webhook:ro,z \
-    docker.io/rabbitmq
+    docker.io/rabbitmq:4.2.6
```
martinpitt, line 25: Oops! Our currently deployed RabbitMQ on CentOS CI is 4.2.5, it's been running since April 23. So apparently the current 4.3.0 breaks API? Let's land this for the time being to get back to a status quo (although I'd prefer a `:4.2` tag) and I'll see if this is fixable with 4.3. If not, I'll send a follow-up for loosening the tag.

martinpitt (approved):
Thanks!

ansible/roles/webhook/cockpituous-webhook.service
martinpitt: See https://www.rabbitmq.com/docs/queues#durability which confirms exactly that, non-durable queues have been deprecated since 2021. Fixed in https://github.com/cockpit-project/bots/pull/9007 , after that lands, we can revert the tag pin.

PR #697, opened 2026-05-11
aws: Update tags according to current policy

martinpitt:
See https://source.redhat.com/departments/strategy_and_operations/it/datacenter_infrastructure/itcloudservices/itpubliccloudpage/cloud/itpc_annoucements/updated_it_public_cloud_tagging_policy

Thanks to @thrix for pointing this out!


jelly (approved):
I need to locally test my AWS access, but changes look good

PR #698, opened 2026-05-12
Move S3 log store to our AWS account

martinpitt:
Part of https://github.com/cockpit-project/bots/issues/9003 and https://redhat.atlassian.net/browse/COCKPIT-1833

I already ran the new bucket deployment playbook. @jelly, @allisonkarlitskaya you (and everyone else with Cockpit Bitwarden access) should already be able to do that. These are the same credentials as for the `aws` CLI (Fedora package `awscli2`), which allows you to do everything the web UI can. Access to the latter still has to be set up at least for @jelle, not sure about @allisonkarlitskaya

I did *not* yet deploy the actual switch-over from the topmost commit. That goes together with https://github.com/cockpit-project/bots/pull/9013 and both should be reviewed/approved first. Then the landing and deployment need to happen in lockstep. I'm fine to do that, if either of you wants to do that, even better of course.

See https://github.com/cockpit-project/bots/issues/9003 for the master plan and details how I tested this.

 - [x] Deploy S3 secrets
 - [x] Deploy `setup-s3-buckets.yml`
 - [x] Land after https://github.com/cockpit-project/bots/pull/9013
 - [x] Deploy new tasks and metrics config
   - [x] tasks
   - [x] metrics

### DIFF
--- a/ansible/aws/README.md
+++ b/ansible/aws/README.md
@@ -1,23 +1,57 @@
 Resources for setting up CI in AWS
 ==================================
 
-Initial URLs
-------------
+AWS user management console
+---------------------------
 
- * Sign in [with your kerberos ticket](https://auth.redhat.com/auth/realms/EmployeeIDP/protocol/saml/clients/itaws); this needs to be set up first for new users, ask Dominik Perpeet or Miroslav Vadkerti about it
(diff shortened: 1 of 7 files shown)


allisonkarlitskaya:
I just redeployed all the tasks runners out of this branch and it's working swimmingly...

allisonkarlitskaya (approved):
Everything is now deployed.  Thanks!

PR #699, opened 2026-05-12
ansible: Add maintenance playbook for updating users

martinpitt:
For rolling out changes to ssh-keys. I previously did some ad-hoc `ansible -m include_role` commands for this, but this is easier and more robust.

---

I tested this with #700 and it worked fine.


allisonkarlitskaya (approved):
Thanks!

PR #701, opened 2026-05-12
Document CentOS CI OpenShift access

martinpitt:


allisonkarlitskaya (commented):
Thanks!  This was definitely a missing corner.
I can login but I can't currently see anything, so let's try to get that resolved first...  @jelly ?

martinpitt:
I filed https://forge.fedoraproject.org/infra/tickets/issues/13340 to get you access to CentOS CI cluster. Maybe @jelly already has it?

martinpitt:
I updated the documentation to point to [the magic group](https://accounts.fedoraproject.org/group/ocp-cico-cockpit/). So I think this can land now.

centos-ci.md
```diff
@@ -0,0 +1,42 @@
...
+persistent/structural changes there -- all deployments happen via YAML files,
+see the per-directory README.md files.
+
+## Members
+
+Access to the project is provided through the [`cockpit` Fedora group](https://accounts.fedoraproject.org/group/ocp-cico-cockpit/).
+
+## Administrators/contact
+
+That OpenShift cluster is maintained by the [Community Linux Engineering](https://docs.fedoraproject.org/en-US/cle/)
```
allisonkarlitskaya, line 20: This is the page that had the stale link to filing issues via pagure.  My understanding is that they semi-promised/said "oh we should fix that", but maybe we should also add something like "Last time we did this it was via https://forge.fedoraproject.org/infra/tickets/" ?

allisonkarlitskaya (dismissed):
This is great, thanks!  And I can verify that I'm now able to access the cluster "the right way".  One very minor helpful suggestion, if you feel like it...

centos-ci.md
martinpitt: Sure.. how's this?

allisonkarlitskaya (approved):
Thanks!

PR #703, opened 2026-05-20
aws: add a bucket-wide policy on cockpit-ci-images

allisonkarlitskaya:
Instead of having to worry about setting ACLs on upload, let's make this dirt simple and add a bucket policy to allow unauthenticated downloads only if the filename doesn't start with "rhel".  Otherwise (ie: in order to download RHEL images) you need access to an AWS token which access to read everything.

We only did the ACL stuff with Linode because it's the only tool we had at our disposal.  AWS is a lot more flexible here so let's use it to simplify things.

### DIFF
--- a/ansible/aws/setup-s3-buckets.yml
+++ b/ansible/aws/setup-s3-buckets.yml
@@ -17,14 +17,21 @@
         state: present
         region: "{{ aws_region }}"
         profile: "{{ aws_profile }}"
-        # Enable per-file ACLs: non-RHEL images get public-read, RHEL images get private
-        object_ownership: ObjectWriter
+        # Use bucket policy for public read access (except RHEL images)
+        object_ownership: BucketOwnerEnforced
         public_access:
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
NB: I already ran this.  I had to add `s3:PutBucketOwnershipControls` to the `arr-cockpit-bootstrap` user to get it working.

martinpitt (approved):
OMG, we can have file name pattern based policies! YESSSSH!

You said you already deployed this, great to hear that you set it up! Did you run into any ambiguities in the docs?

allisonkarlitskaya:
> You said you already deployed this, great to hear that you set it up! Did you run into any ambiguities in the docs?

in the S3 bucket policy docs or the `NotResource`?  I just asked Claude :sweat_smile:

martinpitt:
@allisonkarlitskaya no, I meant for setting up the Ansible creds: https://github.com/cockpit-project/cockpituous/blob/main/ansible/aws/README.md#credentials-configuration

PR #711, opened 2026-06-24
.github/workflows: install yarl

jelly:
Fixup from c1b1e9cbba5fbe822154, bots switched from aiohttp to httpx/yarl.


allisonkarlitskaya (approved):
Thanks.

To clarify: the yarl dependency in the bots isn't new — we've had it since before the change.  What happened is that aiohttp depends on yarl but httpx doesn't, so we lost it because we dropped aiohttp.  Probably it should have always been listed here explicitly, since we use it directly.

PR #712, opened 2026-07-16
test: relax <html> check in log.html

allisonkarlitskaya:
The test currently checks for `<html>` but this is a problem because we have a PR in bots which wants to change it to `<html lang="en">` on the advice of biome.

Relax the check.

### DIFF
--- a/test/test_deployment.py
+++ b/test/test_deployment.py
@@ -673,7 +673,7 @@ def test_real_pr(config: Config, request, pod: PodData, bots_sha: str, real_runn
     print(f'----- PR unit-tests log -----\n{log}\n-----------------')
     assert re.search(r'Running on:\s+cockpituous', log)
     assert 'Job ran successfully' in log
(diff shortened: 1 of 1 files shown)


test/test_deployment.py
```diff
@@ -673,7 +673,7 @@ def test_real_pr(config: Config, request, pod: PodData, bots_sha: str, real_runn
     print(f'----- PR unit-tests log -----\n{log}\n-----------------')
     assert re.search(r'Running on:\s+cockpituous', log)
     assert 'Job ran successfully' in log
-    assert '<html>' in get_s3(config, pod, f'logs/{log_name}.html')
+    assert '<html' in get_s3(config, pod, f'logs/{log_name}.html')
```
tomasmatus, line 676: Sounds like the test would already be broken on dark mode?

```
<html class="pf-v6-theme-dark">
```

tomasmatus (approved):
Sounds good

PR #715, opened 2026-09-03
container: Add fedora-distro-aliases

jelly:
This Python module exposes the current active Fedora releases which we can then query in our bots scripts to automate Fedora image related updates.

### DIFF
--- a/tasks/container/Containerfile
+++ b/tasks/container/Containerfile
@@ -40,6 +40,7 @@ RUN dnf -y update && \
         python3-aiohttp+speedups \
         python3-aioresponses \
         python3-build \
+        python3-fedora-distro-aliases \
         python3-flake8 \
         python3-gssapi \
         python3-httpx+http2 python3-respx \


jelly:
The motivation is https://github.com/cockpit-project/bots/issues/9375

tasks/container/Containerfile
```diff
@@ -41,7 +41,6 @@ RUN dnf -y update && \
         python3-aioresponses \
         python3-build \
         python3-fedora-distro-aliases \
-        python3-flake8 \
```
allisonkarlitskaya, line 43: This is still used on the rhel8 branch...

allisonkarlitskaya (commented):
I'm not vetoing the flake8 thing per se because the rhel8 dependency on it is optional, but... probably leave this here for now?

tasks/container/Containerfile
jelly: Ok, dropping we do pin the container there but lets keep it for now.

allisonkarlitskaya:
awful interesting fail there...

allisonkarlitskaya:
Actually, I think this is just because you proposed from a fork: the workflow doesn't get the ability to write statuses to the commit on the target repo in that case.  Can you try re-proposing?

allisonkarlitskaya:
... or maybe GitHub was just having a bad day yesterday.  It's :green_square: now.

allisonkarlitskaya (approved):
This is obviously OK now

PR #718, opened 2026-09-09
.github: Introduce dependabot to update workflow actions

jelly:
In all other projects we keep our workflow actions up to date with dependabot, use it here as well.

---

We have several old actions such as:

.github/workflows/build-tasks.yml:        uses: actions/checkout@v5


allisonkarlitskaya (approved):
...and should we add the projects workflow to make sure it also ends up on the projects board?

